PR #36, opened 2026-04-25
refactor: Phase 2 — Family merge (5 dispatchers, 13→5 tools, v1.0.0 prep)

Harusame64:
## Summary

Phase 2 of Tool Surface Reduction (`docs/tool-surface-reduction-plan.md` §8.2 / §15 Phase 2). Merges 13 tools into 5 family dispatchers via discriminated `action` parameter.

Design doc: [`docs/tool-surface-phase2-family-merge-design.md`](docs/tool-surface-phase2-family-merge-design.md) (Status: Implemented)

## Changes — 13 tools → 5 dispatchers (62% reduction)

| family | merged tools |
|---|---|
| `keyboard` | `keyboard_type` / `keyboard_press` |
| `clipboard` | `clipboard_read` / `clipboard_write` |
| `window_dock` | `pin_window` / `unpin_window` / `dock_window` |
| `scroll` | `scroll` / `scroll_to_element` / `smart_scroll` / `scroll_capture` |
| `terminal` | `terminal_read` / `terminal_send` + **new `run` action** |

**Highlight: `terminal({action:"run", ...})`** — new workflow that combines send → wait → read in one call, with first-class `completion={reason: 'quiet' | 'pattern_matched' | 'timeout' | 'window_closed' | 'window_not_found'}`.

## Implementation history

- 2a (Sonnet, `f332664`): keyboard / clipboard / window_dock dispatchers
- 2b (Sonnet, `1d18db9`): scroll dispatcher (4 → 1)
- 2c (Sonnet, `afc7cde`): terminal dispatcher + run workflow
- BLOCKING fix (Opus, `647498a`): Sonnet got stuck on Phase 2d (E2E re-ran 4× violating 2-cap rule). Opus took over: LLM-visible string cleanup (suggested-fix-store / _action-guard / _errors / browser / dock / desktop-register / desktop-executor / screenshot / ui-elements / perception / pin / scroll-capture / scroll-to-element / smart-scroll / guards), instructions text update.
- 2d (Opus, `891940a`): README / CHANGELOG / system-overview / design status / test name follow-up

## Done criteria

- [x] `tsc --noEmit` exit 0
- [x] vitest unit: **2052 pass / 24 skipped / 0 fail**
- [x] `tools/list` legacy 13 names **0 hits** in `src/` (description / suggest / error.message / literal types)
- [x] new 5 dispatchers registered (`keyboard` / `clipboard` / `window_dock` / `scroll` / `terminal`)
[...]

PR #37, opened 2026-04-25
fix(terminal): PR #36 follow-up — Codex review of Phase 2c run() workflow

Harusame64:
## Summary

Three pre-existing bugs in `terminalRunHandler` (Phase 2c, merged via #36) that Codex flagged on the merge commit `cf41b08`. Fixing in a follow-up PR so `main` is clean before Phase 3 (browser 再配置).

| Pri | File | Issue |
|---|---|---|
| **P1** | `src/tools/terminal.ts:609` | `sinceMarker` baseline taken AFTER send → fast commands' output sliced off |
| **P1** | `src/tools/terminal.ts:579` | `sendOptions` / `readOptions` forwarded raw, bypassing public schemas (`chunkSize:0` can hang) |
| **P2** | `src/tools/terminal.ts:702` | Final read's `ok:false` ignored → false-success when `source:'uia'` on TextPattern-less terminal |

### P1 — Pre-send baseline marker
Capture `sinceMarker` BEFORE `terminalSendHandler` returns. The same baseline text is reused as the quiet-mode change-detection start so the timer correctly waits "quietMs from when output starts appearing" (net read count unchanged).

### P1 — Strict-mode option whitelists
New `TERMINAL_RUN_SEND_OPTIONS_SCHEMA` / `TERMINAL_RUN_READ_OPTIONS_SCHEMA` derived from the public `terminalSendSchema` / `terminalReadSchema`, with owned fields (`windowTitle`, `input`, `sinceMarker`) excluded. `.partial().strict()` rejects unknown keys, schema-bounds enforce min/max, enums enforce literals. Validation runs before any I/O so invalid options cannot trigger side effects. Returns `failWith` with a clear pointer to the public schemas.

### P2 — Surface read failures
Added `ReadFailurePayload` (`code`, `error`, `suggest`) to `TerminalRunResponse.readError`. When the final read returns `ok:false`, run sets `ok:false` and attaches the error payload + warning instead of silently returning `output:""`.

## Test plan

- [x] `npm run build` (tsc clean)
- [x] `tests/unit/terminal-run-validation.test.ts` — 7 new cases (chunkSize:0, unknown keys, windowTitle override, method enum, lines overflow, source enum, sanity-fallthrough)
- [x] Full vitest unit suite — 0 regressions (2068 prior pass + 7 new = 2075 pass)
- [x] Local E2E (1811 unit + 217 e2e + 22 skipped) — only environmental flakies in `context-consistency.test.ts > hasModal/pageState`, all caused by the local Claude Code window title containing the literal word "errors" (matches `MODAL_RE`); does not reproduce on CI runners
- [ ] CI (GH Actions) — build + CodeQL must stay green; expect 0 failures since no Claude Code window exists on the runner

## Notes for reviewer

- P2 read-failure propagation is not unit-tested. `terminalReadHandler` is closed over via static import inside `terminalRunHandler`, so `vi.doMock` cannot intercept the in-module call. Refactoring to inject the read function is deliberately out of scope; behavioral contract is verified by code review and the `source:'uia'` E2E path.
- `MODAL_RE` substring matching (`/error/i` matches "errors") is a separate latent issue — out of scope here, candidate for its own PR.

### DIFF
--- a/src/tools/terminal.ts
+++ b/src/tools/terminal.ts
@@ -589,21 +589,25 @@ export const terminalRunHandler = async ({
   // Reject invalid sendOptions/readOptions BEFORE doing any I/O so unbounded
   // values (e.g. chunkSize:0 hanging the background loop, source:'uia' on a
   // non-TextPattern terminal) cannot bypass the public schema bounds.
+  // Use fail() directly (not failWith) so we get code:"InvalidArgs" and the
+  // suggest[] array stays at the top level. failWith would classify "Invalid
+  // sendOptions" as the generic "ToolError" code and bury our custom suggest
+  // strings under context.suggest, which mis-classifies argument errors as
+  // internal errors and hides actionable remediation guidance from callers.
   let validatedSendOptions: z.infer<typeof TERMINAL_RUN_SEND_OPTIONS_SCHEMA> = {};
   if (sendOptions !== undefined) {
     const parsed = TERMINAL_RUN_SEND_OPTIONS_SCHEMA.safeParse(sendOptions);
     if (!parsed.success) {
-      return failWith(
-        new Error(`Invalid sendOptions: ${describeZodIssues(parsed.error)}`),
-        "terminal:run",
-        {
-          windowTitle,
-          suggest: [
-            "Refer to terminal(action='send') schema for valid keys/types",
-            "windowTitle, input, and sinceMarker cannot be overridden via sendOptions",
-          ],
-        }
-      );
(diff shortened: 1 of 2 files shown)


Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
@codex review

PR #64, opened 2026-04-27
feat(keyboard, terminal): Focus Leash Phase A — terminal-class auto-route to WM_CHAR

Harusame64:
## Summary

When `keyboard(method:'auto')` or `terminal(action:'send', method:'auto')` targets a known terminal class (`CASCADIA_HOSTING_WINDOW_CLASS` for Windows Terminal, `ConsoleWindowClass` for cmd/PowerShell/conhost), routing now defaults to **HWND-targeted PostMessage (WM_CHAR) delivery** instead of the foreground keystroke path. Keystrokes intended for a terminal can no longer be diverted to other windows when the user grabs focus mid-stream.

### Why this matters

The user's stated problem: "外部ターミナルを desktop-touch で操作中にアクティブウィンドウをユーザが無理矢理動いた場合、間違った場所に書込をしてしまう" — typing 'rm -rf node_modules' to a terminal, user clicks browser address bar, partial command leaks into the URL bar. With Phase A, the WM_CHAR path is HWND-bound so foreground state changes have no effect on delivery.

### What changed

- Extracted `resolveEffectiveInputMethod` helper in `keyboard.ts` (reused by `keyboardTypeHandler` and `keyboardPressHandler`).
- Applied the same auto-pick rule in `terminal.ts` (`terminal_send`) inline using existing `TERMINAL_WINDOW_CLASSES` from `bg-input.ts`.
- `DTM_BG_AUTO=1` env flag still enables BG globally for non-terminal apps; non-terminal targets continue to default to the foreground path.
- Class misclassification falls through to foreground via the existing `canInjectViaPostMessage` gate — fully backwards-compatible.

## Phase context

This is **Phase A of the Focus Leash System** (3-layer defense plan):

| Layer | Status |
|---|---|
| **A**: Transport auto-pick (terminal → WM_CHAR) | ⬅ this PR |
| B: Per-chunk foreground guard (non-terminal apps) | follow-up |
| C: lensId leash + perception language (`identity_changed` + `leash_broken`) | follow-up |

Phase A alone resolves the main user-reported issue. B/C extend coverage to non-terminal apps and unify recovery vocabulary.

## Test plan

- [x] 15 new unit tests in `tests/unit/keyboard-method-resolution.test.ts` cover the decision tree (explicit values, DTM_BG_AUTO precedence, terminal class allowlist, degraded inputs)
- [x] Full suite: 2232 pass / 25 skip / 0 unrelated regressions (one e2e flake `browser-cdp` was confirmed flaky via pinpoint re-run — unrelated to keyboard/terminal)
- [ ] Dogfood: open Windows Terminal, run a long `keyboard(action:'type', windowTitle:'PowerShell', ...)`, programmatically focus another window mid-stream, confirm keystrokes still land in the terminal only

## Files

- `src/tools/keyboard.ts` — extract helper, call from type/press handlers, update method param description
- `src/tools/terminal.ts` — class-aware auto-pick in `terminal_send`, update method param description
- `src/stub-tool-catalog.ts` — regenerated descriptions
- `tests/unit/keyboard-method-resolution.test.ts` — new (15 tests)
- `CHANGELOG.md` — Unreleased entry


Harusame64:
@codex review

Addressed both P1 inline comments (commit 6d71903):

- Extracted `evaluateKeyboardGuards` helper covering lensId guards + auto-guard.
- Both BG paths (`keyboard:type`, `keyboard:press`) now call it before `postCharsToHwnd`/`postKeyComboToHwnd`. Block → return failure (perception attached). Pass → BG send proceeds and attaches `_perceptionForPost` to the response.
- `foregroundVerified=true` is passed because BG is HWND-targeted; that flag only skips the foreground gate per `_action-guard.ts` docstring, while modal/identity/dirty/focusedElement gates still run.
- 11 new unit tests in `tests/unit/keyboard-guard-evaluation.test.ts` cover every branch.

Foreground guard blocks at lines 488+/675+ kept inline for this PR; DRY follow-up tracked separately.

Harusame64:
@codex review

Addressed the 3rd P1 (terminal.ts partial WM_CHAR delivery, commit 75a47e3):

- Collapsed the two asymmetric branches in `terminal_send` BG path (`method:'background'` already failed; `method:'auto'` silently succeeded) into a single `failWith('BackgroundInputIncomplete')`.
- Justification: `sent > 0` → foreground fallback would double-input; `sent === 0` → ok:true would mask command loss. Both arms unsafe.
- Behavior now symmetric with `keyboard:type` BG path which already fails on partial regardless of method.
- Suggest hints include UIPI/elevation hint as the most common cause.

Test rationale documented in commit message: terminal_send has no isolated mock fixture, so verification is by code inspection (symmetric to keyboard.ts) plus existing e2e coverage of the success path.

Harusame64:
@codex review

Addressed round 3 P1 (keyboard:press BG fall-through, commit 3ff3f15):

- Collapsed the press BG path's two branches (`background` always failed, `background-auto` fell through to foreground) into a single `failWith('BackgroundInputIncomplete')`.
- Matches the symmetric behavior in `keyboard:type` BG path and `terminal_send` BG path: all three now fail hard on partial WM delivery, never replay via foreground.
- Suggest hints include UIPI/elevation guidance for the most common cause.

Test rationale: same as terminal.ts fix — handler-level partial-fail path has no isolated mock fixture, but symmetry with two other already-correct paths verifies the fix by inspection. Full suite green (2244 pass).

PR #65, opened 2026-04-27
feat(keyboard): Focus Leash Phase B — per-chunk foreground guard for non-terminal apps

Harusame64:
## Summary

Phase A (#64) handled the terminal-class case. Phase B extends the Focus Leash to **non-terminal foreground SendInput**: when `keyboard(action:'type', windowTitle:'X')` is mid-stream and the user grabs focus, the call now aborts with `FocusLostDuringType` and returns the partially-delivered state so the caller can resume.

### Why this matters

Before Phase B, typing a long string into Notepad while the user clicked elsewhere would silently land partial text in the wrong window. Now the call aborts as soon as foreground changes, and returns:

\`\`\`json
{ \"ok\": false, \"error\": \"keyboard:type failed: FocusLostDuringType\",
  \"context\": { \"context\": {
    \"typed\": 32,                    // chars delivered to original target
    \"remaining\": \"...rest of text\", // unsent tail
    \"total\": 100, \"chunkSize\": 8,
    \"focusLost\": { \"stolenBy\": \"Chrome\", \"stolenByProcessName\": \"chrome\", ... }
  } } }
\`\`\`

The caller can re-focus the target and call \`keyboard(action:'type', text: context.remaining)\` to finish.

## Implementation

- **\`src/tools/_focus.ts\`**: add \`checkForegroundOnce({target, homingNotes})\` — thin wrapper around \`detectFocusLoss\` with \`settleMs:0\`. Reuses the same measurement logic; no duplication.
- **\`src/tools/keyboard.ts\`**:
  - New \`abortOnFocusLoss?: boolean\` param. Default \`true\` when \`windowTitle\` is set; \`false\` otherwise (caller named a target = caller cares).
  - New \`getLeashChunkSize()\` helper: reads \`DTM_LEASH_CHUNK_SIZE\` env (default 8, clamped to [1, 1024]).
  - In the foreground keystroke path, when leash is enabled, the text is split into chunks and \`checkForegroundOnce\` runs before each. On theft: \`failWith('FocusLostDuringType')\` carrying typed/remaining/total/chunkSize/focusLost in context.
  - Clipboard path (atomic Ctrl+V) and BG path (HWND-targeted WM_CHAR — Phase A) are unchanged.

## Phase context

| Layer | Status |
|---|---|
| **A**: Transport auto-pick (terminal → WM_CHAR) | merged in #64 |
| **B**: Per-chunk foreground guard (non-terminal apps) | ⬅ this PR |
| C: lensId leash + perception language (\`identity_changed\` + \`leash_broken\`) | follow-up |

## Test plan

- [x] 12 new unit tests in \`tests/unit/keyboard-leash-guard.test.ts\`:
  - \`getLeashChunkSize\` env parsing (4 cases: unset / valid / out-of-range / invalid)
  - Chunked send behavior (8 cases): default with windowTitle / explicit \`abortOnFocusLoss:false\` / no windowTitle / theft on 1st check / theft on 2nd check / partial response shape / non-multiple chunk size / clipboard bypass
- [x] Full suite green: 2256 pass / 25 skip / 0 fail
- [ ] Dogfood: open Notepad, run a long \`keyboard(action:'type', windowTitle:'Notepad', text:'<200 chars>')\`, click into Chrome mid-stream, verify Notepad has only the first chunks and the response includes the unsent tail in \`context.remaining\`

## Files

- \`src/tools/_focus.ts\` — add \`checkForegroundOnce\`
- \`src/tools/keyboard.ts\` — schema, helper, send loop, description
- \`src/stub-tool-catalog.ts\` — regenerated description
- \`tests/unit/keyboard-leash-guard.test.ts\` — new (12 tests)
- \`CHANGELOG.md\` — Unreleased entry


Harusame64:
@codex review

Addressed P1 (abortOnFocusLoss not reachable through registered schema, commit c606bfb):

- Added the field to the action:'type' branch of `keyboardSchema` (the discriminated union the registered tool actually validates against). Previously the field was only in `keyboardTypeSchema` (the legacy exported field dict, unused at runtime).
- Tried collapsing the two duplicate field sets via Zod object spread (`...keyboardTypeSchema`) — but the stub-catalog generator parses union variants statically and cannot follow spread, so it lost every per-action property in the catalog. Reverted the spread; inlined fields in the union directly to keep the generator working.
- 3 new regression tests pin abortOnFocusLoss reachability through the public `keyboardSchema` (`safeParse` with true/false/optional). Future drift between the duplicate declarations now trips a test rather than slipping through to a runtime stripped-arg bug.
- Code comment on the union flags the constraint + points at the regression tests so the next refactor attempt learns why spread doesn't work here without also changing the generator.

Harusame64:
@codex review

Addressed Gemini + Codex PR #65 round-2 feedback in commit e168735:

**Gemini — modifier release safety valve (defense-in-depth):**
- Added `releaseDanglingModifiers()` emitting KeyUp for L/R Ctrl, Alt, Shift on the leash abort path and any unexpected throw inside the chunk loop. KeyUp is idempotent at the OS level so this is a free safety net against the 'modifier stuck-down' UX hazard. Current implementation aborts at chunk = character boundaries (no real leak), but the safety valve protects against future raw-SendInput chunking, mid-loop exceptions, or undocumented nutjs behavior.

**Codex P2 — surrogate pair chunking:**
- Switched `String.slice` to `Array.from` codepoint iteration so emoji and other non-BMP characters are never bisected at chunk boundaries. `typed` still counts UTF-16 code units to keep the resume contract (`remaining = effectiveText.slice(typed)`) coherent.

7 new unit tests cover both fixes (4 safety valve + 3 emoji surrogate).

PR #66, opened 2026-04-28
feat(desktop_act): include blockingElement in modal_blocking (#63)

### LINKED ISSUE #63: feat(desktop_act): include blocking element info in modal_blocking response
## Background

A 3-model comparison test (Opus / Sonnet / Haiku) running the same desktop_act task on Outlook PWA — click 「新規メール」 → type \"test\" in body — completed successfully on **all three models**. Notably **Haiku 4.5 finished in just 10 tool calls (57s)** by aggressively trusting structured response fields (constraints, attention, recovery hints, post.focusedElement) instead of pre-action reasoning.

During that run, the Haiku agent left an actionable design suggestion that I want to record here.

## Haiku's feedback (verbatim, translated)

> modal_blocking 時のエラーメッセージに、どのモーダルが原因かを示す情報があれば、対応が迅速化したはず。
> (\"If the modal_blocking error response indicated *which* modal was responsible, recovery would have been faster.\")

This came up because clicking 「新規メール」 on Outlook PWA returned \`{ ok: false, reason: \"modal_blocking\" }\` due to a Copilot side-panel being treated as a modal — but the agent had to take a separate screenshot to identify what to dismiss.

## Current shape

\`src/engine/world-graph/guarded-touch.ts:236-238\`

\`\`\`ts
if (this.env.isModalBlocking(entity)) {
  return { ok: false, reason: \"modal_blocking\", diff: [] };
}
\`\`\`

\`isModalBlocking(entity: UiEntity): boolean\` returns only a flag — the offending element is dropped on the floor even though the implementation must have looked at it.

## Proposed change

Extend the env interface so the blocking element identity propagates to the response:

\`\`\`ts
// Option A: replace boolean with a discriminated return
getBlockingModal(entity: UiEntity): { name: string; role: string; automationId?: string } | null;

// Option B: keep isModalBlocking, add a paired query
isModalBlocking(entity: UiEntity): boolean;
findBlockingModal(entity: UiEntity): UiEntity | null;
\`\`\`

Then in guarded-touch.ts:

\`\`\`ts
const blocker = this.env.findBlockingModal(entity);
if (blocker) {
  return {
    ok: false,
    reason: \"modal_blocking\",
    diff: [],
    blockingElement: { name: blocker.label, role: blocker.role, automationId: blocker.automationId },
  };
}
\`\`\`

Schema-wise, the response Zod type for desktop_act needs an optional \`blockingElement\` field, mirrored in the README \`modal_blocking → dismiss modal via click_element, then retry\` recovery note (so the LLM is told to use blockingElement.name as the dismiss target).

## Why this matters

The whole point of pre-digesting expert knowledge into response strings is to raise the floor for fast / weaker LLMs. \`reason: \"modal_blocking\"\` already does this for *what to do* (dismiss + retry), but **leaves *which one to dismiss* to model-side inference / a separate screenshot**. Closing that gap is one syscall away from being free.

## Acceptance criteria

- [ ] desktop_act response includes \`blockingElement: { name, role, automationId? }\` when reason=modal_blocking
- [ ] env interface change reflected in tests/unit/guarded-touch.test.ts
- [ ] README modal_blocking recovery note updated to mention dismissing \`blockingElement.name\`
- [ ] Backwards compatibility: blockingElement is optional (engines that can't identify the modal still return modal_blocking without it)

Harusame64:
## Summary

Closes #63 — Haiku 4.5 dogfood feedback from the 3-model comparison run on Outlook PWA.

When `desktop_act` fails with `reason: "modal_blocking"`, the response now carries `blockingElement: { name, role, automationId? }` identifying the offending modal. The LLM can dismiss it directly via `click_element(name=blockingElement.name)` without taking an extra screenshot to figure out what to close.

## Design (Issue #63 Option B)

- Existing `isModalBlocking(entity): boolean` kept as-is for backwards compat.
- New `findBlockingModal?(entity): UiEntity | null` added as an **optional** sibling on `TouchEnvironment` — env layers without it silently omit the field, so existing test fakes pass without modification.
- `session-registry` default derives both methods from a shared `isModalCandidate()` predicate so they cannot diverge. Custom overrides on `opts.isModalBlocking` should pair with `opts.findBlockingModal`; divergence (`true` / `null`) silently drops `blockingElement` instead of crashing (defensive).
- Name resolution falls back through `locator.uia.name` → `entity.label` → `entity.role` → `"modal"` so the field is always non-empty for downstream `click_element` lookups even when UIA "unknown"-role dialogs lack a label.
- `blockingElement` is attached via spread-omit (`...(blocker ? {…} : {})`) so the key is genuinely absent (not present-as-undefined) when finder returns null.

## Test plan

- [x] `tests/unit/guarded-touch.test.ts` — 3 new cases:
  - happy path: `findBlockingModal` returns entity → `blockingElement` carries `{ name, role, automationId }`
  - fallback: `label` missing → `locator.uia.name` used; `automationId` omitted when locator does not provide one
  - predicate divergence: `isModalBlocking=true` but `findBlockingModal=null` → `"blockingElement" in result === false`
- [x] `tests/unit/desktop-facade.test.ts` — existing E2E test extended to assert the production default propagates `blockingElement` through `session-registry` → `guarded-touch`
- [x] Full suite green: **2269 passed / 25 skipped / 0 failed**

## Docs

- `desktop_act` tool description (`desktop-register.ts`) and MCP server instructions (`server-windows.ts`) recovery hints updated.
- `README.md` / `README.ja.md` `modal_blocking` recovery sections updated in both the user-facing Recovery hints block and the V1 fallback section.
- `CHANGELOG.md` Unreleased entry.

## Backwards compatibility

`blockingElement` is optional. Engines that cannot identify the modal still return `modal_blocking` without it — no breaking change to consumers.A


Harusame64:
@codex review

Issue #257, opened 2026-05-12
[demo finding F3] keyboard tool needs a sequence mode for menu navigation (Alt+<letter>, <letter>)


Harusame64:
## What I observed

To trigger `Insert > Module` in the VBA Editor I had to send two separate `keyboard` tool calls:

```
keyboard(action='press', keys='alt+i')   # opens 挿入 menu
keyboard(action='press', keys='m')        # selects モジュール
```

When a `desktop_state` call landed between them, the Insert menu closed before `M` arrived. Firing them in parallel (single MCP message) appeared to crash the server — see F4.

For an LLM driving menu navigation this is a very common pattern (Alt + menu mnemonic, then sub-letter). The current `keyboard.press` shape isn't a clean fit.

## Proposal

Add a sequence form, e.g.:

```
keyboard(action='sequence', steps=[
  { keys: 'alt+i', holdMs: 50 },
  { keys: 'm' }
])
```

Semantics: each step is executed in order on the same call, modifiers from one step are released before the next step (unless explicitly held), and inter-step delay defaults to something menu-pump-safe (50–100 ms). The whole sequence is one tool call → no risk of concurrent callers re-stealing focus mid-way.

This also unblocks fixing F3-style flake without inventing client-side workarounds.

## Related demo finding

Filed alongside F1 (workspace_launch), F2 (UIA-blind VBE), F4 (parallel keyboard crash).

🤖 Reported by Claude Code during dogfood session 2026-05-12

Issue #255, opened 2026-05-12
[demo finding F4] MCP server disconnects when 2 keyboard tool calls fire in parallel during Excel/VBE session


Harusame64:
## What I observed

During a dogfooding demo run (Claude operating Excel + VBA Editor through the MCP), I sent two `keyboard` tool calls in a single MCP message — `keys=alt+i` and `keys=m`, intended to fire `Insert > Module` in the VBA Editor.

The first call returned `No such tool available: mcp__desktop-touch__keyboard` (i.e. the server had already disconnected). The second returned the same error.

Subsequent system-reminder events declared all `mcp__desktop-touch__*` tools deferred and unavailable, confirming the MCP server process had dropped its connection to the Claude Code client.

## Why this matters

LLM-driven sessions naturally fire multiple tool calls in parallel — that's exactly what reduces round-trip latency. If parallel `keyboard` calls can crash the server, real users will hit this too (and far more often than a hand-driven session would), especially during menu navigation patterns like `Alt+<letter>` + `<letter>` that an LLM is likely to batch.

A single crashed-mid-demo session is exactly the failure mode that kills viral 30-second demo videos — the project's current promotion plan.

## Repro (best-effort, mid-demo conditions)

1. Open Excel via `workspace_launch`
2. Activate it with `focus_window`
3. Press `enter` to open a new workbook (start screen)
4. Press `alt+F11` to open VBA Editor (focused window becomes `Microsoft Visual Basic for Applications - Book1`, `focusedElement: null` since VBE is UIA-blind)
5. In a single MCP message, send two `keyboard` tool uses: `{action: 'press', keys: 'alt+i'}` and `{action: 'press', keys: 'm'}`
6. Observe — the server may disconnect.

I cannot guarantee deterministic repro from a single run, but the timing (crash immediately after firing parallel keyboard calls into a VBE window) makes a race condition the leading hypothesis.

## Possible root causes (need investigation)

- **Race in the keyboard executor** when two SendInput sequences interleave (modifier release ordering breaks)
- **Modifier still held** from call A when call B starts → keys merge into unintended chord
- **VBE-specific** — VBE's window-message pump may panic on rapid Alt-menu sequences
- **napi worker** dropping a panic without surfacing it, killing the IPC pipe

## Suggested first investigation steps

- Wrap the keyboard worker thread in `catch_unwind` and log
[...]

PR #268, opened 2026-05-12
fix(keyboard): serialize concurrent tool calls to avoid libnut crash (#255)

Harusame64:
## Summary
Fix release-blocker bug [#255](https://github.com/Harusame64/desktop-touch-mcp/issues/255) — parallel \`keyboard\` tool calls (the natural chord pattern for menu navigation like \`alt+i\` then \`m\`) crashed the MCP server, taking the entire \`mcp__desktop-touch__*\` namespace down mid-session.

## Root cause hypothesis
@nut-tree-fork/nut-js's libnut backend is not safe for concurrent SendInput invocations. JS scheduling lets two \`await keyboard.pressKey(...)\` calls yield between the press and release of the same key list, leaving libnut's modifier-state tracking in an undefined condition that segfaults the C++ binding. The race window is wide enough to hit reliably with the dual-Alt menu pattern that LLMs naturally produce in a single tool-use turn.

## Fix
Single-writer FIFO at the \`keyboardHandler\` entry (\`src/tools/keyboard.ts:38-69\`). The queue tail tracks the completion *point* of the in-flight call, so a rejection does not poison the queue. Serialization is scoped to the keyboard tool — mouse / scroll / clipboard are unaffected; cross-tool input contention has not been reported.

## Behaviour preservation
- Sequential \`keyboard\` calls: identical timing, identical behaviour
- \`run_macro\` batches: identical (they were already sequential at the dispatcher layer)
- Only true cross-request concurrency is held back, and only at the keyboard boundary

## Test plan
- [x] New regression test \`tests/unit/keyboard-input-serialization.test.ts\` (2 cases):
  - Three parallel \`keyboardHandler({action:'press'})\` calls execute in strict serial order (press-end of N precedes press-start of N+1)
  - An early rejection does not prevent subsequent queued calls from running
- [x] All 50 existing keyboard unit tests pass (4 files)
- [x] Pre-existing main-branch failures (\`tool-naming-phase2\`, \`tool-descriptions\`, \`benchmark-gates\`, \`scroll-raw-verify\`, \`foreground-flash-verification\` e2e) reproduce on \`main\` without this change — unrelated to this fix

## Out of scope (separate follow-ups, all authorized by user)
- Issue [#257](https://github.com/Harusame64/desktop-touch-mcp/issues/257) F3 keyboard sequence mode — held-modifier menu chains in a single call
- Cross-tool input mutex (keyboard + mouse) — not reported; can be added if needed
- Defense-in-depth \`catch_unwind\` around the native binding — libnut is JS-bound, no Rust worker to wrap; the mutex addresses the symptom at the boundary where we control scheduling

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **`keyboard` tool no longer crashes the MCP server when called in
+  parallel.** Two `keyboard` tool calls fired in the same Claude turn
+  (the natural pattern for menu chords like `alt+i` then `m`) could
+  interleave inside the native key-injection backend (libnut) and
+  segfault the Node process. The whole `mcp__desktop-touch__*`
+  namespace then disappeared from the session, requiring a CLI
+  restart. The handler now serializes calls through an internal
+  FIFO so the press/release windows of one call always finish
+  before the next one starts. Sequential `keyboard` calls and
+  `run_macro` batches behave exactly as before — only true cross-
+  request concurrency is held back, and only at the keyboard tool
+  boundary (mouse / scroll / clipboard are unaffected). Fixes
+  [#255](https://github.com/Harusame64/desktop-touch-mcp/issues/255).
+
 ## [1.5.0] - 2026-05-12 — New `excel` tool: author and run VBA macros against Excel via COM (no VBA Editor UI needed)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,18 +4,22 @@
 
 ### Fixed
 
-- **`keyboard` tool no longer crashes the MCP server when called in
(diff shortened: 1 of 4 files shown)


Harusame64:
@codex review

Harusame64:
@codex review

Round 1 にて Opus が P2-1 (silent scope shrink — scroll/terminal が同 libnut race を踏む) を検出。Commit 779f9a9 で lock を engine layer (src/engine/nutjs.ts) に lift、wrapped `keyboard` export で全 caller (keyboard / scroll / terminal) を cover。

Round 2 review 観点:
- engine wrap (Object.assign + Object.create pattern) の正しさ
- raw mock 経由 test の妥当性
- TypeScript type cast `as RawKeyboard` の安全性
- prototype chain / getter/setter properties spread の漏れ

PR #269, opened 2026-05-12
feat(workspace_launch): resolve bare exe names via App Paths registry (#258)

### LINKED ISSUE #258: [demo finding F1] workspace_launch should resolve common app names from App Paths registry (excel, winword, etc.)
## What I observed

`workspace_launch(command='excel.exe')` failed with `SpawnFailed: Command \"excel.exe\" not found. Provide the full path`. The full path is registered under `HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\EXCEL.EXE` — this is Windows' standard mechanism for short-name app launching (works in Win+R, Explorer address bar, etc.).

The error suggestion text told me to provide a full path, but the OS itself can resolve `excel.exe` via App Paths.

## Why this matters

- LLM users will type `excel.exe`, `winword.exe`, `outlook.exe`, `chrome.exe`, `code.exe` and expect them to work — these are all App Paths-registered
- Forcing a full path means the model has to first probe the filesystem / registry from a terminal tool, which is one more round trip and a more brittle prompt
- This is exactly the friction that makes a first-time demo feel rough — the very first action gates on a 2-step recovery

## Proposed fix

In the `workspace_launch` spawn path, when the executable name is not in PATH:

1. Look up `HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\<name>` and `HKLM:\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\App Paths\<name>` and `HKCU:` mirrors
2. If found, use the `(Default)` value as the executable path
3. If still not found, return the existing `SpawnFailed` with the same suggestions

The lookup is essentially the same as `Invoke-Item` / `Start-Process` behavior — calling `ShellExecuteEx` instead of `CreateProcess` would handle this for free, but if we want to stay on `CreateProcess` for control, the registry lookup is ~10 lines of Rust.

Tested registry locations on my Win11 box:

```
HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\EXCEL.EXE  →  C:\Program Files\Microsoft Office\Root\Office16\EXCEL.EXE
HKLM:\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\App Paths\EXCEL.EXE  →  same
```

## Related demo finding

Filed alongside F2 (VBE UIA-blind), F3 (keyboard sequence), F4 (parallel keyboard crash).

🤖 Reported by Claude Code during dogfood session 2026-05-12

Harusame64:
## Summary
Closes [#258](https://github.com/Harusame64/desktop-touch-mcp/issues/258) — \`workspace_launch({command:'excel.exe'})\` and other bare-name Office / Windows app launches now resolve via the App Paths registry (the same mechanism Win+R uses).

## Why
Node's \`spawn\` only searches PATH, not the App Paths key Windows itself uses for short-name app launching. That gap meant LLM-natural prompts like \`workspace_launch(command='excel.exe')\` failed even on systems with Office installed normally — forcing a 2-step recovery (probe filesystem from terminal → retry with full path) on the very first action of a demo.

## Resolution chain (extended)
1. Caller path with separator → identity (unchanged)
2. \`WELL_KNOWN_PATHS\` table (chrome / edge / brave / VS Code, unchanged)
3. **App Paths registry — NEW** (HKCU, HKLM, HKLM\WOW6432Node hives; REG_SZ + REG_EXPAND_SZ; %VAR% expansion; quoted-path strip)
4. Otherwise return unchanged (surfaces as ENOENT — matches prior behaviour)

The unified \`resolveLaunchExecutable\` returns \`{ resolved, source }\` so workspace_launch's result note can label \`via App Paths registry\` (vs identity / well-known).

## Security
App Paths is user-writable. After resolution, \`validateLaunchCommand\` is re-run on the resolved path so a tampered entry can't smuggle a blocked shell interpreter (cmd.exe, powershell.exe, ...) past the basename guard. Tested implicitly: BLOCKED_EXECUTABLES set is unchanged.

## Test plan
- [x] 11 new unit tests in \`tests/unit/launch-resolution.test.ts\`:
  - REG_SZ happy path
  - REG_EXPAND_SZ with %VAR% expansion
  - Quoted-value strip
  - Hive fall-through (HKCU miss → HKLM hit)
  - All-hive miss → null
  - Path-separator → identity (no registry call)
  - .exe suffix auto-append for bare names
  - chain order: identity / well-known beats app-paths / app-paths fallback / no-match identity
- [x] tsc clean
- [x] Existing tests unaffected (resolveWellKnownPath kept as-is; browser.ts still uses it directly)

## Out of scope (separate work)
- \`browser.ts\` direct \`resolveWellKnownPath\` use unchanged — browsers are already well-known and App Paths fallback would not add value there
- \`run_macro\` / other entry points that don't pass through workspace_launch
- Word / PowerPoint / Outlook explicit WELL_KNOWN_PATHS entries — App Paths covers them via registry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,26 @@
 
 ## [Unreleased]
 
+### Added
+
+- **`workspace_launch` now resolves common app names from the Windows
+  App Paths registry.** Calls like
+  `workspace_launch({command:'excel.exe'})`,
+  `workspace_launch({command:'winword.exe'})`, or
+  `workspace_launch({command:'outlook.exe'})` previously failed with
+  `SpawnFailed: Command "excel.exe" not found` because Node's `spawn`
+  only searches `PATH`, not the App Paths key Windows itself uses for
+  Win+R and the Explorer address bar. The launch path now consults
+  `HKCU` then `HKLM` (incl. `WOW6432Node`) for an `App Paths\<exe>`
+  entry, expands `%VAR%` tokens in REG_EXPAND_SZ values, and re-runs
+  the existing security validation on the resolved path so a tampered
+  registry entry cannot smuggle a blocked shell interpreter through.
+  The successful result surfaces the resolution in `note`, e.g.
+  `Resolved "excel.exe" → "C:\\...\\EXCEL.EXE" via App Paths registry`.
+  The existing built-in path lookups (chrome / edge / brave / VS Code)
+  still take priority and are unchanged. Fixes
+  [#258](https://github.com/Harusame64/desktop-touch-mcp/issues/258).
+
 ## [1.5.0] - 2026-05-12 — New `excel` tool: author and run VBA macros against Excel via COM (no VBA Editor UI needed)
 
 ### Added
--- a/src/tools/workspace.ts
+++ b/src/tools/workspace.ts
@@ -1,7 +1,7 @@
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { z } from "zod";
 import { mouse } from "../engine/nutjs.js";
-import { validateLaunchCommand, resolveWellKnownPath, spawnDetached } from "../utils/launch.js";
+import { validateLaunchCommand, resolveLaunchExecutable, spawnDetached } from "../utils/launch.js";
 import { enumMonitors, getVirtualScreen, enumWindowsInZOrder, type WindowZInfo } from "../engine/win32.js";
 import { captureScreen } from "../engine/image.js";
 import { clearLayers } from "../engine/layer-buffer.js";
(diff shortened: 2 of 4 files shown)


Harusame64:
@codex review

Harusame64:
@codex review

Round 1 で P2-1 (reg PATH hijack) + P2-2 (stale registry entry existsSync 不在) を Opus が検出。Commit 30d57b1 で apply:
- REG_EXE 絶対パス定数 (TASKKILL_EXE 同型 pattern)
- fs.existsSync gate で stale entry fall-through
- 2 新 test case 追加

Round 2 review 観点:
- REG_EXE PATH hijack defense の妥当性
- existsSync gate の例外処理
- stale fall-through 順序の正しさ
- test default existsSync=true 変更で既存 case 影響

PR #307, opened 2026-05-15
fix(adr-018-phase-5-followup): retarget Tier 3 WM_MOUSEWHEEL to MDI scroll-leaf (Excel/Word)

Harusame64:
## Summary

Fixes a user-visible Excel scroll regression discovered by dogfood on `26d920b` (2026-05-15): `scroll(action:'raw', windowTitle:'Book1 - Excel')` returns `ok:false ScrollNotDelivered` with `verifyDelivery.reason:'target_unreachable'` because `WM_MOUSEWHEEL` propagates upward only (Microsoft Learn / DefWindowProc) — a POST to the top-level `XLMAIN` never reaches the `EXCEL7` child that owns the cell-grid scrollbar.

Adds `win32_find_scroll_leaf_for_top_level` napi (class-chain table over `FindWindowExW`):

```
XLMAIN  → XLDESK → EXCEL7   (Excel cell grid leaf)
OpusApp → _WwF   → _WwG     (Word document body leaf)
```

`postWheelToHwnd` retargets every subsequent call (`PostMessage`, `GetScrollInfo`, L1 capture push, lParam window-center encoding) to the resolved leaf. Apps not in the table return `null` and behave bit-equal to the previous release.

This is the destination-explicit form of the Phase 5 §2.2 carry-over #6. The generic `win32_enum_child_windows` enumerator originally specified there is no longer needed for the Excel/Word fix — the smaller class-chain shape preserves the destination-explicit principle (ADR §2.3 D3) and adds zero overhead for non-MDI scroll calls.

## Approach in numbers

- **+1 napi** (`win32_find_scroll_leaf_for_top_level`, ~60 lines Rust)
- **+1 dispatcher hook** (in `postWheelToHwnd`, ~25 lines TS)
- **+5 unit tests** (scroll-leaf walker dedicated `describe` block)
- **0 changes** to non-Excel/Word scroll paths (Notepad/Explorer/Chrome dogfood passes pre-merge)

## Files changed

- `src/win32/window.rs` — new napi + `SCROLL_LEAF_CHAINS` table + `get_class_name` internal helper
- `src/tools/_input-pipeline.ts` — `postWheelToHwnd` retargets to leaf when walker returns non-null
- `src/engine/native-engine.ts` — `NativeWin32` interface extension
- `index.d.ts` / `index.js` — hand-maintained re-export
- `tests/unit/input-pipeline-dispatch.test.ts` — 5 new tests (leaf returned / null / undefined / different rect / throws)
[...]

PR #308, opened 2026-05-15
fix(adr-018-leaf-walker): chain-table trust on null pre-snapshot — Excel NUIScrollbar / Word _WwG observation gap

Harusame64:
## Summary

PR #307 (b1d6422) landed the leaf walker, but post-merge dogfood (2026-05-16) revealed Excel `scroll(action:'raw', windowTitle:'Book1 - Excel')` still returns `ok:false code:'ScrollNotDelivered'` with `verifyDelivery.reason:'target_unreachable'`.

**Root cause** (verified empirically against a running Excel HWND via the new `win32FindScrollLeafForTopLevel` napi):

```
XLMAIN top-level class:                "XLMAIN"          — chain table hit ✓
leaf walker resolved:                   2624382 (EXCEL7)  — chain resolved ✓
GetScrollInfo(EXCEL7, vertical):       null              — NUIScrollbar invisible to SB_VERT
```

Excel paints its scrollbar as a custom `NUIScrollbar` control instead of a Win32 `SB_VERT`, so `GetScrollInfo` returns null on EXCEL7 even though the wheel POST actually scrolls the cell grid. The existing `postWheelToHwnd` Case 2 (`pre === null → null → caller emits target_unreachable`) was designed for the input-HWND-only path where a null SB_VERT genuinely means "this HWND does not consume WM_MOUSEWHEEL" (Word `_WwG` MFC custom-paint). Once the leaf walker retargets to a class-chain-table-pinned receiver, null SB_VERT instead means "this leaf uses a non-Win32 scrollbar widget" — a distinct case where the receiver still scrolls.

## Fix

Split Case 2 into two branches:
- **Case 2a (retargeted via leaf walker)** — trust the chain table → emit `delivered_via_postmessage` on null pre. The `SCROLL_LEAF_CHAINS` table pins which HWND classes are documented scroll receivers; if the walker took us there AND PostMessage queued successfully, delivery is assured even when observation channel returned null.
- **Case 2b (no retarget)** — emit null → caller surfaces `target_unreachable` per ADR §2.6.2 path-(b). Unchanged from PR #307.

Tracked via a `retargetedByLeafWalker` flag set at the same site that flips `effectiveHwnd`.

## Files changed

- `src/tools/_input-pipeline.ts` — `postWheelToHwnd` Case 2 split into 2a/2b
- `tests/unit/input-pipeline-dispatch.test.ts` — 2 new tests pinning both branches (64 total, was 62)
- `CHANGELOG.md` — `[Unreleased]` entry refined (the previous entry over-promised that observation would catch the move; clarified that the dispatcher trusts the documented receiver contract instead)

## Test plan

- [x] `npm run build` (tsc) green
- [x] `npx vitest run tests/unit/input-pipeline-dispatch.test.ts` — 64 / 64 pass
- [x] End-to-end dogfood on running Excel (`HWND 3214258`):
  - `scroll(hwnd: 3214258, direction:'down')` → `delivered_via_postmessage` ✓
  - `scroll(windowTitle:'Book1 - Excel', direction:'down')` → `delivered_via_postmessage` ✓
- [x] Word regression check (`scroll(windowTitle:'文書 1 - Word')`) → existing `delivered_via_uia` preserved ✓

## Review guidance (CLAUDE.md §3.3)

Step 1 Opus focus:
- Case 2a/2b split correctness: `retargetedByLeafWalker` flag is the only signal — verify it's set ONLY when walker returned a non-null leaf, never otherwise
- AC1 alignment: Excel now reaches `delivered_via_postmessage` per ADR-018 §6 AC1
- Word `_WwG` AC1 path-(b) still allowed: AC1 explicitly permits BOTH `delivered_via_postmessage` AND `target_unreachable` for Word; this PR makes Word fall into path-(a) when Tier 3 retargets to `_WwG` (Tier 1 UIA still short-circuits first in practice)
- Test count sync: §2.1.4 of the leaf-walker sub-plan now says "5 new cases"; this PR adds 2 more (64 total) — minor count drift, acceptable defer-or-fix

Step 2 Codex focus:
- The chain-trust assertion is a behaviour-significant change — does it accidentally swallow PostMessage failures (e.g., elevated target, UIPI block, message queue exhaustion) by emitting `delivered_via_postmessage` when nothing actually moved? Verify the `postedAny` check upstream still gates this.


Harusame64:
@codex review

Follow-up to PR #307 (b1d6422). Post-merge dogfood found Excel scroll still returns `target_unreachable` because Excel paints `NUIScrollbar` instead of Win32 `SB_VERT` — `GetScrollInfo(EXCEL7, vertical)` returns null even though the wheel POST scrolls the cell grid (empirically verified via napi probe).

This PR splits `postWheelToHwnd` Case 2 (`pre === null`) into:
- **Case 2a**: leaf walker retargeted → trust chain table → emit `delivered_via_postmessage`
- **Case 2b**: not retargeted → return null → caller surfaces `target_unreachable` (unchanged from PR #307)

Tracked via `retargetedByLeafWalker` flag.

Please focus on:
1. **Chain-trust safety** — Case 2a asserts "PostMessage success + chain-table retarget = delivered" without observation. Could this swallow real PostMessage failures (UIPI integrity mismatch, message queue full)? Does the existing `if (!postedAny) return null` upstream gate correctly?
2. **Word `_WwG` AC1 path-(a) vs path-(b)** — ADR-018 §6 AC1 permits both for Word. This PR makes Word fall into path-(a) when Tier 3 retargets. Is the ADR's `target_unreachable` for Word still reachable (e.g., when Tier 1 UIA also fails and PostMessage is rejected)?
3. **ADR-018 §2.6.2 docs sync** — §2.6.2 path-(b) says "every applicable tier exhausted without observable delta". With Case 2a chain-trust, no observable delta now sometimes maps to `delivered_via_postmessage` instead of `target_unreachable`. Should §2.6.2 be amended, or is the unobservable-but-trusted case still consistent with the ADR's framing?
4. **`retargetedByLeafWalker` flag — race / spurious-flip** — flag is set inside try/catch on `findLeaf(hwnd)`. Verify the try/catch correctly leaves `retargetedByLeafWalker = false` when `findLeaf` throws.
5. **Test coverage** — 2 new cases pin Case 2a/2b. Anything missing for the chunk-boundary cases (`notch >= 274` chunking interacting with chain-trust)?

Round 1.

Harusame64:
@codex review

Round 2: Round 1 P1 fix landed (commit ee364c4) — chain-trust path now verifies via dHash diff over the leaf's window rect (Hamming distance ≥ `POSTMESSAGE_CHAIN_TRUST_HASH_THRESHOLD` = 5 bits). Below threshold → null → caller emits `target_unreachable` (preserves the no-delivery signal Round 1 flagged). dHash capture failure → fall back to the bare chain-table assertion (preserves Phase 5+N intent on mixed-version builds).

Please re-verify:
1. **Threshold = 5 bits** matches `RAW_SCROLL_HASH_MOVE_THRESHOLD` in `mouse.ts` — same cutoff already used for the Tier 4 SendInput fallback observation. Consistent across surfaces?
2. **dHash unavailable fall-back** still emits `delivered_via_postmessage` (preserves chain-table assertion on platforms lacking the native dHash path) — does this re-open the false-positive concern for production mixed-version builds, or is it an acceptable graceful-degrade?
3. **ADR-018 §2.6.2 docs sync** — the `delivered_via_postmessage` row now splits into "(i) Win32 SB_VERT diff" vs "(ii) chain-trust + dHash diff"; the `target_unreachable` row absorbs the below-threshold dHash case. Internally consistent with the code?

Round 2.

Harusame64:
@codex review

Round 3: the dHash gate from Round 2 (commit ee364c4) is **reverted** (commit 556223b) after empirical dogfood found it structurally inadequate for Excel-class custom-paint receivers.

**Empirical evidence (Excel HWND 3214258, leaf EXCEL7 = 2624382)**:
- PostMessage(WM_MOUSEWHEEL, delta=-360, 3 chunks) → Excel scrolled A1 → A74 visually (verified via OCR Name Box)
- Raw pixel bytes pre/post: **0.36% changed (5120 / 1422400 bytes)** — real pixel-level change
- 8×8 dHash pre/post: **byte-identical (0 bits Hamming distance)** at every settle interval tested (16/50/100/200/500 ms)
- Conclusion: Excel's mostly-uniform cell grid + dark-gray row-label strip collapses to the same 8×8 perceptual hash regardless of which rows are visible. dHash demoted real scrolls to `target_unreachable` — opposite of the user-visible truth.

**Round 3 state**: chain-trust assertion only (PostMessage queued + leaf in `SCROLL_LEAF_CHAINS` → `delivered_via_postmessage`). Round 1 P1 boundary-case false-positive concern is **accepted as documented trade-off** — semantics match Tier 1 UIA's at-boundary handling (`mouse.ts::evaluateScrollDelivery`: wheel queued + receiver chose not to act = delivered no-op). Future-verification work (higher-resolution hash / raw-byte-diff threshold / multi-frame statistical comparison akin to video-codec inter-frame coding / UIA cell-name observation) carry-over to Phase 5+N+1.

Please verify:
1. The revert correctly removed all dHash plumbing (no orphan imports or dead constants in `_input-pipeline.ts`)
2. ADR-018 §2.6.2 and the sub-plan are internally consistent with the reverted state (no stale references to `POSTMESSAGE_CHAIN_TRUST_HASH_THRESHOLD` or dHash-verification semantics in places that say "current behaviour")
3. The revert's Codex P1 trade-off rationale (matches Tier 1 UIA at-boundary semantics) is sound.

PR #318, opened 2026-05-16
feat(adr-019-stage-4): impl — SSIM local_repaint primitive for click / keyboard BG verifyDelivery

Harusame64:
## Summary

ADR-019 Stage 4 impl — SSIM-based `local_repaint` primitive for `mouse_click.verifyDelivery` and `keyboard.type` BG verify fallback paths. Closes the sub-plan §4 P-task checklist (P1-P13 + P15 + P16; P14 = post-merge dogfood, separate work item).

**Production code 改修 PR** per CLAUDE.md §3.3 Step 0 — Codex review **mandatory**. Opus phase-boundary review requested from parent session (worktree isolation prevents internal `Agent` tool use).

## P-task status

- [x] **P1** `src/ssim.rs` (Wang 2004 scalar impl, 9 Rust unit tests)
- [x] **P2** `src/lib.rs` napi export wiring (`compute_ssim_residual`)
- [x] **P3** `src/engine/native-types.ts` + `native-engine.ts` + `index.d.ts` + `index.js` (ESM `createRequire` shim, drift check passes 67/67)
- [x] **P4** `src/engine/local-repaint.ts` (resolver + orchestrator + §2.5 constants)
- [x] **P5** `tests/unit/ssim-residual.test.ts` (8 cases) + `tests/unit/local-repaint-orchestrator.test.ts` (14 cases)
- [x] **P6** `_mouse-verify.ts` `classifyDeliveryWithLocalRepaint` wrapper + `observation` field
- [x] **P7** `mouse.ts` `mouseClickHandler` wiring (pre-frame capture parallel to pre-snapshot)
- [x] **P8** `tests/unit/mouse-click-verify-stage4.test.ts` (7 cases)
- [x] **P9** `keyboard.ts` `typeHandler` BG verify wiring (OQ #5 option (a))
- [x] **P10** `tests/unit/keyboard-type-stage4.test.ts` (7 cases)
- [x] **P11** `benches/ssim_residual.mjs` (AC6 G4-6 gate)
- [x] **P12** AVX2 + SSE2 dispatch (carry-over) — scalar p99 = 4.215 ms ≤ 15 ms budget, NOT required
- [x] **P13** ADR-019 main + ADR-018 §2.6 docs sync
- [x] **P15** decision lock default **(a)** — `meanSsim?` plumbed across 7 surfaces
- [x] **P16** decision lock default **(b)** — `focused_element` rectSource DROPPED, 6 surfaces revised

## Decision locks (Round 0)

[...]

PR #320, opened 2026-05-16
docs(adr-019-stage-4): persist deferred P2 sweep from PR #318 reviews

Harusame64:
## Summary

Persist the 4 P2 findings from PR #318's post-merge external Opus review (Round 1 + Round 2) into a new §7 entry on \`docs/adr-019-stage-4-followups.md\` so they survive context compact and are recoverable by future v1.7.x patch work.

Per CLAUDE.md §9 (\"残件は memory ではなく docs/\") — converting the 4 todos from the prior session's review-log into permanent docs.

## What's in scope (docs only — no production code change)

- §7.1 keyboard.ts Stage 4 **integration test gap** — current \`keyboard-type-stage4.test.ts\` replicates the gate contract rather than exercising \`typeHandler\` BG verify failover end-to-end
- §7.2 Pre-frame \`captureFrame\` **overhead bench** (R4 commitment) — sub-plan §6 R4 committed to a regression gate; no bench file exists
- §7.3 \`moveTo\` → \`captureFrame\` **hover-render race measurement bench** — bound the ~5-30 ms gap's susceptibility to in-flight hover-state transitions
- §7.4 \`cropRawFrame === null\` **branch unit test** — defensive degrade path covered in code but not in local-repaint-orchestrator.test.ts

## Why now

The 4 P2s were tracked across the prior session's Agent-tool reviews (not in GitHub PR review API → invisible to git history). Without persisting to docs, the next context compact would lose them. CLAUDE.md §9 explicitly requires todo / backlog / pending fixes to live in docs not memory.

## Why not block v1.7.0

All 4 are P2 (not load-bearing). Stage 4 algorithm + production wiring shipped correctly per PR #318. Total scope ~0.5 day if bundled into the first v1.7.x patch that touches Stage 4 production code.

## Test plan

- [x] No production code change (docs only, additive new section)
- [x] Existing §1-6 unchanged; new §7 inserted before References (now §8)
- [x] Each P2 has file:line citation + suggested fix + effort/priority estimate

### DIFF
--- a/docs/adr-019-stage-4-followups.md
+++ b/docs/adr-019-stage-4-followups.md
@@ -157,7 +157,123 @@ R6 (background animation / unstable) explicitly seen and handled in §2.3 cycle 
 
 ---
 
-## 7. References
+## 7. Deferred P2 sweep (post-impl PR #318)
+
+Four P2 findings from the post-merge external Opus review of PR #318 were
+deemed defer-eligible and are persisted here for v1.7.x patch follow-up
+(CLAUDE.md §9 — residuals in docs/, not memory). None of these are P1 / load-
+bearing; the Stage 4 algorithm + production wiring are correct as shipped.
+
+### 7.1 (Opus Round 1 P2-1) `keyboard.ts` Stage 4 integration test gap
+
+**Site**: `tests/unit/keyboard-type-stage4.test.ts` (4 cases, all passing).
+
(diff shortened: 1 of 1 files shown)


Harusame64:
@codex review

Round 1 fix applied in 6fcb60a — the §7.5 arithmetic was identical to Opus's finding and is now resolved (table with explicit 0.5+0.5+0.5+0.25=1.75 day sum).

Issue #327, opened 2026-05-16
v1.7.0 release blocked — Stage 5 dogfood degrade observations (7 items A-G)


Harusame64:
## Summary

v1.7.0 release is blocked by 7 dogfood degrades / regression candidates observed during Stage 5 (PR #326) pre-PR smoke on 2026-05-17. Three items (B/C/D) are flagged as high priority because they look like daily-use regressions that would erode user trust if shipped. This issue tracks all 7 items, narrows PR #326 to lint fix only, and coordinates per-item investigation.

## Observed (2026-05-17)

| ID | Observation | Hypothesis | Priority |
|----|-------------|------------|----------|
| A  | Stage 5 returns `source: dxgi_dirty_rect_unavailable`, `motion: indeterminate`, `framesSampled: 0` on every call | vision-gpu's `DirtyRectRouter` (`src/tools/desktop-register.ts:384`) holds the exclusive `outputIndex 0` subscription, so Stage 5's second concurrent subscription is rejected with `NotCurrentlyAvailable` (documented coexistence per `src/engine/any-change.ts:86-91`, sub-plan §2.6) | medium |
| B  | Back-to-back calls hold `totalElapsedMs ~50ms` constant — no cache amortisation | Cache `unavailable` marker not retained / `sweepStale` bug / `cache.invalidate(0)` called from another path? Violates sub-plan §2.4 R3 (cache hit p99 <1ms) | **high** |
| C  | `desktop_act` returns `executor: "mouse"` despite `capabilities.preferredExecutors: ["uia"]` | UIA click likely failing → silent mouse fallback. Daily-use regression candidate | **high (regression candidate)** |
| D  | A single Notepad text-area click emits `diff: ["modal_appeared", "entity_appeared"]` every time | `findBlockingModal` / `isModalCandidate` false positive — `role: "unknown"` status-bar / app-shell entity misclassified as modal? Breaks LLM recovery inference | **high (regression candidate)** |
| E  | `desktop_act action=type` returns `executor_failed` | Notepad text editor may not expose UIA `ValuePattern` (by design, WM_CHAR is the canonical path) OR `setValue` executor wire gap | medium |
| F  | Lease expires faster than expected during normal Claude Code use | TTL vs Claude Code round-trip wallclock mismatch | low |
| G  | `executor_failed` response carries no `if_unexpected` / `suggest` recovery hint | Documented in tool description but absent from result envelope | low |

[...]

PR #329, opened 2026-05-17
fix(desktop_act): attach if_unexpected on executor_failed (#327 item G)

Harusame64:
## Summary

Issue #327 item G: `desktop_act` returns `reason: "executor_failed"` with no `if_unexpected` / `suggest` recovery hint, despite the tool description advertising one. The agent triaging #327 ranked this as the second-cheapest fix with meaningful LLM-recovery improvement.

This is a **tactical local fix** per `project_path_class_refactor_pending` memory — the broader advisory ↔ runtime drift class (G + C + D + E shared pattern) is deferred to the path-class refactor epic, where the structural fix (unified `Result<Ok, Err>` failure path) will land. Here we just close the executor_failed hole.

## Root cause (three-layer drift)

1. `GuardedTouchLoop.touch` returns `{ok:false, reason:"executor_failed"}` at `src/engine/world-graph/guarded-touch.ts:315-319` — a **return**, not a throw.
2. The L5 wrapper `makeCommitWrapper` (`src/tools/_envelope.ts:2904-3003`) only fires `buildFailureEnvelope` (which injects typed `if_unexpected`) on (a) lease validation failure or (b) handler throw. Handler-return failures skip the typed-injection path entirely.
3. `src/tools/_errors.ts` SUGGESTS dict had **no** `ExecutorFailed` entry (zero hits across 60+ codes). Even if the path fired there was nothing to surface.

## Diff

| File | Change |
|---|---|
| `src/tools/_errors.ts` | New `ExecutorFailed` SUGGESTS entry (4 recovery hints derived from #327 dogfood evidence) |
| `src/tools/desktop-register.ts` | Import `getSuggestsForCode`, export `desktopActRawHandler`, attach `if_unexpected` after `facade.touch()` when `result.ok===false && result.reason==="executor_failed"` only |
| `tests/unit/desktop-register.test.ts` | +3 tests (executor_failed attaches, modal_blocking does NOT, success does NOT) |

## Scope discipline

Other failure reasons (`modal_blocking` / `lease_expired` / `entity_outside_viewport` / etc.) have the same structural drift — they return through the handler-return path that bypasses `buildFailureEnvelope`. **Those are intentionally NOT touched** by this PR. The path-class refactor will fix them all in one structural pass; piecemeal local attaches per-reason is the wrong investment.

The new test `does NOT attach if_unexpected when touch fails with a different reason (e.g. modal_blocking) — scope pin` mechanically enforces this scope: if a future PR widens the attach to other reasons, this test must be updated explicitly.

## Recovery-hint content

The four entries derive from #327 dogfood evidence:

1. `action='click'` → `mouse_click({clickAt})` (covers UIA InvokePattern missing — #327 item C symptom)
2. `action='type'` / `action='setValue'` → `keyboard({action:'type', text})` (covers UIA ValuePattern missing / PowerShell locator filter mismatch — #327 item E symptom)
3. `click_element({name|automationId})` (different UIA path, may succeed when desktop_act's threw)
4. Re-run `desktop_discover` (stale-locator case)

## Test plan

- [x] `npm run build` clean
- [x] `npx vitest run tests/unit/desktop-register.test.ts` → 24/24 pass (3 new)
- [x] Full `npm run test:capture` → 3495 pass / 0 fail / 32 skip / 5 todo
- [ ] Codex review (`@codex review`, §3.3 Step 2 — production code change)

Closes part of #327 (item G).

### DIFF
--- a/src/tools/_errors.ts
+++ b/src/tools/_errors.ts
@@ -273,6 +273,20 @@ const SUGGESTS: Record<string, string[]> = {
     "Try click_element + keyboard({action:'type'}) manually",
     "Check context.attempts for per-channel error codes",
   ],
+  // Issue #327 item G: `desktop_act` returned `reason: "executor_failed"` —
+  // the GuardedTouchLoop selected an executor and the executor threw
+  // (`guarded-touch.ts:315-319`). Most common dogfood causes (issue #327):
+  // (C) UIA Edit/Document control without InvokePattern, so click fell
+  // through then mouse-fallback was also skipped, and (E) UIA Edit without
+  // a usable name/automationId for the `setValue` PowerShell locator filter.
+  // The hints below point to the alternate channels for each action,
+  // matching the wiring the dogfood confirmed actually works.
+  ExecutorFailed: [
+    "For action='click', fall back to mouse_click({clickAt}) using the entity rect center from desktop_discover — common when UIA InvokePattern is missing on the control",
+    "For action='type' or action='setValue', fall back to keyboard({action:'type', text}) after focusing the target — common when UIA ValuePattern is missing or the locator filter cannot re-find the entity",
+    "If the entity has a stable name or automationId, try click_element({name|automationId}) — uses a different UIA path than desktop_act and may succeed where this executor threw",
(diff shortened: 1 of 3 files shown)


Harusame64:
@codex review

PR #330, opened 2026-05-17
fix(desktop_act): UIA setValue → keyboardTypeBg fallback ladder (#327 item E)

Harusame64:
## Summary

Issue #327 item E: `desktop_act({action:'type'|'setValue'})` against a UIA entity returns `executor_failed` when `makeSetElementValueScript`'s PowerShell name filter cannot re-find the entity. The canonical dogfood case is Notepad's RichEditD2DPT main edit area (empty / unstable name) — a path Claude Code will hit constantly, and today it fails outright.

This PR adds a tactical fallback ladder inside the UIA route: when `uiaSetValue` throws, fall back to a new `keyboardTypeBg` dep that posts WM_CHAR to the focused child via the same primitive `terminalSend` already uses (`bg-input.ts::postCharsToHwnd`). The combined-failure path surfaces both diagnostics in one error so the LLM (and PR #329's `if_unexpected.try_next`) sees the full ladder context.

## Why this approach

| | UIA setValue (today) | UIA setValue + keyboardTypeBg (this PR) |
|---|---|---|
| Notepad text input | ❌ `executor_failed` (PowerShell `-like '*…*'` filter rejects empty name) | ✅ Posts WM_CHAR to RichEditD2DPT child |
| Chromium / WT-XAML | ❌ Same failure mode | ❌ Same — `canInjectAtTarget` returns unsupported, joint error carries both diagnostics |
| Already-working UIA-named text fields | ✅ Works | ✅ Same (uiaSetValue succeeds, fallback not invoked) |
| Click / invoke action | (no fallback ladder for these) | (unchanged — scope pin test enforces) |

`keyboardTypeBg` uses **`canInjectAtTarget`** (not `canInjectViaPostMessage`) so the BG class check resolves to the *focused child* — Notepad's RichEditD2DPT child rather than the top-level "Notepad" window. This mirrors `_focus.ts`'s use of the same primitive and is the correct semantic for "send keys to the active edit control".

## Diff

| File | Change |
|---|---|
| `src/engine/world-graph/types.ts:25` | `ExecutorKind` extended with `"keyboard"`. Documented as internal-fallback-only — NOT added to `UiAffordance.executors` or `unsupportedExecutors` (those remain the 4-executor union). Promotion to first-class is a path-class refactor question. |
| `src/tools/desktop-executor.ts` | `ExecutorDeps.keyboardTypeBg(windowTitle, text)` added, real impl in `getSharedRealDeps` (mirrors `terminalSend` but uses `canInjectAtTarget`), UIA route at L154+ wraps `uiaSetValue` in try/catch and falls back to `keyboardTypeBg`. Combined-failure throw carries both diagnostics. |
| `tests/unit/desktop-executor.test.ts` | `mockDeps` updated + 5 new tests (success regression pin, `type` fallback, `setValue` fallback, combined failure, click-action scope pin). |

## Scope discipline

Per `project_path_class_refactor_pending` memory and the user's "リファクタリングのタイミングでやる" directive, **this PR does NOT**:

- Add `"keyboard"` to `AFFORDANCE_MAP.type.executors` (advertised capability stays the same — promotion is a refactor question).
- Add `"keyboard"` to `unsupportedExecutors` union (keeps the LLM-facing capabilities surface stable).
- Touch the CDP `cdpFill` or terminal `terminalSend` paths.
- Add a similar fallback ladder to the UIA click path (that's item C's territory, separate PR).

The new test `text=undefined (no type/setValue): no keyboardTypeBg call (scope pin — click/invoke unchanged)` mechanically prevents future widening.

## Test plan

- [x] `npm run build` clean
- [x] `npx vitest run tests/unit/desktop-executor.test.ts` → 42/42 pass (5 new)
- [x] Full `npm run test:capture` → 3499 pass / 1 pre-existing flake (`benchmark-gates.test.ts` ReplayBackend cold-vs-warm — same flake observed in other recent PRs, not introduced here)
- [ ] Codex review (`@codex review`, §3.3 Step 2 — production code change)

Closes part of #327 (item E).

### DIFF
--- a/src/engine/world-graph/types.ts
+++ b/src/engine/world-graph/types.ts
@@ -22,7 +22,18 @@ export interface EntityLocator {
 }
 export type AffordanceVerb = "invoke" | "click" | "type" | "select" | "scrollTo" | "read";
 export type EntitySourceKind = "uia" | "cdp" | "win32" | "ocr" | "som" | "visual_gpu" | "terminal" | "inferred";
-export type ExecutorKind = "uia" | "cdp" | "terminal" | "mouse";
+/**
+ * `"keyboard"` is a sub-executor used as a fallback from the UIA `setValue` route
+ * when `uiaSetValue` throws (e.g. Notepad's RichEditD2DPT exposes `ValuePattern` but
+ * `makeSetElementValueScript`'s `name -like '*…*'` locator filter cannot reach the
+ * entity — issue #327 item E). The fallback posts WM_CHAR to the focused child via
+ * `bg-input.ts::postCharsToHwnd` (same primitive `terminalSend` uses). It is NOT
+ * a primary executor advertised in `UiAffordance.executors` or `unsupportedExecutors`
+ * — those remain the 4-executor union (`uia | cdp | terminal | mouse`). The
+ * path-class refactor epic may promote it to a first-class executor once the
+ * capability registry consolidates the ladder.
+ */
+export type ExecutorKind = "uia" | "cdp" | "terminal" | "mouse" | "keyboard";
 
 export interface UiAffordance {
   verb: AffordanceVerb;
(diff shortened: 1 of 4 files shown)


Harusame64:
@codex review

Harusame64:
@codex review

PR #332, opened 2026-05-17
fix(desktop_act): surface UIA→mouse downgrade marker on TouchResult (#327 item C)

Harusame64:
## Summary

Issue #327 item C: `desktop_act` against a UIA-sourced entity could return `executor: "mouse"` despite `capabilities.preferredExecutors: ["uia"]` because the UIA route at `desktop-executor.ts:158-172` silently caught `uiaClick` failure and fell through to mouseClick. From the LLM's point of view this was indistinguishable from "UIA was never the chosen route", so capability advisories drifted out of sync with runtime evidence.

This PR adds an **observability marker** without changing the silent-fallback behaviour itself (per agent option 1, low rollout risk). The TouchResult now carries `downgrade: { from: "uia", reason }` when the fallback happened, so the LLM can distinguish "advised executor was tried and failed" from "advised executor was not the chosen route".

### Before
```jsonc
// silent UIA→mouse fallback envelope
{ "ok": true, "executor": "mouse", "diff": [...] }
```

### After
```jsonc
// successful UIA route (unchanged — no downgrade marker)
{ "ok": true, "executor": "uia", "diff": [...] }

// silent UIA→mouse fallback (new marker)
{ "ok": true, "executor": "mouse", "diff": [...],
  "downgrade": { "from": "uia", "reason": "InvokePatternNotSupported" } }
```

The `downgrade` field is **absent** (key omitted, not `undefined`) on the no-fallback path so JSON envelopes stay byte-equal to today on the happy path.

## Diff

| File | Change |
|---|---|
| `src/engine/world-graph/types.ts` | New `ExecutorOutcome` interface — rich return shape signalling silent fallback |
| `src/engine/world-graph/guarded-touch.ts` | `TouchEnvironment.execute` return widened to union; `TouchResult` success variant gains `downgrade?`; `touch()` normalises both shapes and omits the field when no downgrade |
| `src/engine/world-graph/session-registry.ts` | `ExecutorFn` type matches widened return |
| `src/tools/desktop-executor.ts` | UIA route silent fallback returns `{ kind: "mouse", downgrade: { from: "uia", reason } }`; no-rect fallback path attaches `cause` to the thrown Error |
| `tests/unit/desktop-executor.test.ts` | 3 existing UIA-fallback assertions updated + 4 new tests in a new describe block (success regression pin, downgrade shape pin, no-rect-throw without downgrade, non-Error throwable handling) |
| `tests/unit/guarded-touch.test.ts` | 2 new tests injecting `execute()` to return `ExecutorOutcome` directly and asserting `TouchResult.downgrade` propagation + an "executor:'uia' has no downgrade" pin |

## Scope discipline (per `project_path_class_refactor_pending`)

- **No change to advertised capabilities, `AFFORDANCE_MAP`, or `unsupportedExecutors`** — those remain the 4-executor union.
- **No change to other silent-fallback paths**: the UIA setValue ladder added in PR #330 is already explicit (different shape); CDP `cdpFill` and terminal `terminalSend` do not fall through silently — when they fail, `executor_failed` is surfaced directly.
- The path-class refactor epic will fold this marker into the unified `CapabilityRegistry` failure-envelope surface alongside G's `if_unexpected` injection and D's `isChromeControlType` helper.

## Behaviour pin (back-compat)

The bare-string `ExecutorKind` return remains valid — every existing test that injects `execute: async () => "mouse"` etc. still passes without modification. Only the new UIA-fallback path emits the rich shape, and only the new tests assert against it.

## Test plan

- [x] `npm run build` clean
- [x] `npx eslint` clean on all 6 changed files (no `preserve-caught-error` regression)
- [x] `npx vitest run` desktop-executor + guarded-touch + session-registry → 107 / 107 pass (3 existing updated, 6 new)
- [x] Full `npm run test:capture` → 3516 pass / **0 fail** / 32 skip / 5 todo
- [ ] Codex review (`@codex review`, §3.3 Step 2 — production code change with API surface widening)

Closes part of #327 (item C).

### DIFF
--- a/src/engine/world-graph/guarded-touch.ts
+++ b/src/engine/world-graph/guarded-touch.ts
@@ -1,4 +1,4 @@
-import type { UiEntity, EntityLease, ExecutorKind, UiAffordance } from "./types.js";
+import type { UiEntity, EntityLease, ExecutorKind, ExecutorOutcome, UiAffordance } from "./types.js";
 import type { LeaseStore } from "./lease-store.js";
 import type { VisualMotionObservation } from "../../tools/_input-pipeline.js";
 import { isChromeControlType } from "./session-registry.js";
@@ -62,6 +62,17 @@ export type TouchResult =
        * unaffected (additive — sub-plan §2.5 + CLAUDE.md §3.2 carry-over).
        */
       observation?: VisualMotionObservation;
+      /**
+       * Issue #327 item C: surfaced when the executor silently fell back from a
+       * higher-priority executor (e.g. UIA InvokePattern threw and the mouse
+       * rect-center fallback succeeded). Without this marker the LLM sees
+       * `capabilities.preferredExecutors: ["uia"]` ↔ `executor: "mouse"` and
+       * cannot distinguish "UIA was tried and failed" from "UIA was not the
+       * chosen route". The `from` field names the executor that was originally
+       * selected; `reason` is the underlying error message. Absent (= field
+       * undefined) when no fallback happened.
+       */
+      downgrade?: ExecutorOutcome["downgrade"];
     }
   | {
       ok: false;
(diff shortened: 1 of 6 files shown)


Harusame64:
@codex review

PR #351, opened 2026-05-18
docs(instructions): route terminal-output reads to terminal(action='read')

Harusame64:
## Summary

LLMs were defaulting to `screenshot(detail='text')` when asked to read terminal output, even though `terminal(action='read', ...)` exists and the tool description documents it. Reason: the **server-instructions Observation priority list** never named the terminal read action, and `terminal` reads as "send commands" by surface name, so the model fell back to the documented observation ladder (which only mentions desktop_state / desktop_discover / screenshot).

After being told "use desktop_read / terminal(action='read')" mid-conversation, models switch immediately — confirming this is a discovery problem, not a capability problem.

Fix: insert `terminal(action='read', windowTitle=...)` as item 3 in the Observation priority list, just before `screenshot`, with an explicit "prefer over screenshot for terminal content" anchor.

## Notes

- Server instructions are truncated by Claude Code from 4082 → 2048 chars on connect. Verified locally that the new entry sits inside the 2048-char budget (cutoff lands inside the later Attention-signal section).
- No code change beyond the instruction string — no tests touch this string.

## Test plan
- [ ] `tsc --noEmit` passes (done locally)
- [ ] After reload, prompt the model to read terminal output without mentioning `terminal(action='read')` — verify it picks `terminal(action='read')` instead of `screenshot`


Harusame64:
@codex review

PR #376, opened 2026-05-20
feat(_errors): ADR-021 P2-0 — error-model + flat-failure presenter family

Harusame64:
## What

ADR-021 **Phase 2 PR-P2-0** — the prerequisite for migrating `failWith` (175 callsites) to a converter-based failure path. Establishes design **B′: error-model-as-SSOT + presenter family**.

- **`src/errors/typed-errors.ts`** — `ToolFailureError` model (+ `ToolFailurePayload`) carrying optional `toolName` / `displayMessage` / `suggest` / `context` / `rootExtras`; `name === code` (same convention as `CodedHandlerError`).
- **`src/tools/_errors.ts`** —
  - `errorFromMessage(message, toolName, context): ToolFailureError` factory (**OQ-7(c)**): the one place that runs `classify(message)` and splits `context` into the `ROOT_HOISTED_KEYS` half (`_perceptionForPost` / `_richForPost` / `hints` → `rootExtras`) vs the nested half.
  - `toToolFailure(err): ToolFailure` presenter: renders the flat `{ ok:false, code, error, [suggest], [context], ...rootExtras }` shape, **byte-for-byte equal to today's `failWith`**.
- **`failWith` and `toFailureEnvelope` are both untouched** in this PR. The new functions are dormant scaffolding; the P2-2 flip is pre-validated by a `failWith`-equivalence safety net.

## Why B′ (not the plan-literal "extend `toFailureEnvelope`")

The flat `ToolFailure` shape (`code`/`error`/`suggest`, what 175 `failWith` callsites emit) and the envelope/compat-raw shapes (`if_unexpected`/`_version`, what `toFailureEnvelope` emits) are **structurally disjoint, permanent output families** serving disjoint consumers (§2.2 keeps the `ToolFailure` shape 100% backward-compatible). Extending one converter to emit both would give it a **4-way union return switched by an option flag** — the boolean-return-type anti-pattern, itself a "繋ぐと破綻" surface.

Cross-ecosystem research converged on the same principle: the error is **one pure typed value (the SSOT)** and rendering it to a wire shape is a **separate, narrow, per-format concern** — RFC 9457 problem-detail object, Effect `Data.TaggedError`, neverthrow `mapErr`/`match` at boundaries, Rust error-enum vs `Display`/`Serialize`. So: one model, two narrow presenters. This makes the typed boundary **stronger** than a polymorphic converter (technical north star: write drift out at the type level), and leaves the envelope-family snapshot trivially bit-equal because that path is not touched. Same scope; downstream PR structure (P2-2 flip, P2-3 codemod) unchanged.

## MCP protocol

Protocol-neutral. All tool failures serialize to a text content block (`{content:[{type:"text",text:"<json>"}]}`); MCP never inspects the JSON internals. B′ changes only **construction**, not the resulting JSON (bit-equal) nor the wrapping. Bonus: centralising failure construction makes a future `isError` / `structuredContent` adoption a one-place change.

## Tests

`tests/unit/path-class-contract/to-tool-failure-payload.test.ts` (18 cases, all green):
1. **Matrix** — presenter flat shape over the 4 payload axes + suggest presence + key-order, frozen literals **decoupled from `classify()`/SUGGESTS** (PR #373 Codex P2 discipline).
2. **Factory** — `errorFromMessage` code classification + context split.
3. **Equivalence (safety net)** — `fail(toToolFailure(errorFromMessage(m,t,c)))` is **byte-for-byte equal** to `failWith(new Error(m), t, c)` across representative inputs (incl. root-hoisted keys + empty message). This is the invariant P2-2 must preserve.

`npm run build` clean.

## Test suite

`3 failed | 3783 passed | 32 skipped` — the **3 failures are pre-existing & environmental, unrelated to this diff**:
- `tests/unit/benchmark-gates.test.ts > Gate 2 — idle cost` (timing measurement, flaky under parallel load)
- `tests/e2e/keyboard-bg-verification.test.ts` (e2e, needs real conhost)
- `tests/e2e/mouse-verify-delivery.test.ts` (e2e, needs real mouse / failsafe-collateral)

The new functions are dormant (not wired into any runtime path) and touch none of keyboard/mouse/benchmark/visual code. The bit-equal envelope-family safety net (`to-failure-envelope-shape-snapshot.test.ts`) **passes**, confirming the envelope path is intact.

## Plan

`desktop-touch-mcp-internal@4ae0c26:docs/adr-021-result-migration-drift-prevention-plan.md` §3.3.2 PR-P2-0 — [permalink](https://github.com/Harusame64/desktop-touch-mcp-internal/blob/4ae0c262377fb0d6f36bf584be919ad36800a793/docs/adr-021-result-migration-drift-prevention-plan.md) (private repo; logged-in reviewers only).

### DIFF
--- a/src/errors/typed-errors.ts
+++ b/src/errors/typed-errors.ts
@@ -118,6 +118,10 @@ export class ToolFailureError extends HandlerError {
   readonly rootExtras?: Record<string, unknown>;
 
   constructor(code: string, payload?: ToolFailurePayload, options?: ErrorOptions) {
+    // `??` (not `||`) is load-bearing: an empty `displayMessage` ("") must be
+    // preserved, not coalesced to `code`, so a thrown empty message stays
+    // bit-equal with `failWith`. The presenter's `err.displayMessage ?? code`
+    // relies on the same `??` semantics.
     super(payload?.displayMessage ?? code, options);
     this.name = code;
     this.toolName = payload?.toolName;
--- a/src/errors/typed-errors.ts
+++ b/src/errors/typed-errors.ts
@@ -131,8 +131,10 @@ export class ToolFailureError extends HandlerError {
     // long-lived value (carried through `Result.err`), so holding the shared
     // reference would let any downstream mutation (sort/push while enriching or
     // logging) corrupt global suggestion state across requests. `failWith` was
-    // safe only because it serialised the array immediately; the typed model is
-    // not, so it owns an independent copy (Round 1 Codex P2). `context` /
+    // safe only because it spread the reference into a throwaway object and
+    // `JSON.stringify`d it without ever retaining a live reference; the typed
+    // model retains one, so it owns an independent copy (Round 1 Codex P2).
+    // `context` /
     // `rootExtras` are caller-owned fresh containers (same as failWith) — no
     // shared global state, so no clone needed.
     this.suggest = payload?.suggest ? [...payload.suggest] : undefined;
--- a/src/errors/typed-errors.ts
+++ b/src/errors/typed-errors.ts
@@ -60,6 +60,74 @@ export class CodedHandlerError extends HandlerError {
   }
 }
 
+/**
+ * Optional payload carried by {@link ToolFailureError} — the fields the flat
+ * `ToolFailure` presenter (`toToolFailure` in `src/tools/_errors.ts`) renders.
+ *
+ * ADR-021 Phase 2 PR-P2-0 (B′: error-model-as-SSOT + presenter family). The
+ * typed error is the single source of truth for a handler failure (≒ RFC 9457
+ * problem-detail object / Effect `Data.TaggedError` / a Rust error enum);
+ * rendering it into the flat `{ok:false, code, error, ...}` wire shape is a
+ * separate concern done by a narrow presenter, NOT by hand-built object
+ * literals (Phase 4 ESLint `no-tool-failure-shape-direct-construct` enforces
+ * this). This is why the envelope family converter (`toFailureEnvelope`) stays
+ * untouched: the two shapes are different render targets of one error model.
+ *
+ * Field → rendered output (`toToolFailure`, bit-equal with today's `failWith`):
+ *   - `toolName` + `displayMessage` → `error: "${toolName} failed: ${displayMessage}"`
+ *   - `suggest`    → `suggest` (omitted when empty — matches `failWith`)
+ *   - `context`    → nested `context` (the non-hoisted half of `failWith`'s context arg)
+ *   - `rootExtras` → spread onto the failure root (the ROOT_HOISTED_KEYS half:
+ *     `_perceptionForPost` / `_richForPost` / `hints`, read by `_post.ts`)
+ *
+ * The plan §3.3.2 listed the suggest field as `suggestOverride`; under B′ the
+ * model carries an already-resolved `suggest` array — the `errorFromMessage`
+ * factory fills it from `classify(message)`, and an explicit caller may override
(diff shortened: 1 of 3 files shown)


Harusame64:
@codex review

Harusame64:
## Round 1 findings applied (Opus phase-boundary review)

- **P1-1 (blocker, fixed)** — `errorFromMessage` took `message: string` and skipped `failWith`'s `instanceof Error ? .message : String(err)` normalization. Production callsites pass non-Error values (bare strings, `??`-coalesced strings, raw caught values), so the planned P2-2 flip would NOT be bit-equal for those — and the equivalence test masked it by only feeding `new Error(message)`. **Root-cause fix**: factory now takes `err: unknown` and owns the normalization (single SSOT for thrown→message); the equivalence suite now feeds the SAME raw value (Error / bare string / object / undefined / null / number) to both `failWith` and the model path.
- **P2-2 (fixed)** — reconciled the "8 combos over 4 axes + edge pins" matrix wording; empty-message edge kept first-class.
- **P3-1 (fixed)** — documented that `??` (not `||`) is load-bearing for the empty-message contract.
- **P2-1 (fixed in plan repo)** — plan SSOT diverged from B′ in 4 locations; reconciled in `desktop-touch-mcp-internal#4` (Round 5 note).

Commit `ee4e6ed`. Tests: 24/24 in `to-tool-failure-payload.test.ts`; build clean. The 3 suite failures (benchmark-gates Gate 2 timing + 2 e2e) remain pre-existing & environmental, unrelated to this diff.

Harusame64:
@codex review

Round 1 Codex P2 (suggest array aliasing of shared SUGGESTS dict) fixed in 87b0d54 — ToolFailureError now clones the array at construction + regression test. Please re-review the cumulative diff (ee4e6ed Opus P1-1/P2-2/P3-1 + 87b0d54).

Harusame64:
## Round 2

- **Opus Round 2: APPROVED (P1 = 0).** Verified all Round 1 fixes are completely applied; re-ran the bit-equality mental simulation across all 10 input cases (Error / string / object / undefined / null / number / empty-message, incl. key order, suggest/context omission, empty-rootExtras spread) — holds. Clone is order-neutral and correct. Design B′ sound & in-scope. Only an optional P3 wording nit, fixed in `d74dc98`.
- **Codex Round 1 P2 (suggest aliasing): fixed in `87b0d54`** — `ToolFailureError` clones the array at construction + regression test (`getSuggestsForCode` + later failures unaffected after mutating `model.suggest`).

Final commit `d74dc98`. Re-triggering Codex on the final state below.

Harusame64:
@codex review

Harusame64:
### Codex P2 (suggest aliasing) — resolved

Codex's only finding ("Clone suggest list before attaching it to typed error", `_errors.ts` `errorFromMessage`) is fixed in `87b0d54`, exactly as suggested — the clone happens **at construction time** in `ToolFailureError`'s constructor (`this.suggest = payload?.suggest ? [...payload.suggest] : undefined`), so the long-lived typed model never holds the shared `SUGGESTS` array reference. A regression test pins the isolation (mutating `model.suggest` leaves `getSuggestsForCode` and later failures untouched). Opus Round 2 independently verified the clone is correct and order-neutral (bit-equality preserved).

(Re-triggers of `@codex review` on the later commits did not produce a fresh pass — Codex appears to dedup re-reviews on the same PR. The finding itself is addressed.)

**Review state: Opus APPROVED (P1=0, 2 rounds); Codex's sole finding resolved + regression-pinned + Opus-verified.** Final commit `d74dc98`.

PR #379, opened 2026-05-21
refactor(tools): ADR-021 P2-3a — un-deprecate failWith (canonical) + sweep (C) hand-built failure literals via failCode

Harusame64:
## What

ADR-021 PR-P2-3 **part a** (Option B re-scope, OQ-1 RE-DECISION). Two things, both staying inside the B′ presenter family:

### 1. Un-deprecate `failWith`, reframe as the canonical failure window
OQ-1(a) "delete failWith" was decided **before B′ was finalized**. B′ + PR-P2-2 already made `failWith` a thin wrapper over the presenter (the typed error model is the SSOT), so the dual-failure system is gone and removal would only churn the ~171 already-sanctioned callsites. Per **OQ-1 RE-DECISION (Round 6)** `failWith` is **kept** as the canonical single window.
- Remove the `@deprecated` JSDoc added in PR-P2-2; reframe its doc + the 3 cross-ref docs (`typed-errors.ts` / `result.ts` / `_envelope.ts`) to the Option B framing. Behavior unchanged (frozen golden 10/10 green).

### 2. Sweep the (C) code-bearing hand-built failure literals
The real drift surface (`§1.3 row C`): handlers hand-building `fail({ ok:false, code, error, suggest, context })` and bypassing the sanctioned converter. New helper:

```
failCode(code, error, { suggest?, context?, rootExtras? })
```

— the **explicit-code** sibling of `failWith` (classify-derived) and `failArgs` (fixed InvalidArgs). Routes through `toToolFailure(new ToolFailureError(...))`, emits `error` verbatim (no `<tool> failed:` prefix), so the wire shape is **bit-equal** with the literal it replaces.

Swept 9 (C) sites → `failCode`, bit-equal:
- `browser.ts` ×5 (ElementNotInViewport ×2, NavigateFailed, browser_search runtime-computed code, BrowserSearchNoResults)
- `terminal.ts` ×3 (TerminalTextPatternUnavailable, InvalidArgs ×2)
- `wait-until.ts` ×1 (WaitTimeout)

## Tests / gates
- New `failcode-sweep.test.ts`: `failCode` shape contract (frozen) + byte-for-byte pins vs the pre-migration literals (snapshot-first bit-equal proof).
- `issue-211-classify-branch-producer-pin.test.ts`: taught the structural guard that `failCode("<code>")` is a sanctioned producer form (was only `code:"X"` / `failWith`) — otherwise the sweep makes TerminalTextPatternUnavailable / WaitTimeout look like dead classify codes.
- Regenerated `failwith-callsite-shapes.json` (line shifts only; totalCallsites still 171 — no failWith call added/removed). `check:failwith-fixtures` green.
- Unit suite green (pre-existing benchmark-gate + PERSIST-14 timing flakes only, pass in isolation); tsc + lint clean.

## Scope note
This is **part a** (canonical-ize + (C) bit-equal sweep). **Part b** = the (D) code-less literals (`scroll-read` / `pin` / `scroll-capture` / `desktop-register` / `macro`) with **OQ-9 (c) per-site codes** (intentional shape change: adds recovery-hint codes), kept separate because it changes wire shape. The (E) failure-as-success class (`ok()`-wrapped failures) is carried over to OQ-10 (not regex-enumerable).

Plan: `desktop-touch-mcp-internal#7` (Round 6 OQ-1 RE-DECISION) §3.3.2 step 4

### DIFF
--- a/src/errors/typed-errors.ts
+++ b/src/errors/typed-errors.ts
@@ -96,11 +96,14 @@ export interface ToolFailurePayload {
 
 /**
  * Canonical typed model for a handler failure that renders to the flat
- * `ToolFailure` shape — the `failWith` family (171 migratable callsites under
- * `src/tools/**`; 176 grep hits minus the 5 self-references in `_errors.ts`,
- * machine-counted by scripts/extract-failwith-shape-fixtures.mjs). PR-P2-2 made
- * `failWith` a thin wrapper over this model; PR-P2-3 migrates the callsites and
- * PR-P2-4 removes the wrapper (OQ-1(a) full removal).
+ * `ToolFailure` shape. PR-P2-2 made `failWith` a thin wrapper over this model
+ * (`failWith = fail(toToolFailure(errorFromMessage(...)))`), so `failWith` is the
+ * canonical single window for flat failures and this class is its SSOT error
+ * value. ADR-021 OQ-1 RE-DECISION (Round 6): `failWith` is KEPT (not deleted) —
+ * once the wrapper unified the path, removing it would only churn the ~171
+ * already-sanctioned callsites. PR-P2-3 instead routes the remaining hand-built
+ * `{ ok:false, ... }` literals through `failWith`, and Phase 4 ESLint bans new
+ * ones (`no-tool-failure-shape-direct-construct`).
  *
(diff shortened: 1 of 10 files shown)


Harusame64:
@codex review

Harusame64:
@codex review

PR #381, opened 2026-05-21
feat(lint): ADR-021 P4-1 — no-tool-failure-shape-direct-construct ESLint rule (Option B enforcement)

Harusame64:
## What

ADR-021 **Phase 4 PR-P4-1** — the custom ESLint rule `no-tool-failure-shape-direct-construct`. This is the **structural half** of the Option B North Star: it bans hand-building the flat tool-failure wire shape so failures can *only* be emitted through the sanctioned presenter family.

After PR-P2-2/P2-3 swept every existing hand-built literal, this rule makes that sweep **permanent** and closes the R12 "un-deprecate window" (without it, new PRs could re-introduce raw `fail({ok:false,...})` literals).

## The rule

`eslint-rules/no-tool-failure-shape-direct-construct.mjs`:
- **Flags** an `ObjectExpression` with `ok:false` **and** an `error` key when it is the argument to `fail(...)` or `JSON.stringify(...)` (the two wire-emission contexts the swept literals used).
- **Allows**: `failWith` / `failCode` / `failArgs` / `fail(toToolFailure(...))` / `toFailureEnvelope`; `fail(variable)`; success payloads.
- **Does NOT flag** (by design): internal discriminated-union returns (`{ok:false, reason}`, `{ok:false, errorResult}` — no `error` key) and the `(E)` failure-as-success returns (`dock.ts` `return {ok:false, title, error}` consumed by `ok()`) — those are the OQ-10 carry-over (different mechanism, needs semantic detection).
- Wired in `eslint.config.mjs` for `src/tools/**` as **error**, exempting the converters `_errors.ts` / `_envelope.ts`.

## warn vs error (deviation from plan §3.5.1)

The plan said *warn-first → observe → error*. I landed it as **error directly** because:
1. PR-P2-2/P2-3 already swept every violation — `npm run lint` reports **0**, so there is nothing to migrate gradually.
2. A `warn` rule does **not** fail CI (eslint exits 0 on warnings without `--max-warnings`), so it would not actually block new literals — only `error` is real enforcement.
3. Rule correctness is verified up front (RuleTester 14 cases + a live injection check), so the false-positive risk warn-first was meant to surface is already characterized.

(Same shape as the OQ-1 RE-DECISION: the sweep changed the premise the staged rollout assumed.) Flagged for Opus/user; plan §3.5.1 + Phase 4 acceptance sync in a follow-up internal PR.

## Verification
- `tests/unit/path-class-contract/eslint-no-tool-failure-shape.test.ts`: RuleTester, **14 cases** (10 valid sanctioned/non-failure forms + 4 invalid hand-built literals).
- `npm run lint`: **0 errors** (rule clean on current code — sweep complete). Injecting `fail({ok:false,code,error})` reproduces the error; reverting clears it.
- `tsc` clean.

Plan: `desktop-touch-mcp-internal@0c4b960` §3.5.1 (Phase 4)

### DIFF
--- a/eslint-rules/no-tool-failure-shape-direct-construct.mjs
+++ b/eslint-rules/no-tool-failure-shape-direct-construct.mjs
@@ -1,0 +1,98 @@
+/**
+ * ESLint rule: no-tool-failure-shape-direct-construct
+ * — ADR-021 Phase 4 PR-P4-1 (Option B North Star enforcement).
+ *
+ * Bans hand-building the flat tool-failure WIRE shape (`{ ok:false, ..., error,
+ * ... }`) at an emission point — i.e. as the argument to `fail(...)` or
+ * `JSON.stringify(...)`. Every handler failure must instead go through the
+ * single sanctioned presenter family in `src/tools/_errors.ts`:
+ *
+ *   failWith(err, toolName, context?)   — code via classify(message)
+ *   failCode(code, error, { suggest?, context?, rootExtras? })  — explicit code
+ *   failArgs(message, toolName, context?)  — fixed InvalidArgs
+ *   fail(toToolFailure(errorFromMessage(...)))  — the underlying composition
+ *
+ * or the envelope-family `toFailureEnvelope(...)`. This is what makes the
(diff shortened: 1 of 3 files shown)


Harusame64:
@codex review

Harusame64:
@codex review

PR #382, opened 2026-05-21
refactor(macro): ADR-021 Phase 3a — runInnerToolAsResult adapter for inner-step propagation

Harusame64:
## What

ADR-021 **Phase 3a** — the typed adapter `runInnerToolAsResult` for `run_macro` inner-step propagation (drift surface **B**: an inner `ok:false` envelope must halt `stop_on_error: true` like a throw — the F1/F2 silent-success contract).

Extracts the inner-step handler call + content extraction + `ok:false` envelope JSON-parse out of the `run_macro` loop into:

```ts
runInnerToolAsResult(entry, validated): Promise<Result<InnerToolOutcome, InnerToolOutcome>>
```

`run_macro` now consumes the **Result discriminant** (`outcome.ok`) instead of re-deriving step success from a hand-parsed flag. The inner-envelope JSON parse happens **only inside the adapter** — the one place that re-parses an inner envelope.

## Behavior is bit-equal

The step-result push shape (`ok` / `text` / `error` / `code` / `_images` + key order) is unchanged, so the existing end-to-end test (`run-macro-stop-on-error-inner-envelope.test.ts`, 7 cases) passes **unmodified**.

## Tests
- New `run-inner-tool-as-result.test.ts` (6 cases): pins the adapter via the **Result API directly** — `Ok` for `ok:true` / non-JSON first block, `Err` for `ok:false` (carries `code` / `error` / `text` / `images`). This is the revert-detecting Result-API assertion (plan §3.4.4): reverting the adapter's Ok/Err selection or the parse fails it.
- Existing E2E test unchanged → behavior preservation proof.
- Full unit suite green (pre-existing benchmark-gate + PERSIST-14 timing flakes only); tsc + lint clean.

## Scope
- Compile-time closure (making `ToolHandler` itself return `Result`, so F1/F2 silent-success is a *compile* error) stays a **Phase 5 carry-over** — here the boundary is one helper, the runtime contract is test-pinned.
- PR-P3a-2 (audit whether other orchestrators — `desktop_act` retry, `keyboard sequence` — share the run_macro inner-tool pattern) is an optional follow-up; none call inner TOOL_REGISTRY entries the way `run_macro` does.

Plan: `desktop-touch-mcp-internal@beeffca` §3.4 (Phase 3a)

### DIFF
--- a/src/tools/macro.ts
+++ b/src/tools/macro.ts
@@ -3,6 +3,7 @@ import { z } from "zod";
 import { buildDesc } from "./_types.js";
 import type { ToolHandler, ToolResult } from "./_types.js";
 import { failCode } from "./_errors.js";
+import { type Result, Ok, Err } from "../types/result.js";
 import { checkFailsafe } from "../utils/failsafe.js";
 import { assertKeyComboSafe } from "../utils/key-safety.js";
 import {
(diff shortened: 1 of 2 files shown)


Harusame64:
@codex review

PR #387, opened 2026-05-22
feat(terminal): #386 P1 — echo-immune exit-mode sentinel pure helpers + types

Harusame64:
## What & why
#383 anchored single-line `until:{mode:'pattern'}` past the echoed command, but **multiline echo boundaries are undeterminable from the buffer alone (#386)**, so multiline was left to a best-effort full scan (residual self-match). The structural fix is a new **`until:{mode:'exit'}`** that matches a **driver-controlled token whose echo form differs from its output form** — echo-immune by construction for single-line *and* multiline — instead of trying to locate the echo.

This PR is **Phase 1: pure helpers + types only** (no public API surface change yet). The live `until` discriminated-union variant + handler wiring land together in **P2** so `mode:'exit'` is never half-exposed (a P1-only variant would fall through to quiet detection).

## Contents
Pure/exported helpers in `src/tools/terminal.ts` (dead until P2, pinned by unit tests):
- **`buildExitCommand(input, shell, nonce)`** — prologue + input + epilogue. The epilogue prints the token via a **split expression** (`'__DTMCP' "_EXIT_<n>"` / `('__DTMCP'+"_EXIT_<n>")`) so the sent/echoed text never contains the contiguous `__DTMCP_EXIT_<nonce>`; only runtime output assembles it.
- **`parseExitSentinel(slice, nonce, shell)`** — defers until `<token>|<code>` (bash) / `<token>|<code>|<bool>` (powershell) fully renders. PowerShell normalises `$LASTEXITCODE` (native) vs `$?` (cmdlet) with a `$null` prologue clearing stale values.
- **`detectShell(processName)`** — `high` for pwsh/powershell/bash/wsl/cmd; `low` for WindowsTerminal/conhost/OpenSSH/unknown (the SSH/WSL-nesting wall — P2 handler loud-fails low confidence rather than guessing).
- **`isUnsafeForExitMode(input)`** — loud-reject heredoc / PS here-string / trailing continuation / unbalanced quotes (context-tracking scanner so an apostrophe inside `"double quotes"` does not false-trip).
- **`generateExitNonce()`** — per-invocation crypto-random hex.

Types: `CompletionReason += "exited"`; `completion.exitCode?` (exit mode only). **No runtime behaviour change** — `pattern`/`quiet` untouched.

## Tests
21 unit assertions: echo-immunity (sent text never matches the parser), defer-until-rendered, exit-code normalisation (incl. cmdlet `$?` mapping), shell confidence matrix, unsafe-input detection. Full unit suite green (**3621 pass**). `failwith` fixtures regenerated for the terminal.ts line shift.

## Review asks
- Echo-immunity invariant: is there any shell rendering where the **split** epilogue could surface the contiguous `<token>|<code>` in the echo?
- `isUnsafeForExitMode` false-negative risk (silent hang is the bad failure; false-positive just routes to pattern/quiet).
- API shape of `CompletionReason`/`exitCode` additions.

Plan: `desktop-touch-mcp-internal@e92ac3b:docs/issue-386-multiline-echo-completion-sentinel-plan.md` (private; PR #10)


Harusame64:
@codex review

Harusame64:
@codex review

Round 1 fixes applied (commit d4ac6f0b): P1 isUnsafeForExitMode now rejects trailing backtick (PowerShell continuation / bash unterminated command substitution), not just backslash. P2 bash sentinel now emits and requires a trailing pipe terminator after the exit-code field so a multi-digit code cannot match mid-render.

Harusame64:
@codex review

Round 2 fixes applied (commit 16a73dcb): P1 heredoc detection now covers non-letter delimiters (e.g. heredoc with digit/backslash delimiters) and excludes here-strings precisely. P1 added unterminated command-substitution detection (open paren-form) with quote-aware depth tracking. Note: the unsafe-input guard is a best-effort fast-fail; a missed exotic construct degrades to a loud completion.reason=timeout, never a wrong exitCode.

Harusame64:
@codex review

Round 3 fixes applied (commit 83fcca78): P2 replaced the flat command-substitution depth counter with a stack-based scanner so a closing paren inside a double-quoted string no longer wrongly closes an outer substitution (quote/substitution nesting now handled correctly). P2 the PowerShell sentinel regex now accepts a signed exit code (Int32, Windows high-bit status codes); bash stays unsigned.

PR #389, opened 2026-05-22
fix(terminal): #386 P3 — exit-mode send-fidelity hotfix (atomic delivery) + auto warning + real-machine e2e

Harusame64:
Real-machine e2e for `until:{mode:'exit'}` (issue #386) surfaced that the merged P2 was **unusable on conhost** (the most common local terminal). This PR fixes the send-fidelity bug (Q3), the auto-detection honesty gap (Q2), and a P2 `failWith` double-nest, then validates everything with real-hardware e2e (conhost PowerShell + SSH-into-WSL bash + Windows Terminal). **11/11 e2e pass.**

Per the agreed plan, the hotfix is pinned with *unit + real-machine e2e* — so the e2e ships with the hotfix it validates.

## Q3 — atomic delivery (the blocker)
The default send routes conhost through **BG WM_CHAR char-by-char**. A multiline exit command makes conhost execute each embedded newline mid-send; chars posted while it's busy executing saturate/drop (observed: `__DTMCP` → `_DTMCP`), corrupting the byte-exact completion sentinel → the run **timed out**.

Exit mode now delivers the whole command atomically:
- **conhost** (`ConsoleWindowClass`): clipboard + console Paste (`WM_COMMAND 0xFFF1`) via new `pasteIntoConsoleNoFocus()` — **atomic AND no focus steal**. Spike finding: conhost strips lone LF and treats CR as a line break, so the command is CRLF-normalised + a trailing Enter runs the final line.
- **WT / other**: force the foreground clipboard-paste path (atomic; brief focus steal — WT can't take WM_CHAR anyway). Non-exit modes unchanged.

`stripExitArtifacts` rewritten **line-based** (drop the injected prologue/epilogue-echo/sentinel lines by nonce-scoped signature). The previous region-cut assumed the wrong buffer order and ate the real output (which prints *before* the epilogue echoes) — a latent P2 bug the unit fixtures masked. Command echo is kept (same as pattern/quiet).

## Q2 — auto-detection honesty (measured)
A conhost-hosted PowerShell window reports `processName='powershell'` (not `conhost`), so `shell:'auto'` resolves HIGH. Corollary: an SSH-into-WSL bash session under a PowerShell-owned conhost **also** reports `powershell`, so `auto` confidently picks the **wrong (outer) shell** for the remote bash — which degrades to a loud `reason:'timeout'`, not corruption. Exit mode now emits an advisory **warning** on every auto-resolved run (pass `shell` explicitly for nested/remote). `detectShell` TSDoc + plan §12 "conhost→low" claim corrected. No behaviour change.

## Also
P2 `failWith` double-nest fixed (context passed flat, not `{context:{...}}`) so `ExitModeUnsafeInput`'s `context.reason` is exposed (memory `feedback_failwith_third_arg_flat`).

## E2E (real hardware, plan §9 acceptance) — `tests/e2e/terminal-exit-mode.test.ts`
- **conhost PowerShell**: single-line + MULTILINE exit (exitCode 0), native non-zero (`cmd /c exit 7` → 7), cmdlet failure (`$?=False` → 1), cmd reject, unsafe-input reject, auto-resolve + warning.
- **SSH-into-WSL bash** (#383 measurement harness reused): single-line + **MULTILINE #386 core** (byte-exact sentinel, no echo self-match, ~2s elapsed proves it waited through the body), non-zero exit. Output asserted free of `DTMCP` artifacts.
- **Windows Terminal**: `shell:'auto'` → `ExitModeShellAmbiguous` (genuine ambiguous host).
New helper `tests/e2e/helpers/ssh-wsl-launcher.ts`; `powershell-launcher` gains an optional `postScript` hook (reuses its kill/isolation safety to drop into SSH).

## Verification
- `npm run build` clean; full unit suite 234 files / 3644 pass.
- e2e: 11/11 pass on real hardware.
- Guards: failwith-fixtures (regenerated, 175 callsites), stub-catalog, napi-safe, native-types clean.

## Review asks
- **Codex** (API-contract / native-adjacent): the new conhost paste path (clipboard + `WM_COMMAND 0xFFF1`), the exit-mode send routing branch, `stripExitArtifacts` line-based rewrite.
- **Opus**: the delivery-method decision per host (no-steal conhost vs FG-steal WT), the line-based strip correctness (keeps command echo by design), Q2 warning honesty, §3.1 fact sync (detectShell TSDoc + plan), §3.2 (no existing send-path regression for non-exit modes).

Plan: `desktop-touch-mcp-internal@…:docs/issue-386-multiline-echo-completion-sentinel-plan.md` (§9/§12/§13 P3).

### DIFF
--- a/src/engine/bg-input.ts
+++ b/src/engine/bg-input.ts
@@ -360,17 +360,37 @@ export async function pasteIntoConsoleNoFocus(
   return { ok: true };
 }
 
-/** Best-effort restore of a clipboard snapshot captured by getClipboardB64(). */
+/**
+ * Best-effort restore of a clipboard snapshot captured by getClipboardB64().
+ *
+ * Codex #389 P2: when the original clipboard was EMPTY (`""`) or could not be
+ * read (`null`), we must still CLEAR it rather than skip — otherwise the command
+ * we just pasted (possibly sensitive input) lingers in the user's clipboard
+ * after the run. Set-Clipboard rejects an empty value, so clearing uses the
+ * Forms clipboard API.
+ */
 async function restoreClipboard(savedB64: string | null): Promise<void> {
-  if (savedB64 === null || savedB64 === "") return;
   try {
-    const script =
-      `$b=[System.Convert]::FromBase64String('${savedB64}');` +
-      `$t=[System.Text.Encoding]::Unicode.GetString($b);` +
-      `Set-Clipboard -Value $t`;
-    await execFileAsync("powershell.exe", ["-NoProfile", "-NonInteractive", "-Command", script], {
-      timeout: 3000,
-    });
+    if (savedB64 && savedB64.length > 0) {
+      const script =
+        `$b=[System.Convert]::FromBase64String('${savedB64}');` +
+        `$t=[System.Text.Encoding]::Unicode.GetString($b);` +
+        `Set-Clipboard -Value $t`;
+      await execFileAsync("powershell.exe", ["-NoProfile", "-NonInteractive", "-Command", script], {
+        timeout: 3000,
+      });
+    } else {
+      // Originally empty / unreadable — clear so the injected command does not
+      // linger. Clearing an unreadable clipboard is harmless (best-effort).
+      await execFileAsync(
+        "powershell.exe",
+        [
+          "-NoProfile", "-NonInteractive", "-Command",
+          "Add-Type -AssemblyName System.Windows.Forms; [System.Windows.Forms.Clipboard]::Clear()",
+        ],
+        { timeout: 3000 },
+      );
+    }
   } catch {
     /* best-effort */
   }
--- a/src/engine/bg-input.ts
+++ b/src/engine/bg-input.ts
@@ -338,6 +338,10 @@ export async function pasteIntoConsoleNoFocus(
 
   const saved = await getClipboardB64(); // null = read failed; "" = empty clipboard
   if (!(await setClipboardVerified(crlf))) {
+    // Set-Clipboard may have landed before the read-back mismatch (another app
+    // raced, or newline tolerance missed) — restore so we never leave the user's
+    // clipboard clobbered on a failed paste (Opus #389 round 1 P2-1).
+    await restoreClipboard(saved);
     return { ok: false, reason: "clipboard_set_failed" };
   }
 
--- a/src/engine/bg-input.ts
+++ b/src/engine/bg-input.ts
@@ -15,6 +15,8 @@
  * 経由。
  */
 
+import { execFile } from "node:child_process";
+import { promisify } from "node:util";
 import {
   getWindowClassName,
   getWindowProcessId,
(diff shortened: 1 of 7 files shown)


Harusame64:
@codex review

Harusame64:
**Round 1 (Opus review): GO** — fixes applied in b8890851:
- **P2-1**: `pasteIntoConsoleNoFocus` now restores the saved clipboard on the `clipboard_set_failed` path (Set-Clipboard can land before a read-back mismatch, so that path could silently clobber the user clipboard). The post-failure branch already restored.
- **P2-2** (§3.1 doc sync): the plan's normative §9 / §12 confidence-rules / OQ-1 still said "conhost→low loud-fail"; synced to the measured reality (conhost-hosted PS reports `powershell`→high; SSH/WSL nesting is silent-wrong → advisory warning, not ambiguous) — `desktop-touch-mcp-internal@5fe60d4`.
- P3-1 (PS prologue fixed-string strip) accepted as documented (cosmetic, best-effort).

Opus cleared: send-routing (no focus-steal regression — WT already used FG), line-based strip bound (nonce crypto-random), clipboard restore timing, security, §3.2 (pattern/quiet untouched), Lesson 1-4, e2e multi-point assertions.

@codex review

Harusame64:
**Round 2 (Opus re-review): both round-1 fixes verified correct + 1 new §3.1 catch fixed (48c17937).**

Opus confirmed P2-1 (clipboard restore on every early-return path; no-op for null/empty saved) and P2-2 (plan tables synced) are correct/complete. It then caught that the SAME conhost→high correction was still stale in two **live-code** surfaces P3 had missed (§3.1: the LLM-facing schema text is read more than a doc note):
- the `until` exit-variant `shell` schema **description** (LLM-facing) — no longer claims conhost/SSH → ExitModeShellAmbiguous; now: a conhost-hosted shell resolves, SSH/WSL nesting picks the outer shell + warns, only unknown hosts (Windows Terminal) are ambiguous.
- the run-handler shell-resolution comment — same recharacterisation.

This is exactly Opus round 2's GO condition ("after those two edits → GO"). Build clean, exit-mode unit (44) + tool-descriptions pass, stub-catalog + failwith-fixtures guards clean. Awaiting Codex.

Harusame64:
@codex review

Harusame64:
Both Codex P2 findings addressed in d2d8343b:
- **P2-1** (clipboard leak): `restoreClipboard` now actively **clears** the clipboard when the original snapshot was empty (`""`) or unreadable (`null`), instead of skipping — so the pasted command no longer lingers. Non-empty still restores via Set-Clipboard.
- **P2-2** (prologue over-strip): `stripExitArtifacts` now drops only the **first** `$global:LASTEXITCODE = $null` occurrence (the injected prologue echo, which precedes any real output), so a user command that legitimately prints that literal keeps its output. New unit test pins it.

Re-verified: exit-mode unit 45/45, real-machine e2e 11/11.

@codex review

Harusame64:
@codex review

Harusame64:
Codex round-2 findings addressed in a7030ad8 (directions confirmed with Opus):

- **P1** (sendOptions silently ignored on conhost exit): exit mode owns delivery (atomic paste + Enter are required for the sentinel to render), so delivery-shaping options can't be honored. Now **rejected loudly at a single gate before host routing** (uniform across conhost/WT): `method` / `preferClipboard` / `pressEnter` / `chunkSize` / `pasteKey` → `InvalidArgs`. Focus-management options (`focusFirst`/`restoreFocus`/`settleMs`/`forceFocus`/`trackFocus`) stay allowed (no-ops on the no-steal path).
- **P2** (prologue over-strip on tailed output): nonce-tagged the prologue (`… = $null # <nonce>`) and made `stripExitArtifacts` drop lines containing the bare crypto-random nonce — one uniform, nonce-scoped rule (removes the fixed-string match that could delete a real-output line repeating the literal). The `# <nonce>` PowerShell comment runs to EOL and doesn't affect the assignment; echo-immunity preserved.

Unit 51/51 (added a sendOptions-reject matrix + a prologue-literal-kept case), real-machine e2e 11/11.

@codex review

PR #391, opened 2026-05-23
fix(terminal): #384 — opt-in quietMs settle fallback for until:{mode:'pattern'}

Harusame64:
Fixes issue #384: `terminal(action='run', until:{mode:'pattern'})` hung until the hard timeout when a command finished without ever producing the pattern.

## Root cause (research-backed)
A command whose final line has no trailing newline glues the marker to the next shell prompt with no line boundary (`printf X` → `Xuser@host:~$`), so an end-anchored pattern (`X\s*\n` / `X$`) can never bind. Web/deep research (pexpect, expect) confirmed this is the fundamental "no end-marker" problem: `$` premature-matches on streaming output, a literal `\n` requires a newline that may never arrive, and the canonical completion solution is a guaranteed end-marker — which is exactly the already-shipped `until:{mode:'exit'}`.

## Fix — opt-in `quietMs` settle fallback
`until:{mode:'pattern', pattern, quietMs:1000}`: when `quietMs` is set, the run also completes with `completion.reason:'quiet'` (`completion.matchedPattern` absent — check it to tell a match from a settle) once output is stable for that long **without** a match, instead of hanging until `timeoutMs`.
- **Opt-in** — omit `quietMs` to keep the default contract (wait through silent gaps until the pattern appears, issue #196). No regression for long-running commands.
- Match takes priority over settle on the same tick.
- Settle-tracking keys off the raw buffer (like quiet mode), so baseline-lost runs still settle.

## Changes
- schema: pattern variant gains optional `quietMs` (bounds shared with quiet mode).
- handler: pattern branch tracks raw-buffer change + completes on settle when `until.quietMs` set; the immediate post-send check seeds the change-tracking; fallback `quietMs` derived from `until.quietMs` (not the hardcoded 1500) — the three points Opus's design review flagged as load-bearing.
- docs: README (en/ja) completion section + CHANGELOG Fixed entry — anchors can't bind on no-newline output; use exit mode for completion, bare markers for content, or the opt-in settle fallback.

## Verification
- `npm run build` clean; full unit suite green (the 2 transient failures during a parallel e2e run are pre-existing timing flakes — `input-pipeline-dispatch` ADR-019 timeout test + `ui-pattern-store-persistence` debounce — both pass in isolation; neither touches terminal pattern mode).
- e2e: **conhost** never-match pattern → `quiet` (terminal.test.ts); **SSH-into-WSL bash** the actual no-newline repro (`printf X` + `X\s*\n` + quietMs) → `quiet` not timeout (terminal-exit-mode.test.ts). `#383` match-priority e2e unaffected.
- All push-pre guards pass (failwith-fixtures / stub-catalog / napi-safe / native-types).

## Review asks
- **Opus**: the opt-in design (no #196 regression), match-vs-settle priority, raw-buffer change-tracking for baseline-lost, reason:'quiet'+absent matchedPattern honesty, en/ja+CHANGELOG 強制命令 10.
- **Codex** (API-contract): the optional `quietMs` schema field on the pattern discriminated-union variant + the handler completion contract.

Plan: `desktop-touch-mcp-internal@4aa52eb:docs/issue-384-pattern-mode-quiet-settle-fallback-plan.md` (Opus design GO).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,21 @@
   unaffected. This covers single-command idioms (`cmd; echo "DONE"`); multi-line
   commands retain the prior scanning behaviour and are unchanged. (issue #383)
 
+- **`terminal(action='run')` with `until:{mode:'pattern'}` no longer has to hang
+  until the hard timeout when a command finishes without ever printing the
+  pattern.** A command whose final line has no trailing newline glues the marker
+  to the next shell prompt with no line boundary (e.g. `printf X` →
+  `Xuser@host:~$`), so an end-anchored pattern (`X\s*\n` or `X$`) can never match
+  and the run waited the full `timeoutMs`. You can now pass an optional `quietMs`
+  on the pattern spec — `until:{mode:'pattern', pattern, quietMs:1000}` — and the
+  run completes with `completion.reason:'quiet'` (and no
+  `completion.matchedPattern`; check its presence to tell a match from a settle)
+  once output has been stable for that long without a match, instead of hanging.
+  This is opt-in: omit `quietMs` to keep the default behaviour (wait for the
+  pattern until `timeoutMs`, so long-running commands with mid-run silent gaps are
+  unaffected). To detect genuine command completion — and get the exit code —
+  prefer `until:{mode:'exit'}`. (issue #384)
+
 ## [1.7.2] - 2026-05-19 — Emergency-stop now requires a deliberate dwell (no more drive-by failsafe kills)
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


Harusame64:
@codex review

Harusame64:
**Round 1 (Opus review): GO** — no P1/P2. Verified the 3 load-bearing points (change-tracking now live in the pattern branch; immediate-check seeds it; fallback quietMs derived from until.quietMs not the hardcoded 1500), plus match-priority-over-settle, no #196 regression (settle fully gated on quietMs set), reason:'quiet'+absent matchedPattern honesty, baseline-lost raw-buffer tracking, schema (optional quietMs, quiet mode unaffected), 強制命令10 docs en/ja parity, and e2e quality. Implementation matches the approved design exactly. Awaiting Codex + CI; auto-merge on green (auto-mode).

Harusame64:
Codex P2 addressed in the latest commit: the pattern-mode **invalid-regex fallback** now honors the caller's `until.quietMs` (was using the hardcoded 1500 quiet timer). The quiet timer derives from `until.quietMs` in pattern mode — extending the same 'derive from until.quietMs' rule Opus pinned for the valid-regex settle path to the invalid-regex sub-case. Quiet mode + the valid-regex settle path unchanged; build + unit + guards green. @codex review

PR #392, opened 2026-05-23
fix(terminal): #386 — bounded retry for exit-mode conhost clipboard set

Harusame64:
## What & why

Real-machine dogfood (2026-05-23, conhost + Windows Terminal on WSL Ubuntu) of the terminal completion-detection series (#383 / #386 / #384) passed all 12 happy-path runs + 3 documented guardrails. One finding (F1): the **first** exit-mode `terminal(action='run', until:{mode:'exit'})` against a **conhost** window returned `completion.reason:'send_failed'`, empty output, and `warnings:["terminal(action='send') failed: clipboard_set_failed"]`. A manual retry and a back-to-back 5-run probe both succeeded immediately (1/12 across the session) — a textbook transient that coincided with a busy window-launch storm holding the global clipboard.

**Root cause.** Exit-mode delivery on conhost uses `pasteIntoConsoleNoFocus` (clipboard + console Paste). `setClipboardVerified` did a single `Set-Clipboard` + `Get-Clipboard` read-back via a spawned `powershell.exe`; the global clipboard is one OS resource, so a momentary lock by another process makes the set/verify miss — and there was no retry, so a recoverable race became a hard `send_failed`.

## How it's fixed

- `setClipboardWithRetry` wraps the set-and-verify with **3 attempts (1 + 2 retries), 120/240 ms backoff**.
- The clipboard snapshot is captured **once outside the loop**; `restoreClipboard` runs **at most once** (on total failure, or after the paste on success), so a failed set never leaves the user's clipboard clobbered (#389 round-1 P2-1 invariant preserved by construction).
- `setClipboardWithRetry` + `clipboardSetBackoffMs` are **pure / injectable** (`setFn`, `sleepFn`) and unit-tested with no real clipboard or timers.
- `console_paste_post_failed` is intentionally **not** retried (a PostMessage/HWND-liveness failure is not a transient).
- Worst-case added send latency ≈ 1.6 s and only on the failure path — negligible vs exit-mode's seconds-to-minutes completion wait.

Scope is closed by construction: `pasteIntoConsoleNoFocus` has a single caller (exit-mode delivery in `src/tools/terminal.ts`).

## Tests

- `tests/unit/bg-input.test.ts`: backoff growth; first-attempt success (no sleep); fail/fail/succeed → ok at attempt 3 with 120 then 240 ms backoff; all-fail → ok:false with N-1 sleeps; custom `maxAttempts`.
- `npm run test:unit`: 3660 passed (the single failure on a full run was a pre-existing timing-sensitive test in `input-pipeline-dispatch.test.ts` unrelated to this change — passes 70/70 in isolation). `eslint` clean.

## Design / follow-ups

Opus design (option A, bounded retry; B WM_CHAR-fallback and C native-clipboard rejected for this PR). The native-clipboard route (drop the `powershell.exe` spawn — impl already exists in `clipboard_snapshot.rs` but has no standalone napi export) is recorded as future work.

Plan: `desktop-touch-mcp-internal@4b887a8:docs/issue-386-followups.md` (F1)

### DIFF
--- a/src/engine/bg-input.ts
+++ b/src/engine/bg-input.ts
@@ -317,7 +317,10 @@ export const CLIPBOARD_SET_MAX_ATTEMPTS = 3;
 
 /**
  * Backoff (ms) before the retry that FOLLOWS failed attempt `attempt` (0-based):
- * 120 ms after attempt 0, 240 ms after attempt 1, … Pure — unit-testable.
+ * 120 ms after attempt 0, 240 ms after attempt 1, … Pure — unit-testable. Grows
+ * linearly with no cap; with CLIPBOARD_SET_MAX_ATTEMPTS=3 production only ever
+ * uses attempts 0 and 1 (120 / 240 ms) — there is no sleep after the final
+ * attempt — so raise the cap deliberately if you bump the attempt count.
  */
 export function clipboardSetBackoffMs(attempt: number): number {
   return 120 * (attempt + 1);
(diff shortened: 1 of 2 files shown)


Harusame64:
@codex review

Harusame64:
@codex review (round 2 — Opus R1 P2-1/P3-1 applied: prepareClipboardForPaste restore-once invariant + 4 unit tests, backoff comment)

Harusame64:
@codex review (round 3 — test-only: assert sleepFn backoff propagation; production unchanged since round 2)

PR #393, opened 2026-05-23
feat(terminal): #386 — native console-paste for conhost exit-mode (drop powershell clipboard)

Harusame64:
## What & why

Follow-up to the #386 dogfood (PR #392): the conhost exit-mode no-steal console-paste used `powershell.exe` for every clipboard get/set/restore. That spawn was the root of both the one-off `clipboard_set_failed` transient and the failure-path latency blow-up (~6.5 s under load, measured in the post-merge dogfood). This replaces the whole thing with the native Win32 clipboard module that already existed (`clipboard_snapshot.rs`) but had no standalone napi export.

Design (Opus design review + User review applied): `desktop-touch-mcp-internal@571c9d1:docs/issue-386-native-clipboard-plan.md`.

## How

**P1 — Rust**
- `src/win32/console_paste.rs`: `win32_console_paste_no_focus(hwnd, text)` — a **sync** `#[napi]` (wrapped in `napi_safe_call`) that runs the whole save → set → console-Paste (`WM_COMMAND 0xFFF1`) → Enter → restore transaction under one `with_hidden_owner` on the V8 thread, reusing `save_clipboard_supported_formats` / `set_clipboard_unicode_text` / `restore_clipboard_supported_formats` (HGLOBAL save/restore + 3-point sequence race detection + `OpenClipboard` 10×10 ms retry).
  - **sync over AsyncTask**: `win32_foreground_flash_inject` already runs the same clipboard transaction synchronously on the V8 thread and works; a pump-less libuv worker thread is a risky home for a clipboard-owner window. Sync also gets the panic boundary for free via `napi_safe_call`.
  - **restore-once on set failure**: `set_clipboard_unicode_text` empties the clipboard before it can fail at alloc/SetClipboardData, so on that failure the clipboard is empty and MUST be restored. The post-set sequence is captured on **both** success and failure (`get_clipboard_sequence_number`), so the race-aware restore always runs.
  - Reuses `ClipboardError::as_reason` (single source of truth); adds `post_paste_failed`. Enter is **WM_CHAR only** (bit-equal with the old `postEnterToHwnd`).
- `src/win32/input.rs`: factor `pub(crate) fn post_message` shared by the napi export and `console_paste` (Rust never calls the napi wrapper).
- `index.d.ts` / `native-types.ts` / `native-engine.ts`: `NativeConsolePasteResult` + `win32ConsolePasteNoFocus` surface.

**P2 — TS**
- `bg-input.ts`: `pasteIntoConsoleNoFocus` now delegates to the native call; widened `ConsolePasteOutcome.reason`; surfaces `skippedFormats` (non-text formats that could not be preserved — e.g. an image) and `restoreSkippedRace`. Drops `getClipboardB64` / `setClipboardVerified` / `restoreClipboard` / `setClipboardWithRetry` / `prepareClipboardForPaste` / `clipboardSetBackoffMs` and the `execFile`/`promisify` imports. Returns `native_engine_unavailable` (hard-fail) when the addon is missing.
- `terminal.ts`: surface clipboard hints as run `warnings` (skipped non-text formats / restore-skipped-race) on both success and failure paths.

## Tests / guards

- `bg-input.test.ts`: native-mocked outcome + hints mapping (success / reason passthrough / skippedFormats / restoreSkippedRace / `native_engine_unavailable`). The restore-exactly-once invariant moves Rust-side (reason-mapping test in `console_paste.rs`).
- `cargo check` ✓; `check-napi-safe` ✓; `check-native-types` ✓ (68 exports); `eslint` ✓; `tsc` ✓; unit suite ✓ (2 unrelated flaky tests pass in isolation).
- Rust unit tests run on CI/MSVC — the local gnu toolchain lacks `libnode.a` to link the test binary (compile/typecheck passes).
- **The `.node` rebuild + real-machine dogfood is P3** (deferred): the rebuild can't run while an MCP server holds the loaded `.node`. Acceptance is two-tier (transient load = 10/10 hard gate; sustained-hammer = best-effort + typed `clipboard_lock_contention`).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -156,6 +156,19 @@ export interface NativeForegroundFlashResult {
   pasteWarningDetected: boolean
 }
 
+// issue #386 — native no-steal console-paste (conhost exit-mode path).
+export interface NativeConsolePasteSkippedFormat {
+  formatId: number
+  reason: string
+}
+
+export interface NativeConsolePasteResult {
+  ok: boolean
+  reason?: string
+  skippedFormats: Array<NativeConsolePasteSkippedFormat>
+  restoreSkippedRace: boolean
+}
+
 export interface NativeProcessParentEntry {
   pid: number
   parentPid: number
@@ -330,6 +343,7 @@ export declare function win32BuildProcessParentMap(): NativeProcessParentEntry[]
 export declare function win32GetProcessIdentity(pid: number): NativeProcessIdentity
 export declare function win32GetScrollInfo(hwnd: bigint, axis: string): NativeScrollInfo | null
 export declare function win32PostMessage(hwnd: bigint, msg: number, wParam: bigint, lParam: bigint): boolean
+export declare function win32ConsolePasteNoFocus(hwnd: bigint, text: string): NativeConsolePasteResult
 export declare function win32GetFocus(): bigint | null
 export declare function win32VkToScanCode(vk: number): number
 
--- a/src/engine/bg-input.ts
+++ b/src/engine/bg-input.ts
@@ -15,8 +15,6 @@
  * 経由。
  */
 
-import { execFile } from "node:child_process";
-import { promisify } from "node:util";
 import {
   getWindowClassName,
   getWindowProcessId,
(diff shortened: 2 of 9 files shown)


Harusame64:
@codex review

Harusame64:
@codex review (round 2 — R1 findings applied: hints success-only [P2-1], shared hwnd_from_bigint [P3-1], pure map_paste_outcome + branch tests [P3-2], post_paste_failed test [P3-3])

Harusame64:
@codex review (round 3 — extract normalise_crlf so the CRLF test invokes production; no behaviour change)

PR #394, opened 2026-05-23
test(path-class): ADR-021 Phase 4 — DXGI broker SSOT coexistence regression (closes Phase 4)

Harusame64:
## What & why

Closes **ADR-021 Phase 4** (deterministic regression scenarios). Phase 4-1 (the `no-tool-failure-shape-direct-construct` ESLint rule) already landed in #381; this finishes the remaining "pin the 3 past drift cases" part.

**Opus coverage judgment (2026-05-23):** the plan (§3.5.2) prescribed 3 new test files, but that predates the ADR-020 SR contract tests + the Phase 3a adapter test. Two of the three cases are already pinned, so adding new files for them would be redundant (scope violation). Only the DXGI multi-consumer axis is genuinely uncovered:

| past drift case | coverage |
|---|---|
| run_macro F1/F2 | existing `run-macro-stop-on-error-inner-envelope.test.ts` (scenario, hits the `runInnerToolAsResult` adapter) + `run-inner-tool-as-result.test.ts` |
| classifyModal pre/post | existing `d-modal-classifier.test.ts` (property-based, 100 runs) |
| **DXGI broker cacheState** | **NEW** `dxgi-broker-cacheState-coexistence.test.ts` ← this PR |

## The new test

`b-dxgi-cache-state.test.ts` pins the single-consumer 5-value state machine. This pins the **orthogonal** axis the SR-4 `DirtyRectBroker` SSOT exists for (pre-SR-4, vision-gpu / DirtyRectRouter / Stage5 each had their own cache → parallel writers → drift):

1. **SSOT** — N consumers on one `outputIndex` share ONE native subscription (the factory is called **once**, not per consumer). Reverting SR-4 to per-consumer caches makes this fail while `b-dxgi` (single consumer) stays green — the distinct detection power.
2. **per-output isolation** — `outputIndex` entries don't contaminate each other (the `output_index` / secondary-monitor lesson, CLAUDE.md §3.2 / PR #102).
3. **invalidate is output-scoped** — invalidating one output doesn't leak into another's cached subscription.

**Mock note:** the plan named ReplayBackend, but post-SR-4 the broker consumes `factory: (outputIndex) => SubscriptionLike`, whereas ReplayBackend implements `VisualBackend` (recognize/dispose). The correct mock is an instrumented counting `SubscriptionLike` factory (the `b-dxgi` idiom), which also lets us count factory calls per output = the direct SSOT assertion. Time is deterministic via an injected `nowFn` (no real timers).

## Tests

- 3/3 new tests pass; full `tests/unit/path-class-contract/` suite 156/156 green; eslint clean. Test-only change (no production code touched).

Plan: `desktop-touch-mcp-internal@a7df27c:docs/adr-021-result-migration-drift-prevention-plan.md` (§3.5.2 Resolution)

### DIFF
--- a/tests/unit/path-class-contract/dxgi-broker-cacheState-coexistence.test.ts
+++ b/tests/unit/path-class-contract/dxgi-broker-cacheState-coexistence.test.ts
@@ -1,0 +1,145 @@
+/**
+ * tests/unit/path-class-contract/dxgi-broker-cacheState-coexistence.test.ts
+ * ADR-021 Phase 4 PR-P4-3 — deterministic regression: DXGI broker SSOT coexistence.
+ *
+ * `b-dxgi-cache-state.test.ts` already pins the single-consumer, single-output
+ * 5-value state machine. This pins the ORTHOGONAL axis the SR-4 broker exists
+ * for — multi-consumer / multi-output coexistence through ONE SSOT. Pre-SR-4,
+ * vision-gpu / DirtyRectRouter / Stage5 each held their own cache (parallel
+ * writers → cacheState drift, issue #327 item B). SR-4 (PR-SR4-1/2,
+ * `DirtyRectBroker`) made the broker the single owner. This regression pins:
+ *   1. **SSOT** — one native subscription per `outputIndex` regardless of how
+ *      many consumers acquire it (the factory is called once, not once per
+ *      consumer);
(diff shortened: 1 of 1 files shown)


Harusame64:
@codex review

PR #395, opened 2026-05-23
feat(envelope): #352 — success-path advisory hints (keyboard:type → desktop_act)

Harusame64:
## What & why

Implements **issue #352** (ADR-022). Dogfood shows the LLM skips `desktop_discover`, can't hand a lease to `desktop_act`, falls back to `keyboard(type)`, it succeeds, and the LLM then biases toward `keyboard` even where a UIA entity was visible. No tool bug — a **discoverability** problem.

When `keyboard(action='type')` succeeds **but the focused element is a UIA text input** (ValuePattern), attach an additive `advisory` to the response:
```json
"advisory": {
  "preferredPath": "desktop_act",
  "reason": "the focused element is a UIA text input (ValuePattern) — desktop_act runs the lease flow … that keyboard:type does not. keyboard is correct only for UIA-blind targets (PWA / Electron / Canvas).",
  "example": "desktop_discover({target:{windowTitle:'メモ帳'}}) → desktop_act({lease, action:'type', text:'hello'})"
}
```
LLMs follow a concrete corrected example in their own tool history far more strongly than an abstract instruction (compiler "did you mean…?" principle).

## How (probe = zero cost)

The central design decision (Opus design review): the gate reuses the **focused-element snapshot `withPostState` already takes every call** (`post.focusedElement`) — `keyboard(type)` focuses its target, so that *is* the Edit to probe. **No fresh UIA round-trip**, so the non-blocking / latency-budget invariants are structurally true (+0 ms). (The candidate cache is cold on discover-skip; the reactive view is empty for non-critical lenses; verify-readback is text-only — all unworkable, per Opus R1.)

- `src/tools/_advisory.ts` (new): `AdvisoryHint` + `maybeAdvisory(toolName, args, focusedElement)` pure builder + emit counter. Gate = `type ∈ {Edit, Document}` + `value !== undefined` (ValuePattern exposed). Per-tool logic lives here; no `keyboard → desktop_discover` dep (pure over `PostElementInfo`, calls no UIA).
- `_post.ts` `withPostState`: success branch calls `maybeAdvisory(...)` and sets root `obj.advisory`. Writer-ownership doc extended (ADR-021 PR-P2-1 pattern). **No marker / no ROOT_HOISTED_KEYS change** — `withPostState` already holds both `args` and `post.focusedElement`.
- `server-status.ts`: `counters.advisoryEmitted` (OQ-5 — the only objective fire-rate / under-fire signal; the LLM-driven E2E harness was scrapped).

**Suppression (false positives):** UIA-blind targets (PWA / Electron / Canvas) expose no text-input `value`, so the gate fails → no advisory. Definitional, no app-class probe.

## Scope

Round 1 = the keyboard pair only. Other pairs (`click_element` / `mouse_click` / `screenshot`→`terminal`) and a bounded fresh probe (if dogfood shows under-fire) are future rounds. The deep fix (auto-lease on `desktop_act`) is a separate future ADR. **Measurement is dogfood-only** + the emit counter (harness scrapped — acknowledged limitation).

## Tests

- `tests/unit/advisory-hints.test.ts` (10): builder hits / suppression (non-text / no-value / null / non-keyboard / non-type) / example filling + truncation/sanitize / emit counter.
- `post-writer-ownership.test.ts` (+3): `obj.advisory` is a withPostState success-only writer; never on failure; ROOT sibling of `post`.
- Full unit suite 3681 passed; eslint + tsc clean. Additive — advisory-off path byte-equal to today.

Plan: `desktop-touch-mcp-internal@1626d6a:docs/adr-022-success-path-advisory-hints.md`

### DIFF
--- a/src/tools/_advisory.ts
+++ b/src/tools/_advisory.ts
@@ -102,7 +102,9 @@ function truncate(text: string): string {
   return text.length > EXAMPLE_TEXT_MAX ? `${text.slice(0, EXAMPLE_TEXT_MAX)}…` : text;
 }
 
-/** Make a string safe to embed inside the single-quoted example literal. */
+/** Make a string safe to embed inside the single-quoted example literal.
+ *  Escape backslash FIRST (else a later `'`→`\'` would be mis-paired), then the
+ *  quote, then collapse newlines (CodeQL: incomplete string escaping). */
 function sanitize(s: string): string {
-  return s.replace(/[\r\n]+/g, " ").replace(/'/g, "\\'");
+  return s.replace(/\\/g, "\\\\").replace(/'/g, "\\'").replace(/[\r\n]+/g, " ");
 }
--- a/src/tools/_advisory.ts
+++ b/src/tools/_advisory.ts
@@ -28,8 +28,30 @@ export interface AdvisoryHint {
   example: string;
 }
 
-/** UIA control types that represent an editable text input. */
+/** UIA control types that represent an editable text input. NOTE: ComboBox is
+ *  deliberately EXCLUDED — dogfood (ADR-022) showed Chromium exposes web text
+ *  inputs (e.g. Google search) as ComboBox, so admitting it would mis-fire on
+ *  browser content where browser_* / keyboard is the right path, not desktop_act. */
 const TEXT_INPUT_CONTROL_TYPES = new Set(["Edit", "Document"]);
+
+/** UIA automationId of a Chromium web-area root. dogfood (ADR-022) showed a
+ *  browser's focused element resolves to this Document with value=URL — a wrong
+ *  desktop_act nudge (web content uses browser_* / keyboard, not desktop_act). */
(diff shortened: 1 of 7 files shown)


Harusame64:
@codex review

Harusame64:
@codex review (round 2 — applied: P2 unnamed-text-input fix [snapshotFocusedElement gates on controlType not name] + CodeQL backslash escaping in sanitize)

Harusame64:
@codex review (round 3 — R2 P2 resolved: reverted the dead snapshotFocusedElement change [upstream toInfo already nulls name-empty]; documented the named-only Round-1 limitation + backlogged the real upstream fix. CodeQL escaping fix retained.)

Harusame64:
@codex review (round: dogfood-driven gate refinement — browser processName suppression + automationId!=='RootWebArea' added; ComboBox stays excluded. Notepad still fires.)

PR #396, opened 2026-05-23
feat(envelope): #352 — fire success-path advisory for unnamed text inputs

Harusame64:
## What changed
The `keyboard:type → desktop_act` success-path advisory (issue #352, PR #395) only fired for **named** UIA text inputs. The focused-element bridge dropped name-empty elements before they reached the advisory gate — yet many real `Edit`/`Document` fields expose `ValuePattern` with an empty `Name`. Typing into such an unlabeled input now also surfaces the `desktop_act` nudge.

## Why it matters
Live dogfood (2026-05-23) measured the emit rate: 1 fire for the named Notepad case, **0 for any unnamed field**. The headline case worked, but unlabeled inputs systematically under-fired.

## How it's scoped (no behavior change for other consumers)
- `getFocusedAndPointInfo` gains an `includeUnnamed` param (default `false`) via a shared `dropFocusRow` predicate so the native + PowerShell guards cannot drift.
- **Only** `_post`'s `snapshotFocusedElement` opts in (`includeUnnamed:true`) and relaxes its own name-empty belt, so the unnamed editable element reaches `post.focusedElement` and the (already name-agnostic) advisory gate fires. A degenerate row (no name **and** no controlType) is still dropped.
- `desktop_state` / `mouse_click` verification / perception pass **no flag → byte-equal** (they keep dropping name-empty rows exactly as before). `desktop_state` is additionally self-protected by its own belt guards.

## Tests
- `advisory-hints`: unnamed Edit with `value:""` (ValuePattern present) fires; `value` absent (no ValuePattern) drops — pins the `value:"" ≠ undefined` distinction.
- `uia-focused-unnamed` (new): `includeUnnamed` off → name-empty dropped (byte-equal regression pin); on → kept; degenerate row still dropped; named rows unaffected.
- `post-writer-ownership` G4-survival pin: an unnamed Edit reaches `post.focusedElement` (name:"") and fires the advisory.

Full unit suite green; the only full-suite failures were pre-existing flaky/perf/e2e (confirmed: identical failures on `main`, and the unit ones pass in isolation).

## Plan / design
Opus-reviewed 2 rounds (GO). Plan + blast-radius analysis (corrects ADR-022 §5.5): [`adr-022-unnamed-input-followup-plan.md`](https://github.com/Harusame64/desktop-touch-mcp-internal/blob/6553afbf9e7f4533323a6feca1e8ed1fec4c76e4/docs/adr-022-unnamed-input-followup-plan.md) (private repo).

### DIFF
--- a/src/engine/uia-bridge.ts
+++ b/src/engine/uia-bridge.ts
@@ -410,6 +410,25 @@ export interface UiaFocusInfo {
 }
 
 /**
+ * Should a focused / at-point row be dropped? (#352 follow-up, ADR-022 §5.5)
+ *
+ * A `name`-empty row is dropped by DEFAULT — this is the historical behavior and
+ * keeps `_mouse-verify` / `desktop_state` / perception byte-equal (none of them
+ * pass `includeUnnamed`). The advisory/post path (`_post.ts snapshotFocusedElement`)
+ * opts in with `includeUnnamed:true` so an UNNAMED text input (many real
+ * Edit/Document fields expose ValuePattern with an empty Name) can reach the #352
+ * advisory gate. Even when opted in, a degenerate row (no name AND no controlType)
+ * is still dropped. Single predicate so the native + PS guards cannot drift.
+ */
+function dropFocusRow(
+  name: string | undefined,
+  controlType: string | undefined,
+  includeUnnamed: boolean
+): boolean {
+  return includeUnnamed ? (!name && !controlType) : !name;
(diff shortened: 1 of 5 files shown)


Harusame64:
@codex review

Harusame64:
Force-pushed: amended commit message + docs SSOT sync (ADR §7 maybeAdvisory 4-arg). @codex review

PR #399, opened 2026-05-23
docs(release): note npx smoke test must run outside project dir

Harusame64:
## What
Adds a note to the Smoke Test section of `docs/release-process.md`: the post-publish `npx` smoke test must be run from a cwd **outside** the project source dir (e.g. `cd \$env:TEMP` first).

## Why
During the v1.8.0 release, the smoke test failed with `"'desktop-touch-mcp' is not recognized as an internal or external command"`. Root cause: once the project's own `package.json` version equals the just-published `X.Y.Z`, running `npx` from the project root resolves the **local (uninstalled)** package instead of the registry one. This is a false negative caused by cwd, not a release bug — re-running from `\$env:TEMP` succeeds.

This was the leftover learning from the v1.8.0 release that hadn't been committed yet.

## Scope
Docs-only, single self-contained note. No code changes.

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -455,6 +455,14 @@ rm -rf "$USERPROFILE/.desktop-touch-mcp/releases/vX.Y.Z"
 npm cache clean --force
 
 # 2. Download via npx and verify --help
+#    ⚠️ Run this from a cwd OUTSIDE the project source dir (e.g. `cd $env:TEMP` first).
+#    The project's own package.json has name=@harusame64/desktop-touch-mcp; once its
+#    version equals the just-published X.Y.Z, running npx from the project root makes
+#    npx resolve the LOCAL (uninstalled) package instead of the registry one and fail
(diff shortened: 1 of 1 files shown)


Harusame64:
P2 review finding (fixed in aa22e0c):

**Shell mismatch** — the note lives inside a ```bash code fence (the block uses `rm -rf`, `$USERPROFILE`, `for…do`, `test -d`), but it referenced PowerShell-only `$env:TEMP` / `cd $env:TEMP`. In bash, `$env:TEMP` expands to empty (`cd :TEMP` fails), so a reader running the smoke test in Git Bash would hit a broken instruction.

Fix: made the cwd guidance shell-neutral — `PowerShell: cd $env:TEMP | bash: cd /tmp` — and dropped the bare PowerShell var from the prose ("re-run from a temp dir").

Rest of the note is technically accurate (local-package resolution behavior + the older-version cross-check nuance both check out). No P1.

PR #400, opened 2026-05-23
feat(browser): Phase 0 reachability — browser_click scrollIntoView + browser_eval timeout hint

Harusame64:
## What (ADR-023 Phase 0 — reachability improvements, resolver-independent)

Two small, opt-in reachability fixes that resolve the two least-coupled GSC dogfood frictions ahead of the larger semantic-targeting work.

### 1. `browser_click({scrollIntoView:true})` — FR-5
When the target is outside the viewport, scroll it to centre and click instead of hard-failing with `ElementNotInViewport`. Default `false` preserves the explicit-scroll-then-retry workflow (NFR-1 backward compat). Implemented as a single best-effort eval before the viewport gates, so both the auto-guard pre-check and the main coord fetch see the scrolled element.

### 2. `browser_eval` timeout → `wait_until` typed hint — FR-8
A single `browser_eval` runs under the CDP per-command timeout (~15s), so in-page polling loops always exhaust it (dogfood: "60s polling inside eval died at 15s"). Now:
- A timed-out eval returns a typed **`BrowserEvalTimeout`** (was the misleading generic `UiaTimeout`) with fixed-string hints pointing at `wait_until` (`element_matches` / `url_matches` / `ready_state`).
- The `browser_eval` description warns against in-page polling and points to `wait_until`.

`BrowserEvalTimeout` is surfaced via `failCode` with inline suggest (same pattern as the existing `ElementNotInViewport` — no `SUGGESTS`/`classify` change, no catalog drift).

## Verification
- `npm run build` — clean
- `npm run lint` — 0 errors
- `npx vitest run --project=unit` — **3691 passed | 2 skipped**
- Regenerated `src/stub-tool-catalog.ts` + `tests/fixtures/failwith-callsite-shapes.json` (committed)

## Plan
`desktop-touch-mcp-internal@8ced979:docs/adr-023-browser-semantic-targeting.md` §6 Phase 0 (Decision finalized + Round 1 Opus review applied). Phase 0 is independent of the resolver work (Phase 1) by design.

## Scope
browser.ts only (+ regenerated artifacts + CHANGELOG). No Rust, no native binding. No change to the existing CSS-selector click/fill paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,28 @@
 
 ## [Unreleased]
 
+### Added
+
+- **`browser_click` can scroll an off-screen target into view for you.** Pass
+  `scrollIntoView:true` and, when the element is outside the visible viewport,
+  the click first scrolls it to the centre of the view and then clicks — instead
+  of failing with `ElementNotInViewport` and asking you to scroll it in yourself
+  with a separate `browser_eval`. The default stays `false`, so existing calls
+  behave exactly as before.
+  - `browser_click({selector:'#row-42 .open', scrollIntoView:true})`
+
+### Changed
+
+- **`browser_eval` now points you to `wait_until` when an eval times out.** A
+  single `browser_eval` runs under the browser's per-command timeout (~15s), so a
+  polling loop written *inside* an eval will always exhaust it. Two changes make
+  that failure recoverable instead of confusing: the tool description now warns
+  against in-page polling, and a timed-out eval returns a typed
+  `BrowserEvalTimeout` whose hints point at the right waiting tool —
+  `wait_until({condition:'element_matches'})` for a DOM element/text,
+  `'url_matches'` for an SPA route change, or `'ready_state'` for page load —
+  rather than a generic timeout error.
+
 ## [1.8.0] - 2026-05-23 — Terminal `run` can wait for a command to finish (exit code) + a `desktop_act` hint after typing into native fields
 
 ### Added
--- a/src/engine/cdp-bridge.ts
+++ b/src/engine/cdp-bridge.ts
@@ -12,7 +12,10 @@
 import WebSocket from "ws";
 
 export const DEFAULT_CDP_PORT = 9222;
-const CMD_TIMEOUT_MS = 15_000;
+// Exported so callers that surface a CDP-timeout-derived hint (e.g.
+// browser_eval's BrowserEvalTimeout) can build the user-facing seconds value
+// from this single source instead of hardcoding "15s" (ADR-023 Phase 0 review P2-2).
+export const CMD_TIMEOUT_MS = 15_000;
 const CONNECT_TIMEOUT_MS = 5_000;
 
 // ─── Types ────────────────────────────────────────────────────────────────────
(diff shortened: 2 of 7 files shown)


Harusame64:
@codex review

Production code PR (ADR-023 Phase 0). Please focus on:
- `scrollSelectorIntoViewIfNeeded` best-effort error swallowing + interaction with the two `ElementNotInViewport` viewport gates in `browserClickElementHandler` (auto-guard pre-check vs main path).
- The `browser_eval` catch: `msg.includes("CDP timeout")` string-match robustness against `cdp-bridge.ts` (`CDP timeout: ${method} did not respond within ${CMD_TIMEOUT_MS}ms`), and the typed `BrowserEvalTimeout` (failCode + inline suggest, no SUGGESTS/classify change — same pattern as `ElementNotInViewport`).
- NFR-1 backward compat: `scrollIntoView` default false must keep existing `browser_click(selector)` bit-equal.
- API contract / schema regen (stub-tool-catalog + failwith-callsite-shapes fixtures committed).

Harusame64:
Round 1 findings applied in ec47b04:

- **Codex P1 (guard ordering)**: the opt-in scroll now runs AFTER every guard block decision (lens `evaluatePreToolGuards` + auto-guard `runActionGuard`). A denied action no longer scrolls. The auto-guard viewport pre-check skips its hard-fail only when `scrollIntoView` is set (post-guard scroll + main viewport check cover it; `readyState!=complete` + off-viewport still blocks via the guard's `browserSelectorInViewport` path).
- **Codex P2 (timeout scope)**: the `BrowserEvalTimeout` remap is now scoped to the `evaluateInTab(preparedExpression)` call only — a post-eval `getCachedTabContext` timeout keeps its normal classification.
- **Opus P2-1**: extracted `maybeBrowserEvalTimeoutFailure` + pinned by `tests/unit/browser-eval-timeout-typed-code.test.ts` (the cross-file 'CDP timeout' string contract can't silently degrade to UiaTimeout).
- **Opus P2-2**: `CMD_TIMEOUT_MS` exported; the failCode seconds value derives from it (schema description stays literal to avoid breaking cdp-bridge mocks at module load).
- **Opus P2-3**: `scrollIntoView` uses `behavior:'instant'`.

Verification: build clean, lint 0 errors, unit `browser-*` / `failcode-sweep` / `browser-launch-killexisting` / `envelope` all green.

@codex review

Harusame64:
Round 2 reviews resolved (HEAD 55ee6f3):

- **Opus Round 2**: P1 zero, verified all Round 1 fixes (Codex P1/P2 + Opus P2×3) correct, no new P1/P2. Merge-ready.
- **Codex Round 2**: Round 1 P1/P2 inline comments went outdated (resolved). One new **P3** (substring `includes("CDP timeout")` could match a page-script error mentioning the phrase) → **fixed in 55ee6f3**: now matches the bridge's exact `startsWith("CDP timeout:")` prefix (a page-script error is wrapped as `JS exception in tab: ...`, so it can't be mislabeled). Added false-positive coverage to the pin test.

All P1/P2/P3 from both reviewers addressed. Final pass:

@codex review

PR #402, opened 2026-05-24
feat(browser): ADR-023 Phase 1 PR2 — resolver action-target core (gather/decide split)

Harusame64:
## What

ADR-023 Phase 1 **resolver action-target core** — the internal engine that lets `browser_click` / `browser_fill` target elements by meaning (text / role / ariaLabel) instead of CSS selectors. This PR is the resolver only; the public `browser_click({by})` / `browser_fill({by})` surfaces land in the follow-up PRs.

Built as a **gather / decide split** (plan §2.bis, user-confirmed):

- **Gather (injected JS, the only DOM access)** — `buildActionCandidateFactsJs` reuses the exact `browser_search` candidate matching/scoring via a shared `candidateMatchingBodyJs` fragment, takes the top-N=8 by score, and gathers per-candidate raw facts: `chain[self..D=3]` clickable signals, `nearestLabels`/`containerHint`, `receivesEvents` via `elementFromPoint` hit-test, and window-level viewport metrics (so the click rect→screen-px conversion needs no second `querySelector`).
- **Decide (pure TS, no DOM)** — `decideActionTarget` applies actionability + ancestor climb + the uniqueness contract: **auto-act only when exactly one actionable candidate resolves**, else `ambiguous` / `noActionable` (no score-margin guessing). Unit-tested in node.
- **Wrapper** — `resolveActionTarget` runs one gather eval then the pure decide, and on `resolved` converts the viewport rect center to a physical screen point.

## Why byte-equal is safe

The `browser_search` injected IIFE is unchanged — the shared fragment composes into the identical string (the existing candidate-collection snapshot still passes). The matching/scoring contract now has a single source instead of a second copy of the per-axis loops.

## Tests

- **node unit** — pure decision (`clickableStrength` / `isActionable` / `climbToClickable` / `decideActionTarget` uniqueness + dedup + ambiguity) and string snapshots of both injected builders (incl. role filter); `new Function()` syntax-checks all 5 axes.
- **real headless Chrome e2e** (`tests/e2e/browser-resolver.test.ts` + GSC-like fixture) — unique resolution with an off-viewport duplicate (AC-1/2), label→button climb depth 1 (AC-3), two distinct actionable buttons → ambiguous with the non-actionable duplicate still listed (AC-3), no-strong-clickable → noActionable, overlay-occluded button → noActionable, role filter, fill-path resolve without the receivesEvents gate (AC-4), and the ScopeNotFound gather error.

Plan: `desktop-touch-mcp-internal@a380117:docs/adr-023-phase-1-resolver-plan.md` (S1/S2/S3)


Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
@codex review

PR #407, opened 2026-05-24
fix(browser): guard browser_click against a minimized/off-screen window (+ browser-cdp e2e readiness)

Harusame64:
## What & why

Two browser-robustness fixes (separate commits), the second discovered while verifying the first.

### 1. `browser_click` minimized-window guard (production fix)
A minimized Chrome/Edge window reports its origin at the Windows `-32000` parking marker (`window.screenX/screenY`) while the page layout stays valid. `browser_click`'s viewport→screen conversion then yields an off-screen-negative point the OS clamps to **(0,0)** — the cursor lands in the top-left failsafe zone and the dwell watchdog kills the server (`Connection closed`). Found during real GSC dogfooding.

Now both click paths (`selector` and `by`+`pattern`) detect a minimized/off-screen window from the window origin **before** the OS click and stop with a typed **`BrowserTargetMinimized`** error (restore-and-retry hint) instead of mis-firing. `browser_fill` is **exempt** — it writes via a CDP eval, not an OS click, so it still works on a minimized window.

- engine: `MINIMIZED_WINDOW_SCREEN_COORD` + `isOffscreenMinimized()` (`cdp-bridge.ts`); `getElementScreenCoords` now returns `screenX/screenY`.
- tools: `browserMinimizedFailure()` shared by both click paths (inline `failCode`, no SUGGESTS/catalog drift — `BrowserAmbiguousTarget` precedent).
- tests: unit predicate boundaries (incl. legitimate negative-origin secondary monitor is **not** flagged) + headless e2e that overrides `window.screenX=-32000`, asserts both paths stop **without clicking** (the `onclick` side effect stays empty) and that `fill` still succeeds.

### 2. `browser-cdp.test.ts` flaky readiness (test-only fix)
Verifying #1 surfaced a pre-existing flake: under full-suite load the CDP test file failed ~1-in-4 with `#btn-submit not found` / `document.body is null`. Root cause = a **startup timing race**: `waitForLoad` only checked `readyState`+`title`, which passes the instant `<head>` commits — before `<body>` is parsed. `resolveTab(null)` was resolving the fixture correctly on every probe, so **no production change** is needed. Now it polls for the fixture element to exist **and** be laid out (`getBoundingClientRect().width > 0`), the same element-level readiness the other (non-flaky) browser e2e suites use. **Verified: 12-file browser e2e set 6/6 green.**

## Test
- `tests/unit/browser-minimized-guard.test.ts` — 6/6
- `tests/e2e/browser-click-minimized.test.ts` — 4/4 (headless)
- existing browser suites (cdp / click-by-axis / fill-by-axis / cdp-verification) green; `check:failwith-fixtures` regenerated (line-number only).

## Plan / design (private repo)
`desktop-touch-mcp-internal@ac8751b:docs/browser-click-minimized-window-followup.md` — option B (typed-error guard) locked, with carry-over (A) restore-first as a future opt-in.

## Follow-up (separate PR)
opt-in gate for real-OS-input e2e (`mouse-*` / `force-focus` / `tool-chain` clicks) so a normal `test:capture` never moves the physical cursor on a live desktop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,19 @@
   `'url_matches'` for an SPA route change, or `'ready_state'` for page load —
   rather than a generic timeout error.
 
+### Fixed
+
+- **`browser_click` no longer mis-fires at the screen corner when the browser
+  window is minimized.** A minimized Chrome/Edge window reports an off-screen
+  position for its content, so `browser_click` could compute a click point the OS
+  clamps to the top-left corner (0,0) — moving the cursor there and tripping the
+  built-in safety stop, which closed the connection. `browser_click` now detects a
+  minimized/off-screen window up front and stops with a typed
+  `BrowserTargetMinimized` error (with a hint to restore the window and retry)
+  instead of clicking. This applies to both the `selector` and the `by`+`pattern`
+  paths. `browser_fill` is unaffected — it types into the page directly and works
+  even when the window is minimized.
+
 ## [1.8.0] - 2026-05-23 — Terminal `run` can wait for a command to finish (exit code) + a `desktop_act` hint after typing into native fields
 
 ### Added
--- a/src/engine/cdp-bridge.ts
+++ b/src/engine/cdp-bridge.ts
@@ -18,6 +18,29 @@ export const DEFAULT_CDP_PORT = 9222;
 export const CMD_TIMEOUT_MS = 15_000;
 const CONNECT_TIMEOUT_MS = 5_000;
 
+/**
+ * Windows parks a minimized (or pathologically off-screen) top-level window's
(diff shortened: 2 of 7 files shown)


Harusame64:
@codex review

Production change (browser_click minimized-window guard) + a test-only flaky fix. Please focus on:
- the `-32000` threshold / `isOffscreenMinimized` (no false-positive on legitimate negative-origin multi-monitor layouts);
- guard placement on BOTH click paths (selector + by-axis) reached before the OS click in all auto-guard on/off combinations;
- `browser_fill` correctly left unguarded (CDP-eval write, no OS click);
- `ElementCoords` screenX/screenY addition not breaking existing consumers;
- the typed `BrowserTargetMinimized` (inline failCode, no SUGGESTS) API-contract shape.

PR #408, opened 2026-05-24
test(e2e): emergency stop for the live-desktop e2e suite (npm run e2e:stop)

Harusame64:
## What & why

The e2e suite drives **real OS input** (mouse/keyboard) on the live desktop. If a run misbehaves (cursor hijacked, foreground churn) you currently can only Ctrl+C the runner. This adds a clean way to halt a run **without racing the test for the cursor**.

`npm run e2e:stop` (from any terminal) drops a `.e2e-stop` sentinel file; a global `beforeEach` then **skips every remaining test at the next boundary** via `ctx.skip()`. Because it's `ctx.skip()` (not `process.exit`/throw), each file's `afterEach`/`afterAll` still run, so Chrome instances and spawned windows are **torn down cleanly** (no orphans), and the run ends with exit 0.

### Why a terminal command, not a STOP-button window
A clickable window is **unclickable while a test is driving the cursor via SendInput** (it fights you for the pointer), and a persistent top-most window would show up in screenshot / window-enumeration / focus tests and make *their* results waver. The filesystem sentinel sidesteps both — it never touches the desktop. (I prototyped an opt-in STOP window but it did not render reliably when spawned from vitest's globalSetup context, and added no value over the terminal command, so it was dropped.)

## Pieces
- `tests/e2e/helpers/stop-sentinel.ts` — sentinel path + `is/request/clear` helpers (path arg defaults to the real sentinel; overridable for the unit test).
- `tests/e2e/abort-check.ts` — `setupFiles` `beforeEach` that `ctx.skip()`s when the sentinel exists.
- `tests/e2e/global-setup.ts` — `globalSetup` clears a **stale** sentinel at start + on teardown (so a leftover from a crashed run never bricks a fresh one).
- `scripts/e2e-stop.mjs` + `package.json` `"e2e:stop"`.
- `vitest.config.ts` e2e project: `globalSetup` + `setupFiles`.
- `tests/unit/e2e-stop-sentinel.test.ts` — helper round-trip against a temp path (never touches the real sentinel).
- `.gitignore` — `.e2e-stop`.

## Test / verification
- unit 3/3.
- **mid-run stop**: started a 3-file browser e2e run, ran `npm run e2e:stop` ~6s in → 27 ran, **39 skipped**, exit 0, `afterAll` cleanup ran, sentinel auto-cleared by teardown.
- **normal run unaffected**: `browser-cdp.test.ts` → 27 passed / 2 skipped (headed-gated baseline), no spurious skips.
- No production / user-facing change (dev test-infra only) → no CHANGELOG entry.

## Context
Follow-up to #407 (browser_click minimized-window guard). The user is fine running real-input e2e on the live desktop; this just gives a reliable way to abort a run when needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -95,3 +95,6 @@ tests/repros/
 # Per-ADR PowerShell investigation spikes (Windows Terminal, RDP, etc.).
 # Findings live in the private companion repo's ADR docs.
 scripts/spikes/
+
+# e2e emergency-stop sentinel (runtime, never commit)
+.e2e-stop
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "test:capture": "node scripts/test-capture.mjs",
     "test:unit": "vitest run --project=unit",
     "test:e2e": "vitest run --project=e2e",
+    "e2e:stop": "node scripts/e2e-stop.mjs",
     "test:headed": "HEADED=1 vitest run --project=e2e",
     "test:e2e:browser": "vitest run --project=e2e \"tests/e2e/browser-*.test.ts\"",
     "test:e2e:window": "vitest run --project=e2e tests/e2e/dock-window.test.ts tests/e2e/dock-auto.test.ts tests/e2e/focus-integrity.test.ts tests/e2e/force-focus.test.ts tests/e2e/screenshot-electron.test.ts tests/e2e/ui-elements-cache.test.ts",
--- a/scripts/e2e-stop.mjs
+++ b/scripts/e2e-stop.mjs
@@ -1,0 +1,22 @@
+#!/usr/bin/env node
+/**
+ * e2e-stop.mjs — request an emergency stop of a running e2e suite.
+ *
+ * Drops the `.e2e-stop` sentinel that tests/e2e/abort-check.ts watches; the run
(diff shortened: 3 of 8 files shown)


Harusame64:
@codex review

Test-infra only (no production change): a filesystem-sentinel emergency stop for the live-desktop e2e suite. Please focus on:
- the sentinel path is resolved identically in `tests/e2e/helpers/stop-sentinel.ts` and `scripts/e2e-stop.mjs` (both → repo-root `.e2e-stop`);
- `ctx.skip()` in the global `beforeEach` (abort-check.ts) halting remaining tests while still running afterEach/afterAll cleanup;
- globalSetup's stale-clear not racing a just-requested stop;
- normal runs (no sentinel) being unaffected.

Harusame64:
@codex review

Round 1 fix pushed (e25c05c) for your P1: globalSetup no longer clears `.e2e-stop` unconditionally — it now uses `clearStaleStop(launchedAtMs)` which clears the sentinel ONLY if its mtime predates this process's launch (a prior-run leftover); a stop issued during the boot window (mtime ≥ launch) is preserved so the run still halts. New unit tests (utimesSync) pin stale-clears-vs-fresh-preserves. Please confirm.

PR #129, opened 2026-05-02
feat(expansion): mouse_drag commit wrapper (L5 envelope)

Harusame64:
## Summary

`mouse_drag` を `makeCommitWrapper` で wrap (lease-less commit variant)。PR #121 mouse_click 同型 pattern (sub-plan §3 30 分タイムアタック template、raw shape 3a family)。

## scope

- `src/tools/mouse.ts`: module-scope `mouseDragRegistrationHandler` + `mouseDragRegistrationSchema = withEnvelopeIncludeSchema(mouseDragSchema)` export、`server.tool` の 3rd arg を `mouseDragRegistrationSchema` に切替、handler を `mouseDragRegistrationHandler` に切替。`withRichNarration` 内側 (windowTitleKey: \"windowTitle\"、本 PR 前から既に wrap 済) → `makeCommitWrapper` 外側で envelope wrap を追加。
- `src/tools/macro.ts`: `TOOL_REGISTRY.mouse_drag.schema` を `z.object(mouseDragRegistrationSchema)` に切替 + handler を shared instance に切替 (PR #112 shared registration handler pattern、strip risk 防止)。
- `tests/unit/expansion-mouse-drag-wrapper.test.ts`: 4 contract tests (E1: L1 ToolCallStarted/Completed events、E2: include 未指定時 raw shape return、E3: include=causal で `your_last_action = \"mouse_drag(...)\"`、E4: trunk completion contract mechanical copy) — PR #121 mouse_click wrapper test pattern mechanical copy。
- `src/stub-tool-catalog.ts`: `npm run check:stub-catalog` で再生成された差分 (`include` field 追加分)。

## trunk contract conformance check

- engine-perception layer 改変ゼロ (\`npm run check:expansion-disjoint\` OK + expansion-pr-guard.yml)
- run_macro 経路同 instance 共有 (PR #112 pattern)
- handler internal logic + Zod schema + 戻り値 shape 不変 (ADR-010 §1.5)

## Test plan

- [x] \`npm run build\` (tsc) green
- [x] \`npm run check:stub-catalog\` (再生成 diff を本 PR に commit)
- [x] \`npm run check:expansion-disjoint\` OK (trunk lock paths 4 untouched)
- [x] \`npx vitest run tests/unit/expansion-mouse-drag-wrapper.test.ts\` 4/4 pass
- [ ] CI Linux で trunk lock layer 改変ゼロ確認

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #412, opened 2026-05-24
fix(e2e): focusWindow → restoreAndFocusWindow (undefined import, CodeQL #129)

Harusame64:
## What
Fixes CodeQL **`js/call-to-non-callable`** alert **#129** (`tests/e2e/tool-chain.test.ts:39`, "Callee is not a function: type undefined") — and the same latent bug in 3 sibling e2e files.

## Root cause
Four e2e tests imported `focusWindow` from `src/engine/win32.ts`, which **exports no such symbol** (`restoreAndFocusWindow` / `forceSetForegroundWindow` exist instead). The import resolved to `undefined`; `focusWindow(np.hwnd)` threw *"focusWindow is not a function"* at runtime and was **swallowed by the surrounding `try/catch`** — so these tests silently never focused their window in `beforeAll`. `tsc` didn't catch it (the build tsconfig doesn't type-check e2e tests; vitest uses esbuild = no type-check). CodeQL did.

## Fix
Replace with the repo-wide window-focus helper `restoreAndFocusWindow(hwnd)` (SW_RESTORE + SetForegroundWindow — the same call the keyboard / mouse / dock / scroll tools use). `np.hwnd` is a `bigint`, which it accepts.

Files: `tests/e2e/{tool-chain,focus-integrity,rich-narration-edge,ui-elements-cache}.test.ts`.

## Notes
- Pre-existing on `main`; **unrelated to the ADR-023 modal work**.
- Test-only change (no production code). Not running these e2e here — they launch Notepad + focus real windows (live-desktop). The fix is a mechanical correction to a helper used throughout the codebase; `tsc --noEmit` clean.
- Alert #129 auto-closes when this merges (real fix, not a suppression).

### DIFF
--- a/tests/e2e/focus-integrity.test.ts
+++ b/tests/e2e/focus-integrity.test.ts
@@ -27,13 +27,13 @@ import { terminalSendHandler } from "../../src/tools/terminal.js";
 import { launchNotepad, type NpInstance } from "./helpers/notepad-launcher.js";
 import { launchPowerShell, type PsInstance } from "./helpers/powershell-launcher.js";
 import { parsePayload, sleep } from "./helpers/wait.js";
-import { focusWindow } from "../../src/engine/win32.js";
+import { restoreAndFocusWindow } from "../../src/engine/win32.js";
 
 let np: NpInstance;
 
 beforeAll(async () => {
   np = await launchNotepad();
-  try { focusWindow(np.hwnd); } catch { /* non-fatal */ }
+  try { restoreAndFocusWindow(np.hwnd); } catch { /* non-fatal */ }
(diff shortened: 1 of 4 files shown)


Harusame64:
@codex review

test-only fix: 4 e2e が win32.ts に存在しない focusWindow を import→undefined→throw→catch で握り潰し(CodeQL #129)。repo 共通の restoreAndFocusWindow(hwnd) に置換。確認軸: restoreAndFocusWindow が正しい置換か、np.hwnd(bigint) を受けるか、握り潰し no-op→実 focus への挙動変化が各テストの前提を壊さないか。

PR #418, opened 2026-05-25
fix(terminal): route conhost action=send (method:auto) through native console-paste

Harusame64:
## Problem

`terminal action=send` into a **conhost** (`ConsoleWindowClass`) window had two delivery failures for interactive raw/VT sessions (`ssh -tt` / vim / less / `python` REPL). Both reproduced this session via SSH-into-WSL bash dogfood (2026-05-25):

| send path | mechanism | conhost raw-session behaviour | root cause |
|---|---|---|---|
| `method:auto` | chunked **WM_CHAR** (`postCharsToHwnd`, chunkSize=100) | **characters DROP** | conhost executes at each embedded newline; chars posted while it is busy saturate/drop its input queue (`bg-input.ts:238-243`) |
| `method:foreground` | **Ctrl+V** (`typeViaClipboard`) | **silent no-op** (only trailing Enter lands) | Ctrl+V is a conhost accelerator honoured only in cooked/line-input mode; in raw/VT mode conhost forwards `^V` to the app |

### Evidence (SSH-into-WSL bash over conhost, 2026-05-25)
- BG WM_CHAR send of a ~230-char command echoed back corrupted (`-p 2222`→`-p 2`, `OpenSSH\ssh.exe`→`OpenSH\sh.exe`, `root@`→`rot@`); **re-read returned the SAME corruption (identical marker)** ⇒ send-side loss, not a read transient.
- `method:foreground` + Ctrl+V into the same `ssh -tt` bash: only `bash-5.3# ^M` landed (trailing Enter DID land, so not a focus failure). `preferClipboard:false` SendInput typing worked.

## Fix

Route conhost `method:auto` + `pressEnter:true` sends through the atomic native console paste (`pasteIntoConsoleNoFocus`, WM_COMMAND `0xFFF1`) — the same primitive `action=run` exit-mode already uses. It injects the whole buffer atomically, works in raw mode, and steals no foreground.

- **`shouldUseConsolePasteForSend(method, targetClass, pressEnter)`** — pure static gate: `auto` + `ConsoleWindowClass` + `pressEnter:true` only.
- **`background` stays on WM_CHAR** — preserves the #183 hidden-input verifyDelivery contract and the existing `channel:"wm_char"` envelope (and it is the never-clipboarded path for secret entry). **`foreground` unchanged** (Ctrl+V/typing — OQ-1).
- **Secret carve-out** — read the pre-send UIA baseline; skip console-paste for a credential prompt (`isSecretInputPrompt`, which **excludes** bash's PS2 bare-`>` continuation) since paste briefly puts text on the system clipboard. Unreadable (null/throw) baseline → safe default = WM_CHAR.
- **WM_CHAR fall-through** on `paste.ok:false` (native unavailable / UIPI block / post-paste failure) → no regression.
- Trailing newline stripped before paste (native always appends one Enter). `channel:"console_paste"` is additive (was `"wm_char"`).

## Tests

- **14 unit** (`tests/unit/terminal-send-console-paste.test.ts`): pure §3.1 matrix (5) + handler a–g (9): console-paste routing, paste-failure fall-through, `pressEnter:false`, trailing-newline strip, secret carve-out + bash-`>` non-secret + null-baseline safety, `method:'background'` stays `wm_char`, skippedFormats/restoreSkippedRace warnings.
- **Gated SSH-WSL load/stress gate** (`tests/e2e/terminal-console-paste-load.test.ts`): L1 (40× byte-intact, all `console_paste`, p50/p95/p99) / L2 (clipboard integrity + 100% delivery @250ms) / L3 (>1KB single-line contiguous) / L5 (no window-count growth).

### Load/stress results (real conhost SSH-into-WSL bash)
- **L1 PASS** — 40/40 byte-intact, 40/40 `channel:"console_paste"` (~260 ms/call native drain+gap floor).
- **L3 PASS** — >1KB single-line contiguous, no drop.
- **L5 PASS** — no window growth.
- **L2** — clipboard integrity PASS, 100% delivery @250 ms. Under a **pathological 15 ms write-hammer**, ~30% drop (14/20) — see OQ-4.

## Known limitation (OQ-4)
console-paste is clipboard-based, so a process **writing** the clipboard concurrently can clobber the value between our set and conhost's async read, dropping a fraction of sends (~30% under a pathological 15 ms write-hammer; **0% at a realistic 250 ms interval**). This is inherent to any clipboard paste and is **the same property the shipped `action=run` exit-mode path** (`pasteIntoConsoleNoFocus`, PR #393) already has. A pre-Enter `GetClipboardSequenceNumber` guard was evaluated and **rejected** (two Opus reviews): the WM_COMMAND paste is already posted by detect time, so a WM_CHAR fall-through would APPEND to a pre-loaded input line (garbage execution — strictly worse than a silent miss); the sequence number cannot see the async input-buffer ordering race; and there is no shell-agnostic safe line-clear. Ship = accept + document. Real clipboard managers POLL (read), rarely write, so realistic exposure is low.

## Scope / drift
Unchanged: `method:'background'` (WM_CHAR + #183), `method:'foreground'` (Ctrl+V), `pressEnter:false`, conhost `auto` at a hidden-input prompt (WM_CHAR), WT (`CASCADIA_HOSTING_WINDOW_CLASS`) and non-terminal targets. No new public tool, no `_errors.ts` change → catalog/stub unaffected.

## Plan
Opus 2-round plan review (PLAN-GO). Plan doc: [`desktop-touch-mcp-internal@944c32a:docs/terminal-send-conhost-console-paste-plan.md`](https://github.com/Harusame64/desktop-touch-mcp-internal/blob/944c32a86c2bcec13f0da3a48dbdeed9fd3c5df8/docs/terminal-send-conhost-console-paste-plan.md) (private; reviewers logged in).

### DIFF
--- a/src/tools/terminal.ts
+++ b/src/tools/terminal.ts
@@ -250,6 +250,38 @@ export function isSecretInputPrompt(baselineRaw: string | null): boolean {
   return SECRET_INPUT_PROMPT_PATTERNS.some((re) => re.test(line));
 }
 
+/**
+ * Static gate for routing a conhost `action=send` through the atomic native
+ * console-paste (`pasteIntoConsoleNoFocus`, WM_COMMAND 0xFFF1) instead of the
+ * chunked WM_CHAR loop.
+ *
+ * WM_CHAR drops characters on multiline / saturated conhost input
+ * (`bg-input.ts:238-243`), and the foreground Ctrl+V path is a silent no-op when
+ * the console is in raw/VT-input mode (interactive `ssh -tt` / vim / REPL —
+ * conhost forwards ^V to the app instead of pasting). The native console Paste
+ * injects the whole buffer atomically, works in raw mode, and steals no
+ * foreground — the same primitive `action=run` exit-mode already uses.
+ *
+ * Scope is `method:'auto'` ONLY. `background` deliberately stays on WM_CHAR so the
+ * #183 hidden-input verifyDelivery contract and the existing `channel:"wm_char"`
+ * envelope are preserved (and `background` is the never-clipboarded path for
+ * secret entry). `foreground` is unchanged (Ctrl+V/typing).
+ *
+ * Gated to `pressEnter:true` because the native console paste ALWAYS appends one
+ * Enter (`console_paste.rs:184-185`) — it is a paste-AND-run primitive.
+ *
+ * Pure (no Win32) so the static decision is unit-testable. The runtime secret
+ * carve-out (`isSecretInputPrompt` on the pre-send baseline) and the
+ * native-availability / paste-failure fall-through are applied at the call site.
+ */
+export function shouldUseConsolePasteForSend(
+  method: "auto" | "background" | "foreground" | "foreground_flash",
+  targetClass: string,
+  pressEnter: boolean,
+): boolean {
+  return method === "auto" && targetClass === "ConsoleWindowClass" && pressEnter;
+}
+
 function applySinceMarker(text: string, marker: string): { text: string; matched: boolean } {
   // Search for any tail window whose hash matches `marker`. Walk from the tail
   // backward — a recent terminal will hit within a few chars. Capped at 32k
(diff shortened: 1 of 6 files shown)


Harusame64:
@codex review

Production conhost input path. Please focus on: (1) the new console-paste branch inside the `useBg` block — does it correctly fall through to WM_CHAR on every `paste.ok:false` reason and on the secret/null-baseline carve-out; (2) the response-shape contract — `channel:"console_paste"` is additive, confirm no field existing callers/tests branch on (e.g. `verifiedDelivery`, `focusLost`, `sent` semantics) is silently dropped vs the WM_CHAR path; (3) `isSecretInputPrompt` vs `isHiddenInputPrompt` selection and the bash PS2 bare-'>' edge; (4) trailing-newline strip / double-Enter; (5) #183 background contract untouched.

Harusame64:
@codex review

Round 2 fix applied for your P1 (trailing-newline count). `src/tools/terminal.ts` now strips EXACTLY ONE trailing line break (`/(?:\r\n|\r|\n)$/`) instead of all, so the native single-Enter re-add yields N trailing Enters for N>=1 and 1 for N=0 — matching the WM_CHAR path's max(N,1). New unit cases (d2) N>=2 preserved and (d3) trailing CRLF stripped-whole; unused `mockEnter` removed (CodeQL). Please re-check the trailing-newline semantics and confirm no new edge case (empty input / newline-only input).

PR #419, opened 2026-05-25
fix(terminal): send Enter as a real key event so conhost PowerShell executes the command

Harusame64:
## Problem (found by dogfood of v1.9.1)

Sending into a **conhost (`ConsoleWindowClass`) PowerShell** window via `terminal action=send method:auto` (the v1.9.1 console-paste path) **typed the command but never ran it**, leaving a stray `m` on the input line:

```
PS C:\> echo hi_world_m        ← typed, NOT executed, trailing 'm'
```

Confirmed against ground truth (screenshot + a real foreground Enter that then executed the pending line; both UIA and OCR read-back showed the literal `m`).

## Root cause

The final Enter was sent as `WM_CHAR 0x0D`. **`0x0D` is the Ctrl+M control code** — PowerShell's **PSReadLine renders a `WM_CHAR 0x0D` as a literal `m`** and never treats it as accept-line. `bash`/`cmd` read it as CR and execute, which is why the original conhost-**bash** fix and the SSH-WSL gates passed while **PowerShell silently failed**.

The same `WM_CHAR` Enter is shared by the WM_CHAR path (`postEnterToHwnd`) and was **also the cause of the previously "flaky" local PowerShell exit-mode timeout** — the `until:{mode:'exit'}` epilogue never executed, so the sentinel never rendered and the run timed out. It was a real bug masquerading as a flake.

## Fix

Send Enter as a real **KEY EVENT** (`WM_KEYDOWN` + `WM_KEYUP` `VK_RETURN` with the Enter scan code), in both:
- the native console-paste (`src/win32/console_paste.rs`), and
- the TS `postEnterToHwnd` (`src/engine/bg-input.ts`, now delegates to the existing `postKeyToHwnd` helper).

PSReadLine accepts a `VK_RETURN` key event as accept-line; `bash`/`cmd` still see CR, so the original conhost-bash fix **and** exit-mode are preserved. Works in the background (`PostMessage`, no foreground steal).

The mechanism was de-risked before the native change: a background `WM_KEYDOWN/WM_KEYUP VK_RETURN` executed a pending line in conhost PowerShell with no stray char.

## Test hardening

The conhost multiline e2e now asserts **EXECUTION**, not delivery: a run marker appears **twice** (the `echo <marker>` command echo **+** the `<marker>` output line). The previous `toContain(marker)` matched the *typed-but-un-executed* echo and masked this exact bug — the dogfood gate caught what the test could not.

## Verification (local conhost + SSH-WSL, native rebuilt)

| Gate | Result |
|---|---|
| `terminal.test.ts` conhost "EXECUTES every line" | ✅ (was masking the bug) |
| `terminal-exit-mode.test.ts` | ✅ 12/12 — `[powershell]` now passes (the "flake"), `[bash@wsl-ssh]` still passes (no regression) |
| `terminal-console-paste-load.test.ts` | ✅ 4/4 (bash send under load/contention) |
| `terminal.test.ts` full | ✅ 31 passed / 1 skipped |
| console-paste routing unit | ✅ 16/16 |

## Scope / regression
- Native console-paste Enter + TS `postEnterToHwnd` only. No routing/contract change.
- bash/cmd behaviour preserved (CR via key event = canonical Enter); also fixes the pre-existing conhost-PowerShell exit-mode timeout.
- Native binary is rebuilt by CI from source (`console_paste.rs`); the `.node` is gitignored.

Targets a v1.9.2 patch. The v1.9.1 CHANGELOG "conhost windows reliably" claim becomes accurate for PowerShell with this fix.

Plan: [`desktop-touch-mcp-internal@3dd30bd:docs/terminal-send-conhost-console-paste-plan.md`](https://github.com/Harusame64/desktop-touch-mcp-internal/blob/3dd30bd5626fb0104602afdd1308a33bd3108f56/docs/terminal-send-conhost-console-paste-plan.md)

### DIFF
--- a/src/engine/bg-input.ts
+++ b/src/engine/bg-input.ts
@@ -229,8 +229,8 @@ export function postKeyToHwnd(hwnd: unknown, vk: number): boolean {
  * (the command is typed but not run), whereas a VK_RETURN key event is
  * recognized as accept-line. bash / cmd accept the key-event Enter as CR exactly
  * as they did the WM_CHAR form, so terminal command execution is preserved.
- * Functionally equivalent to the native console-paste Enter (`console_paste.rs`)
- * — identical low-32 keystroke lParam (Win32 reads a keystroke lParam as 32-bit).
+ * Bit-identical lParam with the native console-paste Enter (`console_paste.rs`),
+ * which sign-extends its keyup lParam to match this path.
  */
 export function postEnterToHwnd(hwnd: unknown): boolean {
   return postKeyToHwnd(hwnd, VK_RETURN);
--- a/src/engine/bg-input.ts
+++ b/src/engine/bg-input.ts
@@ -229,7 +229,8 @@ export function postKeyToHwnd(hwnd: unknown, vk: number): boolean {
  * (the command is typed but not run), whereas a VK_RETURN key event is
  * recognized as accept-line. bash / cmd accept the key-event Enter as CR exactly
  * as they did the WM_CHAR form, so terminal command execution is preserved.
- * Bit-equal with the native console-paste Enter (`console_paste.rs`).
+ * Functionally equivalent to the native console-paste Enter (`console_paste.rs`)
+ * — identical low-32 keystroke lParam (Win32 reads a keystroke lParam as 32-bit).
  */
 export function postEnterToHwnd(hwnd: unknown): boolean {
   return postKeyToHwnd(hwnd, VK_RETURN);
--- a/src/engine/bg-input.ts
+++ b/src/engine/bg-input.ts
@@ -223,12 +223,16 @@ export function postKeyToHwnd(hwnd: unknown, vk: number): boolean {
 }
 
 /**
- * Send Enter to `hwnd` via WM_CHAR '\r'.
- * Preferred over postKeyToHwnd(VK_RETURN) for terminals (WT/conhost normalise '\r').
+ * Send Enter to `hwnd` as a real KEY EVENT (WM_KEYDOWN + WM_KEYUP VK_RETURN),
+ * NOT WM_CHAR '\r'. A WM_CHAR 0x0D is the Ctrl+M control code: conhost
+ * PowerShell's PSReadLine renders it as a LITERAL 'm' and never accepts the line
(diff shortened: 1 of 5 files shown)


Harusame64:
@codex review

Native input-path fix: the final Enter for conhost console-paste send is changed from WM_CHAR 0x0D to a real key event (WM_KEYDOWN+WM_KEYUP VK_RETURN) because PSReadLine renders WM_CHAR 0x0D (Ctrl+M) as a literal 'm' and never accepts the line. Please focus on: (1) the lParam bit math in `console_paste.rs` (scan code 0x1C, keyup previous-state bit 30 + transition bit 31, `1isize<<31` on win32-x64), (2) the removed `WM_CHAR` const has no remaining use, (3) `postEnterToHwnd` -> `postKeyToHwnd` keeps bash/cmd CR semantics (no regression to the original conhost-bash fix or action=run exit-mode), (4) the native vs TS Enter staying bit-equal, (5) the e2e now asserting execution (marker count >=2) not delivery.

Harusame64:
@codex review

Applied your P2 fix in 18a1b89: the native keyup lParam is now built as `u32 -> i32 -> isize`, so it sign-extends to `0xFFFFFFFFC01C0000` exactly like the TS `postKeyToHwnd` path (`lParam | 0` -> i64 widening). Native and TS Enter lParam are now bit-identical (low + high 32 bits), so signed checks like `lParam < 0` behave the same across both paths. Comments updated to state true bit-identity. conhost "EXECUTES" e2e re-verified green after the native rebuild.

Harusame64:
@codex review

Both P2s addressed:
- **Sign-extend KEYUP lParam** (18a1b89): native keyup lParam now built u32→i32→isize, bit-identical to the TS postKeyToHwnd path (0xFFFFFFFFC01C0000).
- **Align Enter-injection tests** (99a3ff5): bg-input.test.ts postEnterToHwnd now asserts KEYDOWN+KEYUP VK_RETURN (not WM_CHAR 0x0D) + a regression guard against reverting to WM_CHAR; refreshed the stale keyboard-bg-verification comment.

bg-input unit 25/25 green; conhost EXECUTES e2e green after native rebuild.

PR #425, opened 2026-06-03
feat(desktop_act): S2 gate plumbing + visual-only flag persistence (ADR-024 Seed-2)

Harusame64:
## What (ADR-024 Seed-2, Phase S2 — gate plumbing + visual-only flag persistence)

> **Stacked on #424 (S1).** Base = `feature/adr-024-s1-roicapture-contract`. Review/merge after S1. Still **bit-equal** — population is a stub until S3.

- **Visual-only flag persistence (discover → act)**: `SessionState.lastDiscoverVisualOnly`, written by `DesktopFacade.see()` from the discover `warnings` (membership in `UIA_BLIND_WARNINGS`, now **exported** from `compose-providers` so it's the *same SSOT predicate* as the OCR lane's `uiaBlindForOcr` — they cannot diverge). `resolveVisualOnlyForViewId()` reads it with a safe `false` default.
- **Gate wiring**: `desktopActRawHandler` calls `buildRoiCapture()` on success → runs `shouldReturnRoiCapture` **reusing the motion verdict from the observation poll already done above (no extra DXGI poll, acceptance ④)**. `buildRoiCapture` is the **S3/S4/S5 population seam**: in S2 it returns `undefined` whenever the gate passes (no ROI source yet) → responses stay bit-equal.

## Verification
- `npm run build` clean; flag-persistence suite **6/6** (`tests/unit/roi-capture-flag-persistence.test.ts`).
- Full unit suite green except **3 pre-existing load-sensitive timing flakes** (benchmark idle / ui-pattern debounce / winevent sidecar) — all pass in isolation, unrelated subsystems.

## Scope notes for review
- **Byte-equality**: structurally guaranteed in S2 — `buildRoiCapture` returns `undefined` unconditionally (stub tail). End-to-end act-handler roiCapture-absence gets real coverage in S5 when population lands.
- The visual-only flag derives from the **same `rawResult.warnings`** that the existing (shipped) `deriveViewConstraints` consumes, so it's exactly as correct as the production `constraints.uia` signal.
- This PR contains the **visual-only persistence that the sub-plan moved S1 → S2** (per #424 Opus R1).

## Review
production change → Opus + Codex per CLAUDE.md §3.3 (session-contract extension = Codex axis).

Plan: `desktop-touch-mcp-internal@plan/adr-024-seed2:docs/adr-024-seed2-plan.md` (S2)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #424, opened 2026-06-03
feat(desktop_act): S1 contract lock for ADR-024 Seed-2 roiCapture

Harusame64:
## What (ADR-024 Seed-2, Phase S1 — contract lock)

Locks the post-action **`roiCapture`** contract surface for the **visual-only regime** (UIA-blind / RDP / canvas targets). **Value is not yet populated** — this PR only fixes the shape so later phases can fill it without a breaking change.

- `TouchResult` success variant gains optional `roiCapture { roi, somImage, entities, source }` + `RoiPreviewEntity` (lease-less preview, MVP = OQ-8 option **(b)**). Mirrors the existing additive `observation?` layering (populated by the registration wrapper, not the bare loop). Absent today → existing `{ok, executor, diff, next}` responses are **bit-equal**.
- `desktop_act` input schema + tool description gain optional `returnCapture` (`on-change | always | never`). Accepted now so callers can opt in; **wired in S2**.
- New **pure** gate `shouldReturnRoiCapture()` with hard guards (`ok` + `visualOnly`; the visual-only guard protects `feedback_minimize_screenshot_reliance`). **Exhaustively unit-tested over the 5-valued motion enum** (18 tests, `tests/unit/roi-capture-gate.test.ts`).

## Verification
- `npm run build` clean; new gate suite 18/18; full unit suite green except one **pre-existing environmental flake** (`ui-pattern-store-persistence` temp-dir rename ENOENT — passes in isolation, unrelated subsystem).
- `check:failwith-fixtures` clean (no failWith callsites added).

## ⚠ Phase-boundary note for review
The sub-plan put the **visual-only-flag session persistence** in S1. I **deferred it to S2**: its only consumer is the S2 gate wiring, so persisting it here would be unconsumed infra (dead code). The pure gate takes `visualOnly` as a param and is fully testable without it. **Please confirm this phase-boundary refinement** or push back.

## Review
production change → Opus + Codex per CLAUDE.md §3.3. Additive only (new optional field + new optional param + new pure fn).

Plan: `desktop-touch-mcp-internal@plan/adr-024-seed2:docs/adr-024-seed2-plan.md` (S1)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #426, opened 2026-06-04
feat(desktop_act): S3a — surface DXGI dirty Rect[] from the motion poll (ADR-024 Seed-2)

Harusame64:
## ADR-024 Seed-2 — S3a (ROI source, part 1: dirty-rect surface)

Walking-skeleton phase **S3a** for visual-only post-action ROI auto-crop. Stacks on S1 (#424) + S2 (#425, both merged).

**Goal (sub-plan §2 S3a, OQ-11):** let the single `verifyAnyChange` DXGI poll that already produces the post-act `motion` verdict ALSO surface the raw dirty `Rect[]`, so the ROI-capture path reuses that poll with **zero extra DXGI acquires** (no double-poll).

### Changes (additive, opt-in, byte-equal by default)
- **`VisualMotionObservation.dirtyRects?: Rect[]`** — raw per-output dirty rects in **desktop screen-absolute coords**, **NOT yet window-rect filtered** (that lands in S3b). Populated only for `source: "dxgi_dirty_rect"` with ≥1 rect.
- **`VerifyAnyChangeOpts.includeDirtyRects?: boolean`** (default `false`). Reuses the `rects` already drained in the same poll → +0 DXGI acquire.

### Why opt-in (key design decision — please scrutinize)
`verifyAnyChange` has **three** consumers: the `desktop_act` ROI path (this PR), plus the keyboard Stage-4 and mouse-verify Stage-5 DXGI fallbacks. The latter two **serialize the full observation** into their tool responses. Unconditional population would therefore **leak `dirtyRects` into those responses** and break their byte-equality. The opt-in flag confines the rects to the `desktop_act` ROI path, so **every response stays byte-equal until S5** folds the crop into `roiCapture`.

### Scope discipline
Orchestrator capability **only**. The handler / `buildRoiCapture` wiring is deferred to **S3b**, where the window-rect intersection filter that actually *consumes* the rects lands — so there is no dead code or wiring-ahead here (same incremental pattern as S1/S2). In S3a the rects are surfaced + unit-proven; no production consumer reads the field yet.

### Tests
5 new `any-change-orchestrator` cases:
- opt-in → rects surfaced on `any_change` from the same observation
- opt-in → sub-threshold `no_change` still surfaces rects
- opt-in → RAW per-output rects (in-window + out-of-window both pass verbatim; filtering is S3b)
- default (no flag) → `dirtyRects` absent (byte-equal for existing callers)
- empty rects → omitted even when opted in

Local: full suite **4170 passed / 0 failed**; `check:failwith-fixtures` + `check:stub-catalog` clean; `tsc` clean.

Plan: `desktop-touch-mcp-internal@39bfe346:docs/adr-024-seed2-plan.md` (§2 S3a)

### DIFF
--- a/src/engine/any-change.ts
+++ b/src/engine/any-change.ts
@@ -34,6 +34,7 @@
  */
 
 import type { VisualMotionObservation } from "../tools/_input-pipeline.js";
+import type { Rect } from "./vision-gpu/types.js";
 import {
   BROKER_CONSTANTS,
   type CacheAcquireState,
@@ -171,6 +172,15 @@ export interface VerifyAnyChangeOpts {
   broker?: DirtyRectBroker | null;
   /** @internal — test-only override for `enumMonitors`. */
   enumerate?: () => Array<{ bounds: { x: number; y: number; width: number; height: number } }>;
+  /**
+   * ADR-024 Seed-2 S3a — opt-in to surface the raw DXGI dirty `Rect[]` on the
+   * returned `VisualMotionObservation.dirtyRects` (screen-absolute, per-output,
+   * NOT yet window-rect filtered). Reuses the same poll that produces `motion`,
+   * so it adds **zero** extra DXGI acquires. Default `false` (omitted) — every
+   * existing caller (keyboard / mouse-verify Stage-5 fallback) keeps a
+   * byte-equal observation. Only the `desktop_act` visual-only ROI path opts in.
+   */
+  includeDirtyRects?: boolean;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


Harusame64:
@codex review

Harusame64:
## Round 1 — Opus review: GO (P1=0, P2=0)

Opus phase-boundary review passed with **0 P1 / 0 P2**. byte-equal (opt-in confines `dirtyRects` to the act ROI path; keyboard/mouse-verify observations stay byte-equal), single-poll reuse, additive-optional, and honest scope (capability-only, handler wiring deferred to S3b) all structurally satisfied + test-pinned.

Two P3 nits, both addressed in `31926c9d`:
- **P3-1** (plan↔impl field-name drift): sub-plan §2 S3a said `rects?`, impl ships `dirtyRects?`. Synced the private sub-plan (`desktop-touch-mcp-internal@8d6d315`) to `dirtyRects?` + the opt-in rationale; marked S3a done.
- **P3-2** (test rigor): added `brokerWithFactorySpy` + a test asserting the broker factory is invoked **exactly once** when opting in — substantiates acceptance ① "+0 extra acquire" with a measurement, not just structure.

Awaiting @codex review.

PR #431, opened 2026-06-05
feat(desktop_act): S5c-1a — frame-diff motion for visual-only acts (ADR-024 dogfood F1/F2 fix)

Harusame64:
## What

ADR-024 Seed-2 **S5c-1a**. A manual live dogfood of the shipped S5 `roiCapture` surfaced two correctness defects on a real (busy) desktop. This routes the visual-only post-action motion verdict through a **true window frame-diff** instead of DXGI dirty rects, fixing both.

## Why (dogfood findings, Opus-reviewed)

- **F1 — occlusion-blind geometry filter**: `filterDirtyRectsToWindow` is pure geometry / z-order-blind, and DXGI dirty rects are per-output, so background activity (e.g. a video) that overlaps the opaque, top-most target window's rect was passed through → spurious `motion=any_change` on a static window + a ROI locked onto a noise sliver.
- **F2 — action change missed**: the same-process vision-gpu `DirtyRectRouter` fan-out drains the DXGI frame (drain-once) before the act's `sub.next()` cursor is placed → the action's real change read as `no_change`.

Root-caused + reviewed across 3 Opus rounds; see `adr-024-seed2-dogfood-findings.md` and the S5c design in the plan (private repo).

## How

- Capture a pre-action frame (`captureFrame` / PrintWindow) **before** the touch, gated on the visual-only flag → structured targets pay nothing and keep their DXGI observation **byte-equal** (ADR-019 Stage 5 telemetry contract).
- After the touch, reuse `verifyLocalRepaint` (caller pre-frame + settle + native-SIMD `computeChangeFraction`/SSIM). PrintWindow captures only the target window's pixels (**occlusion-immune → F1**); no DXGI broker dependency (**→ F2 drain race gone**). Its background-animation guard degrades to `indeterminate`, which the gate excludes.
- Focal `hint.point` = the clicked entity's screen-absolute centre (new `DesktopFacade.resolveEntityCenterForViewId`) so a small localized repaint isn't diluted across the whole window into a false `indeterminate`.
- `roiCapture.source` is `"frame_diff"`; ROI is provisionally the whole window (empty dirtyRects → full-window fallback) until the **S5c-1b** bbox.

Non-visual-only acts are unchanged (DXGI `tryVerifyAnyChange`, byte-equal).

## Verification

Re-dogfooded on a self-contained visual-only canvas (single full-bleed Panel = `uia_blind_single_pane`, click paints a localized rectangle):

| scenario | before (S5/DXGI) | after (S5c-1a/frame-diff) |
|---|---|---|
| clean bg + change | `no_change` (missed, F2) | **`local_repaint`** + roiCapture(frame_diff) |
| busy bg + change | noise ROI (F1) | **`local_repaint`** (change caught despite overlap) |
| busy bg + no change | `any_change` (spurious, F1) | **`no_change`** + roiCapture absent |

Build clean; unit suite green (new `resolveEntityCenterForViewId` tests + existing roiCapture gate/flag tests).

## Scope / follow-ups (this is S5c-1a of a 3-part fix)

- **S5c-1b**: native bbox aggregation (SIMD scan) → localized ROI instead of full-window; `source` branch for the BitBlt-fallback (occlusion-inclusive) case → full-window degrade.
- **S5c-2**: occlusion/race regression tests (live desktop).

Known 1a limitation: `verifyLocalRepaint`'s `MAX_RECT_AREA_PX` cap means very large visual-only windows degrade to `indeterminate` → roiCapture absent (relaxed in 1b).

Plan: `desktop-touch-mcp-internal@ce5d565:docs/adr-024-seed2-plan.md` (S5c-1a)

### DIFF
--- a/src/tools/desktop-register.ts
+++ b/src/tools/desktop-register.ts
@@ -622,26 +622,35 @@ export const desktopActRawHandler = async (
   let postVerify: { observation: VisualMotionObservation; dirtyRects: Rect[] } | null = null;
   let frameDiffObservation: VisualMotionObservation | undefined;
   if (result.ok && postVerifyEnabled) {
-    if (visualOnly && preFrame !== null && frameDiffHwnd !== null && frameDiffWindowRect !== null) {
-      // Visual-only — frame-diff motion via the existing Stage 4 orchestrator
-      // (caller pre-frame + capturePostFrameUntilStable settle + native SIMD
-      // computeChangeFraction/SSIM). Its background-animation guard degrades to
-      // `indeterminate`, which the gate excludes (so a noisy desktop yields no
-      // spurious roiCapture = F1). `observation.source` becomes `ssim_residual`
-      // (frame-diff family) on this path only; non-visual stays `dxgi_dirty_rect`.
-      frameDiffObservation = await verifyLocalRepaint({
-        hwnd: frameDiffHwnd,
-        hint: {
-          windowRect: frameDiffWindowRect,
-          ...(frameDiffPoint !== null && { point: frameDiffPoint }),
-        },
-        preFrame,
-      });
-      (result as { observation?: VisualMotionObservation }).observation = frameDiffObservation;
+    if (visualOnly) {
+      // Visual-only — frame-diff ONLY. Never fall back to the DXGI verifier:
+      // DXGI is exactly the occlusion-blind / drain-race path this phase replaces,
+      // so a visual-only frame-diff *miss* (capture/hwnd/rect unavailable) must
+      // degrade to no observation — NOT reintroduce F1/F2 via DXGI (Codex PR #431
+      // round 2 P2). With no observation the gate sees `motion=undefined` and
+      // declines (except `returnCapture:"always"`, whose full-window fallback in
+      // buildRoiCapture still applies — a best-effort capture, never DXGI motion).
+      if (preFrame !== null && frameDiffHwnd !== null && frameDiffWindowRect !== null) {
+        // Stage 4 orchestrator: caller pre-frame + capturePostFrameUntilStable
+        // settle + native SIMD computeChangeFraction/SSIM. Its background-animation
+        // guard degrades to `indeterminate`, which the gate excludes (so a noisy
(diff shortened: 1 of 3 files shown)


Harusame64:
@codex review

Harusame64:
Addressed the P2 (frame-diff window resolution) in `3e59daec`.

`verifyLocalRepaint` now resolves the target window via a new `DesktopFacade.resolveTargetHwndForFrameDiff`, which maps `windowTitle → hwnd` through the same `resolveWindowTarget` the discover providers use — so the pre/post frames come from the exact window discover enumerated and the executor routes its click to, independent of foreground. (The production viewport guard already rejects acting on a non-foreground visual-only entity, so this is defense-in-depth for that path; the explicit target resolution is the correct invariant and is robust to window-title collisions where foreground ≠ the discover-resolved window.)

Dogfood re-verified the normal foreground case still yields `local_repaint` + `roiCapture(frame_diff)`; build clean, affected unit suites green.

@codex review

Harusame64:
Both round-2 P2s addressed in `b8e813ad`.

- **P2-A (no DXGI fallback for visual-only misses)**: the post-touch dispatch now branches on `visualOnly` alone. A visual-only frame-diff miss degrades to *no observation* (gate declines; `returnCapture:"always"` still gets buildRoiCapture's best-effort full-window fallback, never DXGI motion). DXGI runs only for non-visual-only targets.
- **P2-B (real title resolver)**: `resolveTargetHwndForFrameDiff` now resolves plain titles via the `findPlainTopLevelWindowByTitle` SSOT (the same top-level-by-title predicate the discover title flow uses) before `resolveWindowTarget` (@active/dialog) / foreground — so the common `desktop_discover({windowTitle})` flow no longer falls through to foreground.

Same-axis hardening: buildRoiCapture's SoM crop uses the same resolver (ROI image == motion window), and `roiCapture.source` is regime-determined (`visualOnly ? "frame_diff" : "dxgi"`).

Dogfood re-verified normal foreground → `local_repaint` + `roiCapture(frame_diff)`; build clean, affected unit suites green (109).

@codex review

PR #444, opened 2026-06-06
fix(mouse): preserve screenshot-time window position for homing delta

licat2023:
## Summary

Two-layer fix for homing delta correction being silently disabled:

### Layer 1: Preserve screenshot-time position within applyHoming

\pplyHoming\ saves the cached window region before Tier 2 (focus) mutates the cache. Delta is then computed from the saved position vs live \GetWindowRect\, not from the freshly-overwritten cache.

### Layer 2: Snapshot cache to survive external cache mutations

Added \saveSnapshot\ / \getSnapshot\ — a separate write-once cache populated by screenshot tools. Unlike the main cache, the snapshot cache is **never mutated** by \ocus_window\, \window_dock\, or Tier 2's \updateWindowCache\. \pplyHoming\ reads from the snapshot cache first.

Together these fix both:
1. Tier 2 self-nullifying the delta within \pplyHoming\ itself
2. External tools (\ocus_window\/docking) nullifying the cache between screenshot and click

## Changes

- \src/engine/window-cache.ts\: add snapshot cache + \saveSnapshot\/\getSnapshot\
- \src/tools/mouse.ts\: \pplyHoming\ uses \getSnapshot\ first, falls back to \getCachedWindowByTitle\; cache miss on known title refreshes cache
- \src/tools/screenshot.ts\: \detail='text'\ and \detail='ocr'\ call \saveSnapshot\ after screenshot

## Testing

Verified on Windows with Doubao desktop app:
1. Screenshot via \screenshot(detail='ocr', windowTitle='豆包')\ at position (1077, 354)
2. Moved window to (8, 8) via \window_dock\
3. \mouse_click(1166, 867, windowTitle='豆包')\ → homing: \window moved -1069,-346\ → correctly clicked target

### DIFF
--- a/src/engine/window-cache.ts
+++ b/src/engine/window-cache.ts
@@ -34,11 +34,50 @@ const cache = new Map<string, CachedWindow>();
 
 /** Cache entries older than this are treated as stale — HWND may have been recycled. */
 const CACHE_TTL_MS = 60_000;
+
+/**
+ * Snapshot cache: persists screenshot-time window positions by title.
+ * Separate from the main cache — NOT mutated by updateWindowCache(),
+ * focus_window(), or window_dock(). Only screenshot tools write to it,
+ * and only applyHoming reads from it.
+ *
+ * This guarantees that mouse_click's homing correction always compares
+ * against the position the LLM saw in the screenshot, even when other
+ * tools have overwritten the main cache between screenshot and click.
+ */
+const snapshotCache = new Map<string, { region: { x: number; y: number; width: number; height: number }; timestamp: number }>();
+const SNAPSHOT_TTL_MS = 90_000;
+
 export const WINDOW_CACHE_TTL_EXPORTED_MS = CACHE_TTL_MS;
 
 /** Get the timestamp this hwnd was last cached, or null if not cached. */
 export function getWindowCacheTimestamp(hwnd: bigint): number | null {
   return cache.get(String(hwnd))?.timestamp ?? null;
+}
+
+/**
+ * Save a screenshot-time window position to the snapshot cache.
+ * Call from screenshot tools after capturing a single-window screenshot.
+ * The snapshot survives mutations to the main cache from focus/dock tools.
+ */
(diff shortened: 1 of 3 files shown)


Harusame64:
Thanks so much for this — what a fantastic catch! 

You were spot on about applyHoming's Tier 2 running updateWindowCache() before Tier 1 computed the delta, causing the screenshot-time position to be overwritten and the homing correction to silently collapse to (0,0).

Actually, your fix covers even more ground than the original issue described! Since the pre-Tier-2 updateWindowCache() was running on every windowTitle click (not just during focus attempts), the delta was being nullified across the board. The two-layer approach you implemented (preserving the screenshot-time region + a focus/dock-immune snapshot cache) is a really solid solution.

I'll go ahead and merge this in. To build on top of your great work, I've opened a small follow-up PR to handle a few extra edge cases and tests we spotted during review:

Hooked up the new getSnapshot export to the existing mouse unit-test mocks (to keep the foreground-refusal contract tests happy!).

Ensured the primary coordinate-copy path (screenshot(dotByDot=true) / detail='image' single-window capture) seeds the snapshot/main cache, allowing homing to engage there just like it does for text/ocr.

Added recycle guards so the cachedBefore / snapshot fallbacks honor the 60s cache TTL.

Added a few extra regression tests for the snapshot cache and the delta path.

Thanks again for the incredibly detailed issue and repro—it made reviewing this a breeze!

Harusame64:
Shipped in **v1.10.2** — now live on all three channels:

- npm: `npx -y @harusame64/desktop-touch-mcp@1.10.2` (latest)
- GitHub Release: https://github.com/Harusame64/desktop-touch-mcp/releases/tag/v1.10.2
- MCP Registry: `io.github.Harusame64/desktop-touch-mcp` 1.10.2

Your homing-delta fix is in users' hands. The follow-up #445 closed the review gaps (test mocks, the `screenshot(dotByDot=true)` / `detail='image'` cache seeding, and the 60s stale/recycle TTL guards), and #446 added an integration regression pin that drives the real window-cache. Thanks again for the great catch + repro. 🚀

Issue #443, opened 2026-06-06
mouse_click: homing delta correction silently disabled when window focus triggers cache overwrite


licat2023:
## Problem

In \src/tools/mouse.ts\, the \pplyHoming\ function has a logic ordering bug where Tier 2 (focus) calls \updateWindowCache()\ **before** Tier 1 (delta correction) runs \computeWindowDelta()\. Since \updateWindowCache\ overwrites the cached window position with the **current** position, \computeWindowDelta\ always computes a delta of \(0,0)\ — effectively disabling homing correction whenever a focus attempt was made.

## Reproduction

1. Take a screenshot via \screenshot(detail='text', windowTitle='SomeApp')\ — this caches the window position at screenshot time
2. Move the window to a different screen position
3. Call \mouse_click(x, y, windowTitle='SomeApp')\ where \(x,y)\ are the screenshot's click coordinates

**Expected**: homing detects the window moved and applies \(dx, dy)\ offset to the click coordinates.  
**Actual**: Tier 2 runs \updateWindowCache\ (line 115 or 161), overwrites the screenshot-time position. Tier 1 then reads the freshly-overwritten cache, computes delta \(0,0)\, and clicks the wrong position.

## Root Cause

In \pplyHoming\ (mouse.ts ~line 112-207):

\\\	ypescript
// Tier 2: focus window
if (windowTitle) {
  const windows = enumWindowsInZOrder();
  updateWindowCache(windows); // ← OVERWRITES screenshot-time position!
  // ... focus logic ...
  updateWindowCache(postWindows); // ← OVERWRITES again!
}

// Tier 1: delta correction
const cached = getCachedWindowByTitle(windowTitle);
// ← cache now has CURRENT position (just overwritten)
const delta = computeWindowDelta(cached.hwnd);
// ← always (0,0) because current == cached
\\\

## Proposed Fix

Save the screenshot-time window region before Tier 2 mutates the cache, then compute delta from the saved position against the live \GetWindowRect\:

\\\	ypescript
// Before Tier 2: save snapshot position
let screenshotRegion = null;
if (windowTitle) {
  const cachedBefore = getCachedWindowByTitle(windowTitle);
  if (cachedBefore) screenshotRegion = { ...cachedBefore.region };
}

// ... Tier 2 focus (may call updateWindowCache) ...

// After Tier 2: compute delta from SAVED position
if (screenshotRegion && windowTitle) {
[...]

PR #445, opened 2026-06-06
fix(mouse/screenshot): close homing snapshot gaps (follow-up to #444 / issue #443)

Harusame64:
Follow-up to #444 (issue #443). #444 correctly fixed the homing root cause
(Tier 2's `updateWindowCache()` ran before Tier 1's delta computation, so the
screenshot-time position was overwritten and the delta collapsed to `(0,0)`).
This PR closes the gaps found while reviewing #444 together with Codex + Opus.

## What this fixes

- **Broken existing tests (CI red after #444).** The new `getSnapshot` export
  was not added to the wholesale `window-cache` mocks in the mouse unit tests,
  so `applyHoming` threw and the foreground-refusal contract tests regressed to
  `ToolError`. Added `getSnapshot` to the three mouse-test mocks.
- **Primary coordinate-copy path was uncovered.** `screenshot(dotByDot=true)` /
  `detail='image'` single-window capture never seeded the snapshot or the main
  cache, so homing could not resolve a live HWND there — i.e. the fix was
  absent for the scenario it most needs to protect. Now seeds both, matching
  the `detail='text'` / `detail='ocr'` paths.
- **Stale-cache guards.** The `cachedBefore` fallback copied a region without
  honouring the 60s cache TTL, and the snapshot delta path read `GetWindowRect`
  on a title-resolved HWND without any TTL/recycle guard. Both now apply the
  `WINDOW_CACHE_TTL_EXPORTED_MS` guard so a recycled HWND / minute-old layout
  can't produce a bogus offset.

## Also

- Drop the unused `saveSnapshot` import in `mouse.ts`.
- Regenerate the failWith callsite fixtures for the shifted line numbers.
- Document `getSnapshot`'s exact (vs substring) title-match semantics.

## Tests

- New `tests/unit/window-cache-snapshot.test.ts` — snapshot save / get / 90s TTL.
- New `tests/unit/homing-snapshot-delta.test.ts` — pins the issue #443 delta
  path: snapshot delta, fresh-cache fallback, stale-cache skip, and the
  HWND-recycle guard.
- All related unit suites green; `build` / `check:stub-catalog` /
  `check:failwith-fixtures` clean.

## Review

Reviewed with Codex (no regressions) and Opus (P1 zero; the snapshot-path
recycle guard was added in response to Opus round 1). Residual lower-priority
follow-ups (integration-level ordering pin, snapshot/main TTL coherence,
cross-call title-key match) are recorded in the internal docs repo.

### DIFF
--- a/src/engine/window-cache.ts
+++ b/src/engine/window-cache.ts
@@ -68,6 +68,13 @@ export function saveSnapshot(title: string, region: { x: number; y: number; widt
 /**
  * Read a saved screenshot-time position for a given window title.
  * Returns null if never saved or expired (TTL > 90s).
+ *
+ * Matching is an exact (case-insensitive) key lookup, NOT the substring match
+ * used by getCachedWindowByTitle. A snapshot is keyed by the screenshot's
+ * effectiveTitle and read back by the mouse_click windowTitle; these coincide
+ * for the normal workflow (same title string passed to both), so a caller that
+ * uses different title strings for the screenshot and the click simply misses
+ * here and degrades to the main-cache path — no incorrect correction results.
  */
 export function getSnapshot(title: string): { x: number; y: number; width: number; height: number } | null {
   const key = title.toLowerCase();
--- a/src/tools/mouse.ts
+++ b/src/tools/mouse.ts
@@ -204,9 +204,14 @@ async function applyHoming(
   let delta: { dx: number; dy: number; sizeChanged: boolean } | null = null;
 
   if (screenshotRegion && windowTitle) {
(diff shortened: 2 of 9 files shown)


Harusame64:
@codex review

Harusame64:
@codex review

Round 2: applied the P2 — added `WINDOW_CACHE_TTL_EXPORTED_MS` to the three mouse test mocks (`mouse-click-coord-order`, `issue-207-foreground-refusal-mouse`, drag variant) so the mock surface matches mouse.ts's import. Commit be5a1a9b.

PR #446, opened 2026-06-07
test(mouse): integration regression pin for homing #443 + document snapshot TTL split

Harusame64:
Follow-up R1/R2 to the homing snapshot work (#444 / #445, issue #443).
Test + comment only — no production behavior change.

## R1 — integration-level regression pin

`tests/unit/homing-snapshot-integration.test.ts` drives the **real**
`window-cache` store and mocks only the win32 boundary
(`enumWindowsInZOrder` / `restoreAndFocusWindow` / `getWindowRectByHwnd`).

This reproduces the *original* defect's interaction, which the fully-mocked
`homing-snapshot-delta.test.ts` cannot: Tier 2's focus path calls the real
`updateWindowCache()` with the post-move enumeration, overwriting the main
cache's screenshot-time region — the exact overwrite that used to collapse the
delta to `(0,0)`. The test asserts the click still lands at the
screenshot→live delta because the snapshot cache survives the overwrite. A
revert that recomputes the delta from the main cache fails this test
(mutation-verified locally).

## R2 — document the snapshot TTL split

Comment at `SNAPSHOT_TTL_MS` explaining why it is intentionally longer than
`CACHE_TTL_MS` (90s vs 60s): the two bound different things — screenshot-position
age vs HWND-recycle safety on the live lookup. No code change.

## Tests
All homing / window-cache / mouse unit suites green (43 tests). `build` clean.

### DIFF
--- a/src/engine/window-cache.ts
+++ b/src/engine/window-cache.ts
@@ -46,6 +46,13 @@ const CACHE_TTL_MS = 60_000;
  * tools have overwritten the main cache between screenshot and click.
  */
 const snapshotCache = new Map<string, { region: { x: number; y: number; width: number; height: number }; timestamp: number }>();
+// Intentionally LONGER than CACHE_TTL_MS (90s vs 60s). The two TTLs bound
+// different things: CACHE_TTL_MS bounds HWND-recycle safety on the *live*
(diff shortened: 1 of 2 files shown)


Harusame64:
@codex review

PR #449, opened 2026-06-10
fix: add GITHUB_TOKEN auth for GitHub API calls and handle PENDING SHA256

licat2023:
## Summary

Fixes two bugs that prevent `npx -y @harusame64/desktop-touch-mcp` from working reliably:

### Bug 1: GitHub API 403 Rate Limit (no auth headers)

`fetchReleaseByTag()` and `downloadFile()` call the GitHub Releases API without an `Authorization` header. Unauthenticated requests hit the 60/hr IP-based rate limit, which is easily exhausted on shared IPs (CI, VPNs, corporate networks).

**Fix**: Added `getGitHubToken()` (reads `GITHUB_TOKEN` or `GH_TOKEN` env vars) and `getGitHubHeaders()` helpers. Both `fetchReleaseByTag()` and `downloadFile()` now send `Authorization: Bearer <token>`, raising the rate limit from 60/hr to 5,000/hr.

### Bug 2: Invalid release SHA256 manifest ("PENDING" placeholder)

`RELEASE_MANIFEST.sha256` is set to `"PENDING"` — a placeholder meant to be filled by `scripts/update-sha.mjs` during CI publish. When the npm package is published without running that script, `expectedReleaseSpec()` rejects it with `Invalid release SHA256 manifest`, making the launcher crash even after a successful download.

**Fix**: `expectedReleaseSpec()` now detects `"PENDING"` and returns `{ sha256: null, sha256Pending: true }`. Downstream consumers (`isInstalled()`, `readCurrentRelease()`, `installRelease()`) handle `sha256: null` by skipping verification and emitting a warning.

### Testing

Verified end-to-end with `DESKTOP_TOUCH_MCP_HOME` pointing to a clean temp directory — the fixed launcher successfully authenticates against GitHub, downloads, extracts, and starts the MCP server without errors.

### DIFF
--- a/bin/launcher.js
+++ b/bin/launcher.js
@@ -38,9 +38,37 @@ function log(message) {
   console.error(`[desktop-touch-mcp] ${message}`);
 }
 
+function warn(message) {
+  console.error(`[desktop-touch-mcp] WARNING: ${message}`);
+}
+
 function fail(message) {
   console.error(`[desktop-touch-mcp] ${message}`);
   process.exit(1);
+}
+
+/**
+ * Reads the GitHub token from the environment.
+ * Supports both GITHUB_TOKEN (GitHub Actions standard) and GH_TOKEN (gh CLI).
+ */
+function getGitHubToken() {
+  return process.env.GITHUB_TOKEN || process.env.GH_TOKEN || null;
+}
+
+/**
+ * Returns headers for GitHub API and release download requests.
+ * Includes Authorization when a token is available to avoid rate limits.
+ */
+function getGitHubHeaders(extra = {}) {
+  const headers = {
+    "User-Agent": "desktop-touch-mcp-launcher",
+    ...extra,
+  };
+  const token = getGitHubToken();
+  if (token) {
+    headers["Authorization"] = `Bearer ${token}`;
+  }
+  return headers;
 }
 
 export function isDisconnectError(error) {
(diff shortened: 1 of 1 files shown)


licat2023:
**PR #449: CodeQL CI failure — not related to this PR's changes**

The CodeQL job failing on Initialize CodeQL with HttpError: Requires authentication is a **GitHub Actions limitation for fork PRs**, not caused by the changes in this PR.

**Why it happens**: For PRs from forks, GitHub downgrades GITHUB_TOKEN to near-anonymous access regardless of the permissions block. codeql-action/init@v3 requires authenticated API access to check feature enablement, hence the 401.

**Quick fix** — skip CodeQL on fork PRs (the push to upstream main still runs it):

`yaml
jobs:
  analyze:
    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
`

This PR only modifies in/launcher.js and doesn't touch the CodeQL workflow, so the failure is safe to ignore. But the same thing will happen on any external contribution — the if guard above fixes it for good.

Harusame64:
Thank you so much for this PR! I truly appreciate the time and effort you've put into tackling this. The GitHub API authentication part is fantastic and addresses a very real launcher reliability issue.

I need to think through the PENDING SHA256 handling a bit more before merging. In this project, PENDING is not just a temporary placeholder: it is part of the release procedure for the lightweight npm launcher + GitHub Release zip distribution flow, and that flow is also tied to the later MCP Registry (registry.modelcontextprotocol.io) publication step.

The source tree intentionally keeps RELEASE_MANIFEST.sha256 as PENDING until the release workflow builds the zip, computes the actual SHA256, and publishes the npm package with the finalized launcher manifest. That guard prevents accidentally publishing a launcher that downloads an unverified runtime zip.

So I want to preserve that release-safety property while still fixing the GitHub API rate-limit / authentication problem.

I will review this properly tomorrow and follow up with the concrete path forward. Again, huge thanks for your awesome contribution!

Harusame64:
Thank you again @licat2023 — merging this now! 🎉

The GitHub API authentication work is excellent and fixes a real reliability problem for users behind shared IPs. Merging your PR as-is so the commit history credits you properly.

Two follow-ups will land right after (no action needed on your side):

1. **PENDING SHA256 handling**: as mentioned above, the hard fail on `PENDING` is an intentional release-safety guard in our npm-launcher + GitHub-Release-zip flow. A follow-up PR will restore fail-closed as the default while keeping your use case (running the launcher straight from source) available via an explicit opt-in environment variable. Your diagnosis of the failure mode was spot on — we just want skipping integrity verification to be a conscious choice rather than a silent default.
2. **CodeQL failure**: you are absolutely right that it is the fork-PR token downgrade and unrelated to this diff. We will adopt the `if` guard you suggested in a separate PR, credited to you.

Really appreciate the thorough investigation and the end-to-end testing!

Harusame64:
🎉 Shipped in **v1.10.3** (released 2026-06-12) — thank you so much, @licat2023!

Your change is now live for everyone across all three channels (npm, the GitHub Release zip, and the MCP Registry). The first-run launcher now authenticates its GitHub Releases API request with `GITHUB_TOKEN` / `GH_TOKEN`, lifting the anonymous 60 req/hr limit to 5,000 req/hr, so installs no longer fail on shared / CI networks.

As a small follow-up we restored the fail-closed behavior for the `PENDING` SHA256 case (an unfinalized launcher refuses to run an unverified runtime) and added a `DESKTOP_TOUCH_MCP_ALLOW_UNVERIFIED=1` opt-in for running straight from a source checkout.

You're credited in the CHANGELOG's Thanks section. Really appreciate the contribution! 🙏

PR #474, opened 2026-06-27
ADR-026 Phase 1: screenshot disk-cache + by-ref output (walking skeleton)

Harusame64:
## ADR-026 Phase 1 — screenshot disk-cache + by-ref output (walking skeleton)

Reduces screenshot token cost by persisting capture bytes to a per-user disk-cache and returning a cheap `resource_link` by default instead of inline base64. Inline pixels are returned only when genuinely needed.

**Plan:** `desktop-touch-mcp-internal@c69c8ac:docs/adr-026-screenshot-token-reduction-disk-ref.md`

### What this PR does (Phase 1 = trunk / contract lock)
- **`getRuntimeDir()`** (`src/utils/runtime-dir.ts`) — pure env→path resolver for `%USERPROFILE%\.desktop-touch-mcp` (`DESKTOP_TOUCH_MCP_HOME` override).
- **Disk-cache + secure resolver** (`src/engine/screenshot-cache.ts`) — `persistCapture()` writes bytes + appends an NDJSON index; `readCaptureBytes()` resolves an **opaque captureId** to a file *inside the canonical cache root only*. Security (ADR-026 §4): basename-only index entries, **symlink rejected before realpath**, separator-aware containment via `path.relative` (not `startsWith`), **no-follow open + dev/ino identity gate** so a lstat→open symlink swap is caught, bytes read straight from the gated fd (never re-opened by path).
- **`screenshot` detail=image → by-ref** (`src/tools/screenshot.ts`) — `buildImageResponse()` persists and returns a `resource_link` by default; inline image only on `confirmImage` / `dotByDot` (coordinate use). Dimension/coordinate/hints text is always preserved. The guard now blocks only `detail='som'` (migrated to ref in Phase 2). All three image targets (window / display / fullscreen) routed through the helper. **R6 degrade:** a cache-write failure falls back to inline + warning, never an error.
- **`screenshot://by-ref/{captureId}` resource** (`src/tools/screenshot-resources.ts`, registered always-on) — read handler streams the validated bytes as a blob; a dangling/out-of-cache ref surfaces an explicit error (AC3 / R7), never a silent empty.
- `_types.ts` — `resource_link` added to `ContentBlock`.

### Read-policy (the North Star, ADR-026 §2.2)
Reading a resource re-expands base64, so the saving only materialises when reads are **rare/deferred**. Confirmed via research that Claude Code does **not** auto-read `resource_link` (it surfaces a deferred metadata block). The tool + resource descriptions steer the agent to open a ref only when pixel inspection is genuinely required.

### Tests
- 18 new unit tests green: `runtime-dir` (4) + `screenshot-cache` (9, incl. AC3 path-traversal: not_found / basename / absolute / nested-sep / symlink) + `screenshot-response` (5, incl. default-ref / confirmImage-inline / round-trip / R6 degrade).
- Full unit+e2e suite: **4298 passed**. The single failure (`browser-cdp-verification` headless rescue) is a pre-existing flake that **passes in isolation** and is unrelated to this diff (no browser file touched).

### Out of scope (later phases)
- Phase 2: remaining emitters (som / detail=text SoM fallback / diffMode / mode='background' / scroll-capture / workspace_snapshot / roiCapture).
- Phase 3: `screenshot_query` + `screenshot_gc` (29→31) + GC policy.
- Phase 4: catalogue sync + CHANGELOG + release.

### Review
Security-critical (path-traversal resolver) — requesting **Opus** (architecture / spec alignment) + **Codex** (API contract / security) per CLAUDE.md §3.3.


Harusame64:
@codex review

Security-critical PR (ADR-026 Phase 1): opaque-captureId → disk path resolver. Please focus on the path-traversal surface in `src/engine/screenshot-cache.ts` (symlink-before-realpath ordering, `path.relative` containment vs `startsWith`, lstat→open TOCTOU dev/ino identity gate, no-follow open on Windows), the resource read handler in `src/tools/screenshot-resources.ts` (dangling-ref error vs silent empty), and the output-contract change in `src/tools/screenshot.ts` (default ref / confirmImage inline, guard now blocks only detail=som).

Harusame64:
@codex review

Round 2 — applied all three findings from your last pass:
- **P1** (resource completion): removed the `captureId` completion callback entirely; refs are no longer enumerable, the exact ref must come from a tool response.
- **P2** (dangling ref): `asNotFound()` now maps ENOENT from lstat/realpath/open to `CaptureRefError("not_found")`, so a GC'd file surfaces the opaque error (no abs-path leak). Regression test added.
- **P2** (docs): `detail`/`confirmImage` schema descriptions + tool caveats now describe the by-ref default.
Plus: `persistCapture` writes with `flag:"wx"` (exclusive create), and the two flagged test temp writes use `fs.mkdtempSync`.

Harusame64:
@codex review

Final pass on commit `7510ea94`. Your last P2 (stale stub catalog) is already resolved: `src/stub-tool-catalog.ts` was regenerated with the by-ref descriptions in 7510ea94. CI is fully green (build / docker-build / CodeQL all pass; the js/insecure-temporary-file finding on the validated read-only open was dismissed as a reviewed false positive). Please confirm clean.

Harusame64:
@codex review

Applied your macro P2 on commit `70644041`: `runInnerToolAsResult` now carries `resource_link` blocks through `InnerToolOutcome.links` → `StepResult._links` → final macro content (stripped from the summary JSON like `_images`). A `run_macro` screenshot step with the by-ref default now propagates its `screenshot://by-ref/{id}` URI to the caller. Adapter test added. Please confirm clean.

Harusame64:
@codex review

Applied your native-test P2 on commit `3a8a21ac`: `buildImageResponse` moved to `src/tools/screenshot-response.ts`, which depends only on the pure cache layer (`screenshot-cache`) + tool-result types — no native bindings. The test now imports the native-free module, so it runs on a host without libnut. Behaviour unchanged. Please confirm clean.

PR #477, opened 2026-06-27
feat(desktop_act): ADR-026 Phase 2c — roiCapture crop by-ref (additive)

Harusame64:
## What

ADR-026 **Phase 2c** — the **last** image emitter. `desktop_act`'s `roiCapture` embedded the SoM crop as **base64 inside the JSON envelope** (the worst inline-base64 offender, on the highest-frequency tool). Move it by-ref, **additively**, with no v2 envelope break.

### Changes
- **`RoiCapture`** (`guarded-touch.ts`): `somImage` widens `string → string | null` (**null by default** — meaning unchanged, nullability only, additive-safe) + new `somImageRef?` (the `screenshot://by-ref/` URI) + `somImageWarning?` (R6 only). No internal reader of `roiCapture.somImage` exists (grep 0), so null-by-default is internally safe.
- **`assembleRoiCaptureFromSom`** (the single fold + legacy assembly point) persists the crop and returns `somImage:null` + `somImageRef`. **R6**: a disk-cache write failure keeps `somImage:null` + a warning — never inline base64, never throws (the act already succeeded; `roi`/`entities`/`source` stay intact).
- **`desktopActRawHandler`** attaches the crop as a `resource_link` content block when a `somImageRef` exists; the L5 commit wrapper preserves `content[1+]` (verified `_envelope.ts:3122` `...result.content.slice(1)`).
- Default **never** inlines base64 in the act envelope (OQ-2 default).
- Synced the two public descriptions (`returnCapture` param + recovery-path blob) + internal comments. No stub-catalog mirror (desktop_act is dynamic v2).

### Tests
- `desktop-act-frame-diff-dispatch`: by-ref case (`somImage` null / `somImageRef` set / `resource_link` block) + R6 case (`somImage` null / warning / no ref / act still ok), cache pointed at a temp dir.
- e2e `desktop-act-roi-capture`: resolves the ref via `readCaptureBytes` (PNG magic bytes).
- bench measures `somImageRef` bytes. 103 passed across the desktop-act/roi/macro regression set; build + lint clean.

## Review

Opus + Codex (production change; **most contract-sensitive** — v2 public envelope). This is the final emitter; Phase 2 closes after this. Branched from main after 2b merged (plan §4).

Plan: `desktop-touch-mcp-internal@6b9b096:docs/adr-026-phase2-emitter-fanout-plan.md` §3.6.

### DIFF
--- a/benches/adr-024-seed2-roundtrip.mjs
+++ b/benches/adr-024-seed2-roundtrip.mjs
@@ -7,8 +7,8 @@
 //   desktop_act -> desktop_state -> screenshot   (3 calls; 2 post-act confirms)
 // and S5c makes it
 //   desktop_act(returnCapture:"on-change")        (1 call; 0 post-act confirms)
-// because the act response carries roiCapture { somImage (≈ the screenshot crop),
-// entities (≈ the discover/state preview) }.
+// because the act response carries roiCapture { somImageRef (≈ a by-ref link to
+// the screenshot crop, ADR-026), entities (≈ the discover/state preview) }.
 //
 // This bench QUANTIFIES that reduction live. It is **headed-gated** (needs a real
 // GUI + the full native engine: DXGI / PrintWindow / OCR) and therefore is NOT a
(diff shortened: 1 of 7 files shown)


Harusame64:
@codex review

Harusame64:
Round 1 P1 fix pushed (commit 7c6542ef, doc-only): synced the canonical roiCapture shape table in `docs/system-overview.md` to the by-ref shape (`{ roi, somImageRef, entities, source }`, somImage null by default). No code change. (Opus R1 P3-2 fold-path orphan-persist recorded in the private ADR §8 R11 for Phase 3 gc.)

@codex review

PR #478, opened 2026-06-27
feat(screenshot-cache): ADR-026 Phase 3a — query / gc / retention engine core

Harusame64:
ADR-026 **Phase 3a** (engine core, security-critical). Phases 1/2 routed every emitter through the by-ref model, so the cache grows unbounded. This adds the engine layer for observing (query), reclaiming (gc), self-bounding (auto-prune), and orphan recovery as a **pure addition**. The public tools (29→31) are **PR-3b (#479)**.

## Storage: per-capture sidecars (no shared index — multi-process safe by design)

Each capture is **two independent files**, written/removed atomically: the image `{captureId}.{img}` (pixels) + `{captureId}.json` (metadata sidecar). **There is no shared index file.** Liveness = both files exist; the directory IS the index and `readIndex` folds the sidecars.

Because no two processes ever write the same file and nothing is rewritten in place, **concurrent access from multiple desktop-touch processes on one PC is correct without any lock** — the read-modify-rewrite race that a shared index would need a lock to guard simply cannot exist. (Earlier iterations of this PR tried an index lock and a tombstone log; both still raced on a rewrite/compaction path, which a local `codex review` confirmed. The sidecar model removes the shared mutable state entirely.)

- **persist**: write the image (`wx`), then publish the sidecar atomically (temp→rename).
- **readIndex**: readdir sidecars; a capture is live only if both its sidecar and image exist.
- **deleteCapture / gcCache**: securely unlink the image (full validation gauntlet) + remove the sidecar — no shared rewrite, nothing to lose.
- **orphan sweep**: reclaim our own crash residue (image without sidecar, sidecar without image, stale `*.tmp`) after a fixed grace.

## Engine API (`src/engine/screenshot-cache.ts`)
- `queryCaptures(filter, env)` — read-only listing (tag case-folded; opaque captureId + uri only, never a path; whole-cache stats).
- `deleteCapture(captureId, env)` — secure delete of image + sidecar; dangling → `{deleted:false}`.
- `gcCache({dryRun,policy,includeOrphans,now,protectCaptureId}, env)` — retention union (age/count/bytes) + tag scope + orphan sweep + keep-newest invariant.
- auto-prune from `persistCapture` (throttled, env-tunable caps) + R9 opaque error coercion.

## Security
- **Strict ownership gate** `CAPTURE_ID_RE` (`base36-16hex`) at every boundary — `assertSafeCaptureId` (read/delete entry points), the `readIndex` sidecar listing, and the orphan sweep — so a shared/overridden `DESKTOP_TOUCH_SCREENSHOTS_DIR` never lists or deletes a foreign file (`settings.json`, an unrelated `*.tmp`, a non-id image).
- Image-file validation gauntlet preserved: basename assert → lstat-before-realpath symlink reject → realpath → `isWithinRoot` (never `startsWith`) → no-follow open → dev/ino identity gate (defeats lstat→open/unlink swap).

## Multi-LLM eviction floor
The sidecar model removes the cross-process race, but the retention budget (maxCount/maxBytes/auto-prune) is per-cache-dir global, so a flood of captures from one process could auto-prune a ref another process was just handed. An **eviction floor** (`DESKTOP_TOUCH_SCREENSHOT_MIN_EVICT_AGE_MS`, default 60s) never evicts a capture younger than the floor, so a just-handed ref survives long enough to be read regardless of another process's load.

## Tests / review
Engine + handler + emitter unit tests rewritten to the sidecar model with strict captureIds (atomic publish / liveness gate / foreign-file protection / disjoint delete+persist / eviction floor); build + lint clean. Local `codex review`: **P1 = 0 (race-free confirmed)**; Opus architecture review: **GO**. The new public tools (29→31), all public-surface count strings, env-var docs, and the CHANGELOG are **PR-3b / Phase 4**.

Plan: `desktop-touch-mcp-internal plan/adr-026-phase3-query-gc:docs/adr-026-phase3-query-gc-plan.md` (OQ8 → sidecar).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,8 @@ tools/win-ocr/obj/
 .vitest-out*.txt
 .vitest-out*.json
 .vitest-verbose.txt
+# vitest run sentinel (machine-local PID, e2e:stop mechanism)
+.vitest-running
 
 # cargo check / build output (captured for diagnostics)
 .cargo-check-out.txt
--- a/src/engine/screenshot-cache.ts
+++ b/src/engine/screenshot-cache.ts
@@ -656,7 +656,14 @@ function removeIndexEntries(root: string, ids: Set<string>, _env: NodeJS.Process
   withIndexLock(root, () => {
     const all = [...readIndex(root).values()];
     if (!all.some((e) => ids.has(e.captureId))) return;
-    writeIndexAtomic(root, all.filter((e) => !ids.has(e.captureId)));
+    try {
+      writeIndexAtomic(root, all.filter((e) => !ids.has(e.captureId)));
+    } catch {
+      // Best-effort (plan §3.3): a failed rewrite leaves stale entries → read
+      // surfaces not_found (AC3) and the next gc self-heals. Never propagate —
+      // deleteCapture must not throw after a successful unlink, and gcCache must
+      // still return its reclaimed-bytes summary.
+    }
   });
 }
 
--- a/src/engine/screenshot-cache.ts
+++ b/src/engine/screenshot-cache.ts
@@ -141,15 +141,20 @@ function withIndexLock<T>(root: string, fn: () => T): T {
       break;
     } catch (e) {
       if ((e as NodeJS.ErrnoException).code !== "EEXIST") break; // unlockable → best-effort
+      // Lock is held — steal it ONLY if it is stale AND we actually removed it.
+      let cleared = false;
       try {
         const st = fs.statSync(lockPath);
         if (Date.now() - st.mtimeMs > LOCK_STALE_MS) {
-          try { fs.unlinkSync(lockPath); } catch { /* raced another stealer */ }
-          continue;
+          try { fs.unlinkSync(lockPath); cleared = true; } catch { /* can't remove → fall through */ }
         }
       } catch {
-        continue; // lock vanished between open and stat → retry immediately
+        cleared = true; // lock vanished between open and stat → retry acquire immediately
       }
+      if (cleared) continue; // lock removed/gone → retry openSync now
+      // Held-and-fresh OR stale-but-unremovable (e.g. a directory named _index.lock):
+      // wait, then hit the backstop — never `continue` past it, or a wedged lock
+      // would spin forever instead of proceeding best-effort (Codex P2).
       if (Date.now() - start > LOCK_MAX_WAIT_MS) break; // give up → proceed best-effort
       sleepSync(LOCK_RETRY_MS);
     }
--- a/src/engine/screenshot-cache.ts
+++ b/src/engine/screenshot-cache.ts
@@ -110,6 +110,16 @@ export function persistCapture(
   };
   fs.appendFileSync(indexPath(root), JSON.stringify(entry) + "\n");
 
+  // Bound cache growth (R2) without the agent ever calling screenshot_gc. Throttled
+  // + best-effort + protects this very captureId — a prune failure must NEVER fail a
+  // capture, and auto-prune must never delete the ref this call is about to return
+  // (ADR-026 Phase 3 §3.4, keep-newest invariant).
+  try {
+    maybeAutoPrune(captureId, env);
+  } catch {
+    /* never fail a capture because a background prune errored */
+  }
+
   return {
     captureId,
     uri: REF_URI_PREFIX + captureId,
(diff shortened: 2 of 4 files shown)


Harusame64:
@codex review

Security-critical disk-cache. Please focus on the delete / traversal / TOCTOU axis: (1) the re-validation right before `deleteCapture` unlinks, (2) orphan-sweep basename validation, (3) the `gcCache` keep-newest invariant and atomic index rewrite, (4) auto-prune never deleting the just-returned ref, (5) the R9 opaque coercion not leaking a path.

Harusame64:
@codex review

Round 1 findings applied (commit af96c0e9): P1 — `gcCache` index rewrite re-reads + merges before the rename; P2-1 — `validateCacheFileForDelete` re-lstats the realpath-resolved target and returns ITS identity; P2-2 — `deleteCapture` coerces a non-ENOENT unlink failure to an opaque ref error (no path leak).

Harusame64:
@codex review

Round 2 findings applied (commit 9ee233b5): P1 — all index mutations (persist append / delete / gc rewrite) are serialized by an exclusive `_index.lock`, with the fresh read taken inside the lock so no concurrent append is lost; P2 — `deleteCapture` now also removes the index entry (real delete and dangling).

Harusame64:
@codex review

Round 3 P2 applied: the lock give-up (`LOCK_MAX_WAIT_MS`) was raised from 2s to 15s, longer than the 10s stale window, so a legitimately held lock is always resolved first (acquire on release, or stale-steal) and the unlocked backstop never fires as a premature give-up.

Harusame64:
@codex review

Round 3 (commit 035a4c72): the orphan delete now carries a dev/ino identity gate (parity with the index-backed delete). The lock stale-steal residual is deferred as a documented limitation (ADR OQ8): every locked op is a synchronous sub-ms write, so a >10s hold means a frozen/dead process, and any loss self-heals via the orphan sweep.

Harusame64:
@codex review

CodeQL flagged `writeIndexAtomic` temp write as js/insecure-temporary-file (high); added the exclusive-create flag (commit 16347986), mirroring `persistCapture`.

Harusame64:
@codex review

P2 (wedged-lock infinite loop) fixed: only `continue` after a successful stale-unlink (a `cleared` flag); a stale-but-unremovable lock now falls through to the timeout backstop + sleep instead of spinning forever. Added a stale-steal-success test.

Harusame64:
@codex review

Two items (commit e04fe0fc): (1) CodeQL js/insecure-temporary-file on the lock `openSync(path, "wx")` — rewrote the lock to use `writeFileSync` with the wx flag (CodeQL recognizes it; functionally equivalent, lock = file existence released by unlink); (2) P2 — the orphan grace is now a fixed short append-race window, no longer scaled by `maxAgeMs` (unreadable orphans should not survive a multi-day retention policy).

Harusame64:
@codex review

Harusame64:
### Re: the latest automated review (commit `9a6d0af6`)

Thanks — triaged every suggestion against the **current** code. The bot re-pins findings from earlier rounds onto the new HEAD, and this PR's storage layer was rewritten from an index-lock to a **per-capture sidecar** model, so most of them point at code that no longer exists. Mapping each to the current tree:

| Suggestion | Pinned commit | Status on current HEAD |
|---|---|---|
| Migrate legacy `_index.ndjson` before sidecar lookup (`:319`) | `9a6d0af6` (current) | **Moot.** The disk cache + by-ref output is itself unreleased (it lands in this train, `v1.10.4..main`). No shipped release ever wrote `_index.ndjson`, so an upgrade starts from an empty cache — there is nothing to migrate. A dev box that ran an intermediate build just has its old images reclaimed by the orphan sweep; a stray `_index.ndjson` doesn't match `CAPTURE_ID_RE` and is left untouched. |
| Coerce unlink failures to opaque ref errors (`:500`) | `5f957339` | **Done.** `unlinkValidatedCapture` coerces non-ENOENT unlink failures via `asRefError(...)` (no raw path leak, R9). |
| Remove deleted captures from the index (`:611`) | `9a6d0af6` | **Obsolete.** `deleteCapture` unlinks the `{id}.json` sidecar, so `queryCaptures` (which folds sidecars) no longer lists it. No shared index to repair. |
| Don't let `maxAgeMs` suppress orphan cleanup (`:758`) | `bc7d1e23` | **Done.** The orphan grace is a fixed `ORPHAN_GRACE_MS_DEFAULT` (5 min); it is not scaled by `maxAgeMs`. |
| Restrict orphan GC to cache-owned filenames (`:784`) | `e04fe0fc` | **Done.** The orphan sweep gates every branch (image / sidecar / `*.tmp`) on `CAPTURE_ID_RE` / `SIDECAR_TMP_RE`, so a shared dir's foreign files are never reclaimed. |
| Pin orphan identity before unlinking (`:788`) | `9fa41fa7` | **Done.** The sweep carries the scanned `dev/ino` and requires the re-resolved target to match that inode before unlinking (mirrors the index-backed delete gate). |
| All `_index.lock` stale-steal / wedged-lock / append-loss suggestions | `e04fe0fc` / `16347986` / … | **Moot by design.** The lock and the shared index are removed entirely. Two processes never write the same file and nothing is rewritten in place, so the append-loss race these guarded cannot occur — confirmed P1=0 by local `codex review` on the sidecar model.

Net: no open P1/P2 on the current design. Happy to take a fresh `@codex review` pass now that everything is pinned to `9a6d0af6`.

PR #485, opened 2026-06-28
feat(wgc): ADR-027 Phase 4 — WGC latency bench + CHANGELOG (no release)

Harusame64:
## ADR-027 Phase 4 (final) — WGC latency bench + CHANGELOG

The last ADR-027 phase. **No release** — the 3-channel publish is deferred and bundled with ADR-028; the `[1.11.0]` CHANGELOG entry accumulates and ships later.

### What's here
- **`benches/wgc_latency.mjs`** (+ `npm run bench:wgc-latency`): head-to-head WGC vs PrintWindow (flag 2 = PW_RENDERFULLCONTENT) capture latency on the same live window. Reports cold (1st) vs warm (device reused), mean/median, and the WGC/PrintWindow ratio. **Non-hermetic** (needs a live DWM-composited window — the D3 gate rejects minimized/hidden/cloaked), so it is a local dogfood tool, **not** a CI step.
- **`CHANGELOG.md`** `[1.11.0] - Unreleased`: two user-facing bullets — GPU-rendered / occluded windows now capture real pixels via Windows.Graphics.Capture instead of black, and an explicit `captureBlocked` signal when content genuinely can't be captured (with the hedged "could be a legitimately black window OR uncapturable" framing). Headline extended.
- **`package.json`**: `bench:wgc-latency` script.

### Bench results (this host, GPU window, 20 iterations)
| backend | latency |
|---|---|
| WGC cold (1st) | 177.3 ms (one-time D3D device build) |
| WGC warm (device reused) | mean 79.1 / **median 78.8 ms** |
| PrintWindow (flag 2) | mean 8.4 / median 8.3 ms |

- **AC2 (device reuse):** cold/warm = **2.25× faster** after the first capture — the worker thread holds the D3D device across captures.
- **R2 / D3 (quantitative backing):** WGC warm is **~9.5× slower** than PrintWindow. This is why WGC is *not* promoted to a blanket hot-path primary — it stays the `mode='background'` primary + the normal-mode black-rescue rung, where its unique value (real pixels for GPU/occluded windows PrintWindow blacks out) outweighs the cost.

### CI compile guard
No new CI code needed. `src/win32/wgc.rs` (which uses the `Graphics_Capture` windows-rs feature) is compiled by the existing `check:rs-workspace` (`cargo check`) and `check:rs-test-compile` (`cargo test --no-run`, incl. wgc.rs's `#[cfg(test)]` tests). Dropping the feature would fail those checks. Live WGC capture is non-hermetic, so compile-only is the correct guard.

No `src/` change — the capture ladder, AC8/AC9, and all tool behavior are untouched.

Plan / acceptance: `desktop-touch-mcp-internal:docs/adr-027-wgc-capture-layer.md` §6.4 (Phase 4), AC2.

Reviewed locally: Codex (clean) + Opus phase-boundary review (GO; one P3 CHANGELOG wording nit fixed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,31 @@
 # Changelog
 
-## [1.11.0] - Unreleased — Screenshots now cost a fraction of the tokens, plus tools to manage the cache
+## [1.11.0] - Unreleased — Screenshots cost a fraction of the tokens, capture real pixels from GPU-rendered & occluded windows, plus tools to manage the cache
 
 ### Added
 
+- **GPU-rendered and occluded windows now capture real pixels instead of black.**
+  Hardware-accelerated windows — Chrome / Edge, Electron apps (VS Code, Slack,
+  Discord), and many WinUI / game windows — often came back as a black rectangle
+  because the classic capture path can't read their GPU-composited surface, and
+  the on-screen fallback can only grab a window that isn't covered by another.
+  `screenshot` now adds a Windows.Graphics.Capture path that reads the same
+  composited surface the desktop shows, so those windows (including ones sitting
+  behind others) capture their actual contents. It's used automatically:
+  `mode='background'` captures eligible visible windows through it, and a normal
+  `screenshot` that comes back black is retried through it before any fallback —
+  no new options to set. Minimized, hidden, and off-desktop windows keep using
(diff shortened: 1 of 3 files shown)


Harusame64:
@codex review

PR #492, opened 2026-07-01
ADR-014 v2 R3 Key Locker — L0 trunk (contract lock)

Harusame64:
## ADR-014 v2 R3 Key Locker — L0 trunk (contract lock)

L0 stands up the Key Locker trust boundary: a compiled C# component that opens a WPF
PasswordBox secure dialog on request, encrypts/decrypts secrets with Windows DPAPI at rest,
serves them over a named-pipe server reusing S1's auth, and registers its window for engine
tool-exclusion. Retrieve/inject (askpass / SendInput) is L2; capture-on-use + landed-detection
is L3; the MCP tool surface is L4.

### What's in this PR

**C# locker** (`tools/key-locker/`, landed in the first commit)
- `NamedPipeServerStream` with `PipeOptions.CurrentUserOnly` (cross-user block) +
  `FILE_FLAG_FIRST_PIPE_INSTANCE` fail-loud + kernel client-verify
  (`GetNamedPipeClientProcessId == -McpPid`) — the S1 bridge-host auth pattern.
- DPAPI (CurrentUser) at-rest store — the value envelope only (opaque id → wrapped blob);
  the binding-key schema is L1.
- WPF PasswordBox secure capture dialog — D1-spike-proven un-capturable (no UIA value, masked,
  no clipboard).
- Identified-publisher metadata + least-privilege `app.manifest` (asInvoker, no elevation) so
  the shipped exe presents as a legitimate credential helper to Windows reputation checks.
- Headless `-SelfTest` seam: DPAPI round-trip + wrong-value + corrupt-tag rejection, no pipe/GUI.

**Node client** (`src/engine/key-locker-host.ts`)
- `KeyLockerHost` mirrors S1's `bridge-host.ts` handshake/framing (the proven C#↔C# direction).
- Control verbs: ping / version / capture / exists / delete / shutdown.
- **The secret never crosses the pipe** — `capture()` returns only `{captured, rt}`; the wire
  carries the opaque key + control, never the secret.

**Engine tool-exclusion** (`src/engine/tool-exclusion.ts` + `win32.ts` + `_resolve-window.ts`)
- A PID-keyed registry; the locker registers itself on spawn (unregisters on dispose / failure).
- `enumWindowsInZOrder()` drops any window owned by an excluded PID (feeds
  screenshot/perception/discover/dialog resolution). Gated on a non-empty set → **zero extra
  syscalls when no locker is alive**.
- `resolveWindowTarget()` Cases 1/2 (explicit `hwnd`, `@active`) bypass the enumerator, so they
  refuse an excluded target (`WindowExcluded`, L0-local; wired into `_errors.ts` at L4).

### Threat model (verbatim)

The guarantee is **BOUNDED to the honest-Claude automation path** — Claude never sees the secret
and never types passwords — **not a structural guarantee** against an adversarial AI or a hostile
same-user process. Tool-exclusion protects the honest tool path; it is not a defense against a
process calling Win32 directly.

### Tests

- `tests/unit/tool-exclusion.test.ts` — registry + REAL `enumWindowsInZOrder()` filter.
- `tests/unit/key-locker-host.test.ts` — client wire protocol vs a Node fake peer, incl. an
  assertion that the secret never appears on the wire.
- `tests/unit/resolve-window.test.ts` — Cases 1/2 `WindowExcluded` refusal + zero-overhead gate.
- `tests/e2e/key-locker.e2e.test.ts` — live exe: DPAPI `-SelfTest` (round-trip + rejections +
  no plaintext at rest); pipe control plane; rogue-client rejection; `FIRST_PIPE` fail-loud (exit 3).

Build clean; unit 4079 passed (2 unrelated timing flakes pass in isolation); e2e key-locker 4/4;
lint 0 errors; stub-catalog (29 tools) + failwith fixtures unchanged.

Plan: `desktop-touch-mcp-internal@main:docs/adr-014-v2-r3-slice-plan.md` (L0)


Harusame64:
@codex review

Harusame64:
@codex review

Round 2: applied both Round 1 P1s (P1-A discover-by-hwnd → OCR, P1-B background-screenshot-by-title) in commit 2b4301e2. Please verify the fixes and check for any REMAINING by-identity enumeration/read path that still reaches the excluded window.

Harusame64:
@codex review

Round 3: the UIA-by-title P1 you flagged (and Opus independently found) is fixed in commit 30b148e5 — every uia-bridge public function taking a windowTitle now refuses an excluded title BEFORE any native/PS UIA read (via a new unfiltered isExcludedTitle predicate), plus a resolveWindowTarget front-door refusal, getActiveWindow guard, and recognizeWindowByHwnd guard. Please verify and check for any REMAINING by-identity path that still reaches an excluded PID's window.

Harusame64:
@codex review

Fixed your P2 (commit 373dea88): the runSomPipeline / recognizeWindowByHwnd / getActiveWindow exclusion guards now only apply to a non-null explicit hwnd — a title-only call (targetHwnd=null) is handled by the filtered enumerator and no longer false-refused while a locker is alive. Added a regression test. Also regenerated the failWith shape fixtures (the getActiveWindow failWith callsite drifted the ADR-021 gate → CI build fix).

Harusame64:
Re: the P2 on `Program.cs` ("Request first-pipe creation explicitly") — verified this is a false alarm; the fail-loud guarantee already holds.

The premise (that `new NamedPipeServerStream(name, InOut, 1, Byte, ...)` does not request `FILE_FLAG_FIRST_PIPE_INSTANCE`) is incorrect: the .NET runtime sets it when `maxNumberOfServerInstances == 1` (`NamedPipeServerStream.Windows.cs` computes `openMode |= (maxInstances == 1 ? FILE_FLAG_FIRST_PIPE_INSTANCE : 0)`).

Tested the exact race described — a same-user squatter pre-creates the pipe with `MaxAllowedServerInstances` (multi-instance, no first-pipe flag), then the real helper (`maxInstances=1`) runs on the same name. The helper's create is **refused** (ERROR_ACCESS_DENIED → exception → `return 3`), so `KeyLockerHost` observes the child die and aborts with `KeyLockerSpawnFailed` — it never connects to the squatter. Result: **locker exit code = 3**.

Encoded as a regression rather than argued in prose: added a "MULTI-INSTANCE squatter" e2e (`tests/e2e/key-locker.e2e.test.ts`) that pins exit 3, and expanded the `Program.cs` comment to cite the runtime behavior. Commit `4a89c852`.

(Server-side verification is infeasible for the Node client — it can't obtain the pipe's OS handle, so `GetNamedPipeServerProcessId` is unavailable; this was resolved in S1. And an adversarial same-user process guessing the 128-bit pipe name is out of the §8 bounded threat model regardless — the fail-loud is the honest-path guarantee.)

Harusame64:
@codex review

PR #495, opened 2026-07-02
ADR-014 v2 R3 Key Locker — L3 (partial): SessionTracker (§3)

Harusame64:
## Summary

First implementation increment of **L3 capture-on-use** (design converged in internal PR #30). This lands the **crux new surface** — the per-pane `SessionContext` provider. Pure engine-side TypeScript, no Win32/IO.

Plan: `desktop-touch-mcp-internal@<sha>:docs/adr-014-v2-r3-l3-capture-plan.md` §3.

## What's in

- **`src/engine/key-locker/session-tracker.ts`** — the `SessionTracker`. The terminal layer tracks none of `{execHost, isRemote, cwd}` (L1 declared it "L3 owns supplying"). This maintains it per-pane from the **dispatched-command stream** (authoritative, never screen-scraped):
  - `beginLocalSession` anchors a KNOWN-LOCAL pane; a never-anchored pane is **UNKNOWN → caller declines to derive** (never guesses localhost for a maybe-already-remote pane).
  - interactive `ssh user@host` **pushes a remote frame** (a one-shot `ssh host cmd` / query mode does NOT); nested ssh stacks; `noteSessionEnd` pops one frame (the manager calls it on the observed ssh child exit); `markUnknown` sinks to UNKNOWN on an **unconfirmable** session-end (closes the Opus-review stale-remote-frame wrong-target axis).
  - `cd` best-effort cwd (fails safe to `undefined`; remote cwd stays unknown → L1 declines cwd-dependent derivation rather than wrong-target).
  - `recordDispatch` ordering contract pinned (Opus R1 #495 P2-1): derive-then-record — a credential command's own binding is derived from its PRE-effect frame, the effect applies for subsequent commands.

## Scope note — the L2 `consolePid` doc-fix lives in #496, NOT here

Earlier revisions of this description claimed a doc-fix to L2's frozen `consolePid` comments (`key-locker-host.ts` + `Injection.cs`). That doc-fix is owned by **PR #496** (L3 §4 InjectTarget assembly), whose diff actually touches those files. This PR (#495) is **TypeScript-only** — `session-tracker.ts` + tests — and carries no C# / `key-locker-host.ts` change. Claim removed to keep the description fact-consistent with the diff (CLAUDE.md §3.1).

## Verification

- **25 unit tests** (`tests/unit/key-locker-session-tracker.test.ts`): anchoring, the UNKNOWN fail-safe, ssh in/out/nested, one-shot vs interactive, query modes, cwd tracking, `markUnknown`, conditional-reach segments. Green.
- `npm run build` clean; `npm run lint` 0 errors; L0 `-SelfTest` + L2 `-SelfTestL2` regressions green. (No C# in this PR ⇒ no `dotnet build` step.)

## Scope / what's next

This is **L3 §3 only**. Remaining L3: InjectTarget assembly (§4, PR #496), two-mode landed-detection (§2), the capture-on-use loop (§1), the `KeyLockerManager` + `consentGate` (§6); then L4 (the `key_locker` tool + first-run consent + the 31→32 catalogue cascade). No new MCP tool here ⇒ no catalogue delta.

## Not blocked-on but tracked

The L2 §8c live SendInput dogfood (real foreground conhost + real ssh-keygen/git) needs the user's desktop and gates R3 *release*, not L3/L4 implementation.

### DIFF
--- a/src/engine/key-locker-host.ts
+++ b/src/engine/key-locker-host.ts
@@ -102,11 +102,11 @@ export interface InjectTarget {
   /** The console window HWND (decimal string). */
   hwnd: string;
   /**
-   * The pid that OWNS the console window (`GetWindowThreadProcessId(hwnd)`), which for a modern
-   * pseudoconsole-hosted shell is the SHELL process (powershell/pwsh), not literally conhost — L3
-   * supplies `getWindowProcessId(hwnd)` and the locker re-verifies with the SAME API on the same
-   * hwnd, so the value matches by construction (L3 plan §4). The `ConsoleWindowClass` allowlist is
-   * what excludes a WT multiplexer; this pid is the window-owner, not a separate conhost pid.
+   * The pid that OWNS the console window — whatever `GetWindowThreadProcessId(hwnd)` returns for the
+   * `ConsoleWindowClass` window. L3 supplies `getWindowProcessId(hwnd)` and the locker re-verifies
+   * with the SAME API on the same hwnd, so the value matches by CONSTRUCTION (L3 plan §4). The
+   * `ConsoleWindowClass` allowlist is what excludes a WT multiplexer; this is the window-owning pid,
+   * not asserted to be a specific conhost-vs-shell process (Opus R1 P3-1).
    */
   consolePid: number;
   /** Opaque hash of the expected pane identity (secondary anchor). */
--- a/src/engine/key-locker-host.ts
+++ b/src/engine/key-locker-host.ts
@@ -101,7 +101,13 @@ export interface LockerReply {
 export interface InjectTarget {
   /** The console window HWND (decimal string). */
   hwnd: string;
-  /** The console-HOST (conhost.exe) pid that owns the window — NOT the child pid (§2.2). */
+  /**
+   * The pid that OWNS the console window (`GetWindowThreadProcessId(hwnd)`), which for a modern
+   * pseudoconsole-hosted shell is the SHELL process (powershell/pwsh), not literally conhost — L3
+   * supplies `getWindowProcessId(hwnd)` and the locker re-verifies with the SAME API on the same
+   * hwnd, so the value matches by construction (L3 plan §4). The `ConsoleWindowClass` allowlist is
+   * what excludes a WT multiplexer; this pid is the window-owner, not a separate conhost pid.
+   */
   consolePid: number;
   /** Opaque hash of the expected pane identity (secondary anchor). */
   titleFp: string;
--- a/src/engine/key-locker-host.ts
+++ b/src/engine/key-locker-host.ts
@@ -101,13 +101,7 @@ export interface LockerReply {
 export interface InjectTarget {
   /** The console window HWND (decimal string). */
   hwnd: string;
-  /**
-   * The pid that OWNS the console window — whatever `GetWindowThreadProcessId(hwnd)` returns for the
-   * `ConsoleWindowClass` window. L3 supplies `getWindowProcessId(hwnd)` and the locker re-verifies
-   * with the SAME API on the same hwnd, so the value matches by CONSTRUCTION (L3 plan §4). The
-   * `ConsoleWindowClass` allowlist is what excludes a WT multiplexer; this is the window-owning pid,
-   * not asserted to be a specific conhost-vs-shell process (Opus R1 P3-1).
-   */
+  /** The console-HOST (conhost.exe) pid that owns the window — NOT the child pid (§2.2). */
   consolePid: number;
   /** Opaque hash of the expected pane identity (secondary anchor). */
   titleFp: string;
--- a/src/engine/key-locker/command-derivation.ts
+++ b/src/engine/key-locker/command-derivation.ts
@@ -143,12 +143,14 @@ export function tokenizeCommandSegmentsWithOps(cmd: string): CommandSegment[] {
       i++; // consume the `&`
       continue;
     }
-    if (ch === "&" && (cmd[i - 1] === ">" || cmd[i - 1] === "<" || cmd[i + 1] === ">")) {
-      // A `&` that is part of a REDIRECTION — fd-dup `2>&1` / `>&2` / `<&-` (abutting a `>`/`<`) or
-      // redirect-both `&>file` / `&>>file` (followed by `>`) — is I/O plumbing, NOT a job-control
-      // background operator. Keep it as a literal token char so an UPSTREAM `ssh host 2>&1 | tee log`
-      // (the `2>&1 |` long form of the `|&` case above) is not mis-marked `backgrounded` and wrongly
-      // denied its remote-frame push (Codex #495 P2). A job-control `&` never abuts a `>`/`<`.
+    if ((ch === "&" && (cmd[i - 1] === ">" || cmd[i - 1] === "<" || cmd[i + 1] === ">")) ||
+        (ch === "|" && cmd[i - 1] === ">")) {
+      // A `&` or `|` that is part of a REDIRECTION operator, not a job-control / pipe separator: `&` in
+      // fd-dup `2>&1` / `>&2` / `<&-` (abutting a `>`/`<`) or redirect-both `&>file` (followed by `>`),
+      // and `|` in noclobber-override `>|file` (abutting a `>`). Keep it a literal token char so the
+      // redirect stays ONE token — the ssh segment is then neither mis-split nor mis-marked
+      // `backgrounded`, so an upstream `ssh host 2>&1 | tee log` still pushes its remote frame (Codex
+      // #495 P2, Opus R6 P3). A job-control `&` / a real pipe `|` never abuts a `>`/`<`.
       cur += ch; started = true; continue;
     }
     if (ch === "&" || ch === "|") {
@@ -277,8 +279,11 @@ async function deriveGit(args: string[], session: SessionContext, exec: ExecFn):
   if (session.isRemote) return null;
 
   // Global options before the subcommand: honor `-C <path>` (changes the effective cwd), skip
-  // `-c k=v` / `--git-dir=…`-style value options, null on version/help.
-  let cwd: string | undefined = session.cwd;
+  // `-c k=v` / `--git-dir=…`-style value options, null on version/help. An empty-string cwd is a
+  // caller's "unknown" sentinel (the guard below already declines on it) — normalize it to undefined
+  // UP FRONT so a relative `-C sub` is not anchored at the agent's process cwd via `resolvePath("", …)`
+  // before that guard is reached (Codex #495 P2 wrong-target).
+  let cwd: string | undefined = session.cwd !== undefined && session.cwd.length > 0 ? session.cwd : undefined;
   let i = 0;
   let sub: string | undefined;
   for (; i < args.length; i++) {
--- a/src/engine/key-locker/command-derivation.ts
+++ b/src/engine/key-locker/command-derivation.ts
@@ -40,15 +40,23 @@ export interface DeriveBindingDeps {
   exec?: ExecFn;
 }
 
-/** One command segment plus whether the shell reaches it CONDITIONALLY (preceded by `&&` / `||`). */
+/** One command segment plus how the shell reaches it (conditional guard / background). */
 export interface CommandSegment {
   tokens: string[];
   /**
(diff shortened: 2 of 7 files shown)


Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
Round 4 applied (commit 612286f) — the three live wrong-target holes from the last two Codex reviews:

- **P1** non-session ssh stopped the scan (`ssh -G prod ; ssh host-b` / `ssh prod uptime ; cd X`) → now only a decisive interactive push stops the scan; a query/one-shot continues.
- **P1** backgrounded `ssh host &` pushed a remote frame → `CommandSegment.backgrounded` added; a backgrounded ssh is non-session (no push, keep scanning).
- **P2** guarded pipeline `false && echo ok | ssh prod` lost its guard → a single `|` now propagates the conditional across the pipeline; the ssh sinks to UNKNOWN.

155 key-locker tests green, build + lint clean. Comment-only threat-model invariants retained.

@codex review

Harusame64:
Round 4b (commit d9bb047) — closes the general **unguarded** `cmd | ssh h` wrong-target that Round 4's pipe change only half-covered (guarded variant only).

A downstream pipe stage's ssh has its **stdin bound to the pipe, not the tty**, so it cannot become an interactive login — the pane returns to the local prompt. Pushing a remote frame there wrong-targeted a later local sudo/git to host h.

- `CommandSegment` gains `pipedStdin` (downstream side of a single `|`).
- `recordDispatch` treats a piped-stdin ssh, like backgrounded / one-shot / query, as non-session (no push, keep scanning).
- Asymmetry pinned: the **upstream** stage `ssh h | tee log` keeps the tty → still pushes.

157 key-locker tests green, build + lint clean. @codex review

Harusame64:
Round 5 (commit b2819d1) — Codex R5 surfaced 5 more wrong-target vectors (P1x2 + P2x3). All fixed conservatively (precise where determinate, sink-to-UNKNOWN where ambiguous):

- **P1** trailing after an interactive login (`ssh a ; ssh b` / `ssh a ; cd X`) dropped the post-exit trajectory → now sinks to UNKNOWN when a SEQUENTIAL command follows a login; a downstream pipe stage (`ssh a | tee`, concurrent) still pushes.
- **P1** Windows backslash paths (`cd C:\repo` was stripped to `C:repo`) → tokenizer keeps `\` as a path separator, still escapes `\ ` / `\&` / `\"`.
- **P2** clustered arg-flags (`ssh -G -4p 2222 h` read `2222` as the host) → parseSshCommand walks the whole cluster; a with-arg letter anywhere consumes its value.
- **P2** bash `|&` treated as a pipe (`2>&1 |`), not a background `&`.
- **P2** `ssh -W host:port bastion` (stdio-forward, implies -N/-T) added to the non-session set beside -N/-f.

+9 test pins; 165 key-locker tests green, build + lint clean.

@codex review

Harusame64:
Round 6 (commit 3e5c596) — the two P2 wrong-target vectors Codex R6 surfaced, both around shell ampersands the R5 tokenizer mis-read as job control:

- **P2** redirect `&` is not a background job (`ssh host 2>&1 | tee log`): the `&` inside `2>&1` (fd-dup) / `>&2` / `<&-` / `&>file` is I/O plumbing. Two-part fix so the upstream ssh actually **pushes** (matching the `|&` and `ssh h | tee log` equivalence, never staying local → wrong-targeting a later local sudo to the remote host):
  - tokenizer keeps a redirect `&` as a literal token char (abutting `>`/`<`, or followed by `>`) → the ssh segment is no longer mis-marked `backgrounded`;
  - `interactiveSshTarget` strips trailing redirect tokens before the one-shot check, so a whole `2>&1` token no longer looks like a remote command and demote the interactive login to a one-shot.
- **P2** backgrounded / piped `cd` has no parent-shell effect (`cd C:/other &`, `… | cd C:/other`): a background job / pipeline stage runs the `cd` builtin in a **subshell**, so the foreground prompt stays put. `recordDispatch` now leaves the pane cwd **as-is** (mirrors the backgrounded/piped ssh non-session rule), so a later configured-remote `git -C` can't bind the wrong repo/host.

+6 test pins; 171 key-locker tests green, build + lint clean.

@codex review

Harusame64:
Round 7 (commit 16506b6) — four wrong-target holes around shell redirections + an empty-cwd sentinel (Codex R7 P2x2 + Opus R6 P2/P3):

- **P2 (Codex)** redirect **before** the destination (`ssh 2>&1 host`, `ssh >log host`): the R6 trailing-only strip left a *leading* redirect in place → `parseSshCommand` read it as the destination and the real host as a one-shot → pane stayed local while an interactive login opened. Now strip output redirects from the **whole argv** up front.
- **P2 (Codex)** empty-string cwd sentinel + relative `-C sub`: `cwd=""` let `resolvePath("", "sub")` anchor the `-C` path at the **agent process cwd** → configured remote resolved from the wrong repo. Normalize `""` → `undefined` before `-C` resolution.
- **P2 (Opus)** space-separated fd-dup target (`ssh host >& log`): a lone `>&` must consume its **following** token as the file; the regex matched only `>` and stranded `log` as a fake remote command → one-shot → local. Strip helper now recognizes a bare operator + drops its target token.
- **P3 (Opus)** `>|` (noclobber-override) mis-split: the tokenizer redirect guard now also keeps a `|` abutting a `>` literal, so `>|file` stays one token.

Also guards a regression the full-argv strip would have introduced: **only OUTPUT redirects are stripped**. An INPUT redirect (`ssh host < file`, heredoc, `<&-`) feeds ssh's stdin from a non-tty → non-interactive, so those tokens are kept and the ssh correctly classifies as a one-shot (no push).

+5 test pins; 176 key-locker tests green, build + lint clean.

@codex review

Harusame64:
Round 8 (commit ced3635) — the last member of the redirect/fd grammar class Codex R8 surfaced:

- **P2 (Codex)** an explicit fd-0 output redirect (`ssh 0>stdin.log deploy@prod`): `0>file` is an output *operator* but its fd is **0 = stdin**, so it takes ssh's stdin off the tty exactly like `< file` and makes the ssh non-interactive. The `OUTPUT_REDIR_OP_RE` leading-fd group was `(?:\d+)?`, so it stripped `0>…` as harmless plumbing, left a lone host, mis-classified it as an interactive login, and pushed a remote frame → a later local sudo/git would wrong-target the remote host. Fixed: leading fd is now `[1-9]\d*`, so fd 0 (`0>f`, `0>|f`, `0>&2`) stays in argv and the trailing-token count classifies the ssh as a one-shot (no push) — the same handling as input redirects. Non-zero numbered output redirects (fd 1/2/10) are unaffected and still strip.

Tests: fd-0 before/after the destination + fd-dup all keep the pane local; a fd-10 output redirect still pushes (no regression on ordinary numbered fds). Build + the 48-case session-tracker suite green.

With fd-0 folded in, the redirect taxonomy is now complete over the shell grammar: output redirects touching only fd 1/2 (and `&>`) strip and keep the login interactive; anything that redirects fd 0 / stdin (`<`, `<<`, `<&`, `<>`, `0>…`) is kept and makes the ssh non-interactive.

@codex review

Harusame64:
Round 9 (commit 93eb202) — Codex R9 found two more redirect wrong-target shapes (a redirect **before** the program `>log ssh host`, and a **nonzero-fd input** redirect `ssh 3<in host`). Rounds 2–8 chipped at this class one shape at a time. Rather than patch a 9th shape, R9 **collapses the entire redirect class onto the ONE bit that decides whether an ssh keeps the pane's tty on stdin: does the redirect touch fd 0?**

- **touches fd 0** (`<`, `<<`, `<&`, `<>`, `0<…`, `0>…`) → stdin off the tty → ssh non-interactive → **no push**.
- **does not touch fd 0** (output on fd ≥1, `&>`, nonzero-fd input like `3<`) → pure plumbing → **strip** so it can't skew destination / trailing-command parsing.

This single axis subsumes **position** (before/after the program), **fd number**, and **direction** (`<` vs `>`) — the three dimensions the prior rounds were enumerating by hand. Two regexes (`STDIN_REDIR_RE` / `NONSTDIN_REDIR_RE`) + `redirTokenAt` classify each token; `scanRedirectionsForStdin` reports the stdin bit for ssh, `stripRedirections` cleans `cd`'s argv, and `recordDispatch`'s scan now skips leading redirects (not just env-assigns) to find the program token.

Tests (+4): redirect-before-program pushes; leading fd-0 stays local; nonzero-fd input pushes; `cd > log /srv` records `/srv` not `>`. 52 session-tracker + 12 ssh-resolve green, build clean.

@codex review

Harusame64:
Round 10 (commit 292b24a) — Codex R9's second wave moved off the (now structurally-closed) redirect class onto two shell-syntax classes: a `#` comment and here-doc bodies. This round **closes the comment case** and **defers here-doc** to the capture-loop wiring PR, with rationale below.

**Comment — FIXED.** The tokenizer left an unquoted `#` and its trailing words in the argv, so `ssh host # note` (an interactive login once the shell strips the comment) was mis-read as `ssh host <remote-command>` → one-shot → no push → a later remote sudo/git wrong-targeted localhost. Fix: a `#` at a **word boundary** (not mid-word, not quoted) starts a comment stripped to end-of-line; the `\n` is left for the segment separator so following lines still parse. Correct for L1's `deriveBinding` too. Mid-word `#` (`--format=%h#%s`, `a#b`) and quoted `"a # b"` stay literal, matching the shell.

**Here-doc — DEFERRED to the capture-loop wiring PR** (tracked in the internal handoff doc's "deferred from PR #495 — MUST close in the wiring PRs" section). Reasoning: (1) whether a multi-line here-doc even *reaches* `recordDispatch` as one string is a **dispatch-framing decision owned by the capture-loop wiring**, not this pure tracker; (2) correct here-doc skipping (`<<-` tab-strip, quoted/`'DELIM'` delimiters, multiple here-docs per line, delimiter matching) is its own grammar mini-class that would re-open a redirect-style whack-a-mole if bolted onto the tracker blind to the framing; (3) L3 detection is **non-load-bearing** — the wrong-target defense is L1 fingerprint + L2 injection-moment re-verify + the exit-0/auth-accepted save-gate — so a deferred here-doc mislabel on an uncommon multi-line dispatch is defense-in-depth degradation, not a breach. The wiring PR will decide dispatch granularity and, if multi-line here-docs can arrive, make the tokenizer skip `<<DELIM`→`DELIM` (bodies are never commands) with `<<-`/quoted/multi tests.

Tests (+2): tokenizer comment-strip (word-boundary vs mid-word vs quoted vs comment-only); `ssh host # note` pushes. 101 tracker/derivation + ssh-resolve green, build clean.

@codex review

Harusame64:
Re: the Codex R10 P2 (`#` stripping is shell-dependent — cmd.exe has no `#` comment, it uses `rem`). Valid observation; **deferring the refinement to the capture-loop wiring PR** rather than patching it here, for these reasons:

- **R10's strip is correct for the majority pane.** PowerShell (the Windows Terminal default) and bash both treat a word-boundary `#` as a comment, so R10 fixes the common case. It is wrong only for **cmd.exe**, where `#` is a literal argument — the lower-probability pane.
- **The precise fix needs the pane's SHELL IDENTITY, which the pure tracker does not have.** Gating `#`-stripping on shell ∈ {pwsh, bash} (and keeping `#` literal / `markUnknown` for cmd or an unknown shell) requires threading the shell into the tokenizer. That identity is owned by the **capture-loop wiring** (it launches/observes the pane), not this state-only tracker.
- **Same bucket as here-doc + the `&&`⏎ line-continuation** — all three are shell/dispatch-framing–dependent and are tracked together in the internal handoff doc to **close in the capture-loop wiring PR**.
- **L3 detection is non-load-bearing.** The wrong-target defense is L1 fingerprint + L2 injection-moment re-verify + the exit-0/auth-accepted save-gate, so a cmd-pane `#` mislabel is defense-in-depth degradation on an uncommon shape, not a breach.

Net for this PR: the **redirect/fd grammar class is structurally closed** (R9, one "touches fd 0?" bit subsuming position × fd × direction), the comment case is fixed for the common shells, and the three residual shell/framing-dependent refinements (cmd `#`, here-doc bodies, `&&`⏎) are documented as MUST-close-in-wiring-PR follow-ups. Opus R10: P1=0, P2=0 on the tracker itself. Handing off for merge.

PR #496, opened 2026-07-02
ADR-014 v2 R3 Key Locker — L3 §4: InjectTarget assembly

Harusame64:
## What

L3 §4 (LOCKED CONTRACT #4) — the `pane` SendInput target assembly for the Key Locker capture-on-use loop. One small engine module + unit tests.

`assembleInjectTarget(hwnd, submit): InjectTarget | null`
- `hwnd` → decimal string (the `InjectTarget.hwnd` contract; the locker's `ParseLong` accepts a decimal string).
- `consolePid` = `getWindowProcessId(hwnd)` — the window-owning pid.
- `titleFp` = `sha256hex(utf8(GetWindowTextW(hwnd)))` — read via the SAME Win32 APIs the locker re-verifies with, so both match by construction (a title change between derive and inject aborts `target_mismatch`).
- Declines (`null`) when the window is gone / hwnd invalid (`getWindowProcessId === 0`) — no garbage target (whose `titleFp` would be `sha256("")`) is round-tripped to the locker.

`titleFingerprint(title)` is **byte-identical to `Injection.cs` `TitleFp`** — pinned in the test against expected hex computed **independently with .NET** (`[SHA256]::HashData([Text.Encoding]::UTF8.GetBytes(t))`, lowercased hex), including a CJK title, so an encoding/algorithm drift is caught here rather than as a silent inject-time abort.

## Scope / design notes

- Channel selection (`pane`/`askpass`/`git-credential`) and `submit` classification stay in the capture-loop (§4 second half / §5 detection); this module is handed the resolved hwnd + submit flag.
- **No engine→tools import**: the tool layer that observed the credential prompt already located the console window, so it passes the hwnd down — `inject-target.ts` does not import `findTerminalWindow` from `src/tools/`.

## Tests

`tests/unit/key-locker-inject-target.test.ts` — 8 tests: 4 cross-language byte-match fingerprints (incl. empty + CJK) + lowercase-hex shape + assembly (decimal hwnd, same-hwnd reads, optional `submit`, gone-window decline).

Build clean, lint clean, 8/8 green.

Plan: `desktop-touch-mcp-internal@<sha>:docs/adr-014-v2-r3-l3-capture-plan.md` (§4)

### DIFF
--- a/src/engine/key-locker-host.ts
+++ b/src/engine/key-locker-host.ts
@@ -97,7 +97,7 @@ export interface LockerReply {
 
 // L2 wire contracts (the locker owns these frame shapes; the injector orchestrator consumes them).
 
-/** The dedicated-conhost target of a SendInput (`inject`) — §2.1 of the L2 plan. */
+/** The dedicated-console target of a SendInput (`inject`) — §2.1 of the L2 plan. */
 export interface InjectTarget {
   /** The console window HWND (decimal string). */
   hwnd: string;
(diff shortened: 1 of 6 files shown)


Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
@codex review

PR #502, opened 2026-07-05
ADR-014 v2 R3 Key Locker — L3 §1: capture-loop (capture-on-use state machine)

Harusame64:
## What

The **capture-on-use state machine** (L3 §1) — the Chrome-model loop that turns L1's binding + L2's injectors into "the locker learns and fills credentials".

A credential prompt L3 dispatched is detected → the binding is derived from the **dispatched command** (L1, authoritative — never the spoofable prompt text) → then:

- **MATCH** (a stored secret exists): D2 confirm backstop (per-binding policy) → autofill. Nothing to capture or save.
- **NO MATCH**: capture into the secure dialog → inject the just-captured secret → **landed-detect** (§2 two-mode) → **OFFER** → persist the mapping **only on [Save]** (P1-2: the offer precedes the persist, so a declined save never lingers). Every non-commit exit — reject / not-landed / [Not now]/[Never] / **any throw** — **deletes** the captured secret from the locker (P2-1 reverse-orphan closure of L1 §5.3).

## Scope (L3-2)

This is the **PANE (SendInput) channel** loop — the fully-reactive "echo-off prompt appeared, fill it now" flow (`sudo`/`ssh` password prompts). `selectInjector(scheme, "pane")` gates it: an `sshkey` (always askpass — P3-2) or `https-cred` (git-credential) binding is **not pane-injectable** and DECLINES this loop, handled by the forward askpass/git-credential flow (SP-L3-OQ-5, **not built here** — its inject happens at dispatch time with a pre-set env and its landed signal spans the consumer's whole lifetime, a different control flow).

## Design

- **Pure orchestrator**: no Win32 / host / terminal import. Every effect is an injected `CaptureLoopDeps` seam, so tests drive fakes and the KeyLockerManager (L3-3) wires the live primitives + the confirm/offer dialogs (secret-free, like the `-Consent` dialog).
- Records **no** dispatch into the SessionTracker — the §3 derive-then-record ordering is the wiring's job (the loop only reads `getSession`).
- **Touches no merged file** (zero blast radius).
- **Secret invariant** (L0/L1/L2): the secret is entered into the locker's secure dialog, injected by the locker, and never crosses the control pipe / touches Node / hits the raw pane. This loop handles opaque ids + landed verdicts only.

## Deferred (forward obligations, recorded in the impl-handoff doc)

- The askpass / git-credential deferred-fill flow (SP-L3-OQ-5).
- The confirm/offer UI mechanism (a secret-free locker dialog mode, like `-Consent`) — wired in L3-3/L4.
- The "never" persistence + pre-capture gate (the loop discards on `not_now`/`never` identically; remembering "never for this binding" is L4 management storage).
- `confirmPolicyFor` is a deps abstraction here; L3-3 binds it to the binding row's `confirmEveryInjection ?? true`.

## Tests

22 unit tests (declines / MATCH+confirm / NO-MATCH save+discard / P1-2 offer-before-persist ordering / P2-1 delete-on-throw / per-scheme meta). Full key-locker suite **258/258** green. `build` + `eslint` + `check:failwith-fixtures` clean. No Rust/native/tool-catalogue change.

Plan: `desktop-touch-mcp-internal:docs/adr-014-v2-r3-l3-capture-plan.md` (§1)

### DIFF
--- a/src/engine/key-locker/capture-loop.ts
+++ b/src/engine/key-locker/capture-loop.ts
@@ -108,6 +108,15 @@ export interface CaptureLoopDeps {
   confirmInjection(binding: BindingUri): Promise<boolean>;
   /** The Chrome-model save offer AFTER a landed NO-MATCH fill (P1-2 — the loop persists only on "save"). */
   offerSave(binding: BindingUri): Promise<SaveChoice>;
+  /**
+   * Record a "[Never] — don't offer to save this binding again" decision (plan §1 `policyStore.setNever`).
+   * OPTIONAL: the loop ALWAYS discards on "never" regardless; this seam lets the wiring persist a tombstone
+   * so the save is not re-offered next time. The tombstone STORE (a negative-binding store, distinct from
+   * the saved-row `confirmEveryInjection` policy) + the pre-capture suppression it enables are L4 management
+   * surface — DEFERRED (SP-L3-OQ handoff). When unwired, "never" simply behaves like "not now" (re-offered
+   * next prompt). The seam EXPRESSES the §1 contract point here rather than silently dropping it (強制命令 9).
+   */
+  onNever?(canonicalKey: string): void;
   /** Mint the opaqueId the loop will bind on save (`randomBytes(16).toString("hex")`; a fake in tests). */
   mintOpaqueId(): string;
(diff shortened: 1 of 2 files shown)


Harusame64:
@codex review

PR #503, opened 2026-07-05
ADR-014 v2 R3 Key Locker — L3-3 PR1: fix inverted confirm-policy default + getPolicy

Harusame64:
## What & why

The `key_locker` tool and the credential-autofill loop disagreed on the default of the per-binding **autofill-confirmation** policy — a user-facing safety bug, not cosmetics.

- **The loop (authoritative consumer):** an unset binding **confirms** every autofill (the safe backstop); `confirmEveryInjection: false` is the per-binding **opt-out**.
- **The tool (before this PR):** `list` displayed unset as `false` ("won't ask") while the loop **does** ask, and `set_policy`'s help said `false = default` — so a user setting `false` to "restore the default" would silently **turn the confirmation off**.

This PR makes the tool tell the truth and match the loop.

## Changes
- `BindingStore.getPolicy(key)` — the resolved policy `confirmEveryInjection ?? true` (additive; unknown/unset ⇒ confirm).
- `key_locker` `list` reports the resolved policy; `set_policy` help + the metadata doc state the correct default (unset/`true` = confirm; `false` = opt out).
- Adds tool-level `save` tests for the **ssh** path (a test-only faked `ssh -G`/`ssh-keygen`) that pin the save↔autofill key parity and the "host not in known_hosts ⇒ fail before capture" behavior.

## Verification
- `npm run build` clean; `npm run lint` 0 errors.
- `key-locker-*` unit suite 267 passed; `tools-list-schema` 6 passed.
- No public tool added ⇒ no catalogue change.

### DIFF
--- a/src/engine/key-locker/binding-store.ts
+++ b/src/engine/key-locker/binding-store.ts
@@ -40,9 +40,11 @@ export interface BindingMeta {
   /** ISO-8601, stamped by the caller. */
   createdAt: string;
   /**
-   * Per-binding injection policy (L4 §1 `set_policy`): when true, the locker asks the user to confirm
-   * EVERY autofill for this binding (a per-binding opt-in — never a global no-confirm). Absent/false =
-   * the default confirm-every behaviour. The L1 plan §5.1 RESERVED this as "an L3 field, not set in L1";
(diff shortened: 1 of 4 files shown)


Harusame64:
@codex review

PR #504, opened 2026-07-05
ADR-014 v2 R3 Key Locker — L3-3 PR2: [Never] tombstone store + capture-loop suppression

Harusame64:
## What & why

When the credential-autofill loop offers to save a newly-captured secret, the user can answer **[Never]**. Until now that answer was honored for the current prompt but not remembered, so the same "save this?" offer would reappear next time. This adds the persistent tombstone that makes **[Never]** stick.

## Changes
- `never-store.ts` (new): a small `never.json` sidecar next to the locker's data — `has`/`add`, atomic writes, tolerant load (a corrupt/missing file just starts empty). Stores only binding identifiers, never a secret.
- `capture-loop.ts`: a `[Never]`'d binding is now skipped **before** the secure-entry dialog even opens, returning a `never_suppressed` decline. This only affects the *save offer* for a binding with no stored secret — a binding that already has a saved secret still autofills normally (so manually re-saving a previously-declined binding works).

## Verification
- `npm run build` clean; lint clean on the changed files.
- `key-locker-*` unit suite 273 passed (adds the `never-store` suite + loop-suppression cases).
- Additive only — no public tool / schema / catalogue change.

### DIFF
--- a/src/engine/key-locker/capture-loop.ts
+++ b/src/engine/key-locker/capture-loop.ts
@@ -65,8 +65,10 @@ export interface CredentialEvent {
  */
 export type CaptureLoopOutcome =
   /** Nothing touched the locker: the pane session is UNKNOWN (§3 fail-safe), the command carries no
-   *  credential L1 attributes, or the binding is not pane-injectable (askpass/git-credential forward flow). */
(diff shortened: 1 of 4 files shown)


Harusame64:
@codex review

PR #505, opened 2026-07-05
ADR-014 v2 R3 Key Locker — L3-3 PR3: ssh session-end watch (OQ-7) + idle-dispose

Harusame64:
## What & why

When you `ssh` into a remote host from a tracked terminal pane, the key locker remembers you're now "on that host" so it fills the right credentials. This PR adds the missing half: noticing when that ssh session **ends**, so the pane goes back to "local" and a later local command doesn't get a remote host's secret (a wrong-target fill).

Because an interactive `ssh user@host` never exits with a code we can watch, the reliable end-of-session signal is the **ssh child process disappearing** — observed from the process tree.

## Changes
- `ssh-session-watch.ts` (new): watches each pane's ssh child process and, when it exits, tells the session tracker. Pure logic over an injected process snapshot (fully unit-tested; the live Win32 snapshot + timer are wired later with the terminal integration).
- **Nested ssh guard:** only the outermost ssh is visible locally, so when the tracker is 2+ hops deep the watch can't tell which hop ended — it conservatively marks the pane "unknown" (declines to autofill) rather than risk targeting the wrong host. Adds `SessionTracker.remoteDepth()`.
- `KeyLockerManager.disposeIfIdle()`: shuts the locker helper down after an idle period, never mid-operation.

## Verification
- `npm run build` clean; lint clean on changed files.
- `key-locker-*` unit suite 276 passed (adds the watch suite + `remoteDepth` + `disposeIfIdle` cases).
- Additive only — no public tool / schema / catalogue change.

### DIFF
--- a/src/engine/key-locker/key-locker-manager.ts
+++ b/src/engine/key-locker/key-locker-manager.ts
@@ -74,6 +74,8 @@ export function keyLockerDisabled(): boolean {
 export interface KeyLockerManagerOptions extends KeyLockerStartOptions {
   /** Override the store dir (tests); production uses the locker default. */
   storeDir?: string;
+  /** Injected monotonic-ish clock for idle-dispose dormancy (default `Date.now`); overridden in tests. */
+  now?: () => number;
 }
 
 /**
@@ -86,8 +88,14 @@ export class KeyLockerManager {
   private host: KeyLockerHost | null = null;
   private starting: Promise<KeyLockerHost> | null = null;
   private consenting: Promise<void> | null = null; // in-flight `-Consent` dialog (dedupes concurrent prompts)
+  private inFlight = 0;        // secret ops currently running through withHost (idle-dispose never runs mid-op)
+  private lastActivityMs = 0;  // clock stamp of the last withHost completion (idle-dispose dormancy timer)
 
   constructor(private readonly opts: KeyLockerManagerOptions = {}) {}
+
+  private now(): number {
+    return (this.opts.now ?? Date.now)();
+  }
 
   /** The store dir this manager reads consent + bindings from. */
   get storeDir(): string {
@@ -116,7 +124,30 @@ export class KeyLockerManager {
       throw new KeyLockerConsentRequiredError();
     }
     const host = await this.ensureHost();
-    return fn(host);
+    // Count the op as in-flight so idle-dispose never tears the host down mid-operation (e.g. while a
+    // capture dialog is open for minutes); stamp the activity clock on completion (dormancy timer).
+    this.inFlight++;
+    try {
+      return await fn(host);
+    } finally {
+      this.inFlight--;
+      this.lastActivityMs = this.now();
+    }
+  }
+
+  /**
+   * Idle-dispose (dormancy): tear the live host down if no secret op has run for `idleMs`, so a locker
+   * that spawned once but is no longer in use does not linger for the process lifetime (the long-lived-
+   * resource discipline). The assembly calls this on a timer. Returns whether it disposed. NEVER disposes
+   * while an op is in flight (`inFlight > 0`) or when nothing is live. `dispose()` awaits any in-flight
+   * start, so this is safe to race with a lazy `withHost`.
+   */
+  async disposeIfIdle(idleMs: number): Promise<boolean> {
+    if (this.inFlight > 0) return false;                              // an op is running — not idle
+    if (this.host === null && this.starting === null) return false;   // nothing live to dispose
+    if (this.now() - this.lastActivityMs < idleMs) return false;      // used within the window
+    await this.dispose();
+    return true;
   }
 
   /**
--- a/src/engine/key-locker/key-locker-manager.ts
+++ b/src/engine/key-locker/key-locker-manager.ts
@@ -123,11 +123,13 @@ export class KeyLockerManager {
     if (!this.isConsentAccepted()) {
       throw new KeyLockerConsentRequiredError();
     }
-    const host = await this.ensureHost();
-    // Count the op as in-flight so idle-dispose never tears the host down mid-operation (e.g. while a
-    // capture dialog is open for minutes); stamp the activity clock on completion (dormancy timer).
+    // Count the op as in-flight BEFORE acquiring the host, so idle-dispose never tears the host down
+    // mid-operation NOR during its lazy START (Opus R1 P2-B: `inFlight++` after `ensureHost` left the
+    // start phase — and every post-idle restart — unguarded, so a racing `disposeIfIdle` could dispose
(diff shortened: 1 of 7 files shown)


Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
@codex review

Round 2 findings applied in 28c801ab:
- **P2 (Codex R2, confirmed by Opus)** — the tick() exit check read liveness from `identify()` alone, but `getProcessIdentityByPid` returns an empty identity for BOTH a gone pid AND a live-but-unreadable one (OpenProcess ACCESS_DENIED on an elevated/other-user ssh). A still-running registered session ssh whose identity read transiently failed was misread as exited → pop-to-local while still remote → inverse wrong-target. Fixed: liveness now read from the Toolhelp `parentMap`; a present-but-empty identity is doubt ⇒ `markUnknown`, not a pop. Confirmed pid-reuse still pops via the depth pivot.
- Contract docs corrected ("" = gone OR present-but-unreadable). New lock test added.

Harusame64:
@codex review

Round 3 findings applied in 5db1a196:
- **P2 (Codex R3, confirmed by Opus)** — the same empty-identity conflation remained at the REGISTRATION site. `noteSshOpened` treated `identify().name !== "ssh"` (which includes the `""` a live-but-unreadable pid returns — OpenProcess ACCESS_DENIED on an elevated/other-user ssh) as "not a live ssh", silently dropping a real registration → the just-pushed remote frame was never watched → stale-remote wrong-target (REMOTE secret into a LOCAL prompt).
- Fixed: register only a pid positively confirmed as a live `ssh` via the Toolhelp `parentMap.has` liveness gate + a readable `ssh` identity. Otherwise clear the slot and, if a frame was already pushed (`remoteDepth > 0`), sink it to `markUnknown` rather than strand an unwatched `isRemote:true`. A speculative call on a still-local pane (depth 0) preserves the local anchor.
- Added a systemic `tick` backstop (`session === null && remoteDepth > 0` ⇒ `markUnknown`) for the non-atomic push↔register window, plus a push-before-register contract in the TSDoc. Both wrong-target directions are now closed. 89 key-locker unit / 4374 total green.

Harusame64:
@codex review

Round 4 findings applied in 381a4211 — folded to a single structural axis (per our process for a recurring class) rather than two more point-fixes. Adjudicated with Opus.

**STRUCTURAL AXIS**: the watch trusts EXACTLY ONE state — the single depth-1 outer ssh, alive, readable as `ssh`, with a NON-ZERO creation time matching the registered baseline. Every other state declines to `markUnknown`. This subsumes R2/R3 and both R4 edges:
- **R4 #1 (nested ssh)** — `ssh a` then `ssh b`: the inner ssh runs on host-a, invisible locally, so its end can never be observed. Now a nested pane (`remoteDepth >= 2`) declines at BOTH `noteSshOpened` and `tick` (self-enforced, not trusting wiring discipline). The old depth-pivot `else markUnknown` (>= 2) is now formally unreachable — kept as a defensive assertion.
- **R4 #2 (zero creation time)** — verified src/win32/process.rs:144-175 reads name/time independently, so a live ssh can return `{name:"ssh", startTimeMs:0}`. The trusted-continue now requires `startTimeMs !== 0 && === baseline`; a live ssh with time 0 is `markUnknown` (doubt), and registration refuses to store a 0 baseline.

Also documented the `recordDispatch` isRemote-only-for-ssh invariant + the P3 shell-pid-reuse residual (self-healing, tracked in plan, not gated). 21 ssh-watch unit tests / 4394 total green; CI build now passes (the earlier failure was a pre-existing #503 failwith-fixture drift, fixed by merging main + regenerating the fixture).

Harusame64:
@codex review

Round 5 applied in 99fee481. This completes the R4 fold's own stated invariant (the file header claimed "trust depth EXACTLY 1" but the tick only declined `remoteDepth >= 2`).

**R5 (Codex, adjudicated with Opus)** — the trusted-continue now runs ONLY at `remoteDepth === 1`. The `depth !== 1` guard is liveness-gated at depth 0 (Opus's fixed point, not a flat markUnknown):
- **depth 0 + a LIVE registered ssh** ⇒ `markUnknown` — the tracker re-anchored local while the ssh lives; trusting local would derive a LOCAL binding for a live remote (wrong-target).
- **depth 0 + a GONE pid** ⇒ DROP the stale watch, KEEP the local anchor — a benign re-anchor race, so a flat markUnknown would needlessly downgrade a correct local pane.
- depth >= 2 (nested) ⇒ markUnknown (unchanged from R4).

This is the SAME monotone axis as R1–R4 (each round declines MORE, never re-trusts) — R3's "leave depth-0 local" was for `session === null`; R5 declines `session !== null` + live. The confirmed-exit `else markUnknown` (depth >= 2) is now provably unreachable (kept as an assertion). 23 ssh-watch / 4396 total unit green; build + lint clean.

Harusame64:
@codex review

**Round 6 applied (2af9921) + Opus closing-sweep certification — this is the pure-module fixed point.**

R6 fixes:
- **CodeQL** "useless comparison": removed the now-dead `else markUnknown` on the confirmed-exit path (the R5 `!== 1` guard already declined depth 0 and ≥ 2, so `depth <= 1` was always true) → bare `noteSessionEnd()`.
- **P2 (watchPane)**: re-registration no longer silently drops a live ssh session — a re-anchor over a live session now `markUnknown`s before resetting (the R5 tick guard needs `session !== null` to see the contradiction).

**Convergence certification (Opus exhaustive contract-sweep).** This is a PURE module whose live wiring is deferred. It now reconciles **every steady (watch, tracker) state to safety within one tick** — the wrong-target class is provably closed at the watch layer. The only residual is the **inter-tick window**: a credential fill derives straight from `tracker.get()` (never through this watch), so a wiring mutation can transiently create a wrong-target state readable before the next tick reconciles. A pure module cannot close that (no callback on tracker mutations; the fill bypasses it) — it is a **wiring obligation**, documented in the module header ("INVARIANTS THE LIVE WIRING MUST HONOR") + plan §Risks and **deferred to the live-wiring PR** per our defer policy:
- **W1 re-anchor atomicity** (Edge 3), **W2 close atomicity** (Edge 7), **W3 derive/mutate atomicity + tick liveness** (Edges 1/2/4/9).

So: a re-flag of edges W1/W2/W3 (re-anchor/unwatch/derive-across-window orderings) is a **known-deferred wiring contract**, not a new module bug. A genuinely new *pure-module* wrong-target is in scope. 26 ssh-watch unit tests green; build + lint clean; CI build previously green (the earlier failure was a pre-existing #503 fixture drift, fixed by merging main + regenerating the fixture).

PR #510, opened 2026-07-06
ADR-014 v2 R3 L3-4 W-0: getProcessCommandLineByPid native binding

Harusame64:
## What

New win32 native binding `getProcessCommandLineByPid(pid): string[] | null` — reads a process's launch argv.

This is **W-0** of the ADR-014 v2 R3 Key Locker **L3-4 live-wiring** slice set (the last-mile before the R3 release dogfood). It unblocks the OQ-W-7 = **option B** decision: the wrong-target fail-safe reads an unregistered ssh child's argv to decline ONLY an actual interactive in-bound login (not a tunnel / one-shot / scp / git-over-ssh), so those panes keep autofilling.

## How

- `NtQueryInformationProcess(ProcessCommandLineInformation)` (Win 8.1+) — needs only `PROCESS_QUERY_LIMITED_INFORMATION` (the SAME right the identity query already uses; **no** cross-process memory read — the OS copies the string into our buffer).
- `CommandLineToArgvW` splits it for **byte-exact parity** with how the process was actually launched (quote/backslash rules), so the caller can reuse the merged `interactiveSshTarget` classifier.
- Returns `null` on **any** failure (dead pid / ACCESS_DENIED on an elevated-or-cross-user target / non-success NTSTATUS / parse failure) → the caller **fails safe** (a null read = "possibly interactive" ⇒ decline, never disclose).
- RAII handle guard + `napi_safe_call` panic guard mirror the existing identity path; `LocalFree` releases the argv allocation.

## Tests / verification

- `cargo test -p desktop-touch-engine cmdline_tests` → **3/3 on real Windows** (reads the test runner's own argv; zero-pid ⇒ None; nonexistent-pid ⇒ None fail-safe, no panic).
- `tsc` + `check:napi-safe` clean.
- Native compiles (`cargo check --workspace`, `build:rs` compile step). Local `.node` artifact-copy is skipped only because the running MCP server holds the addon lock; CI builds it clean.

## Review focus (native axis)
`unsafe` correctness (UNICODE_STRING buffer bounds, `CommandLineToArgvW`/`LocalFree` lifetime), the ACCESS_DENIED / partial-read fail-safe-to-null contract, and the alloc cap on a bogus returned length.

Plan: `desktop-touch-mcp-internal@afcd0bb:docs/adr-014-v2-r3-l3-4-live-wiring-plan.md` (W-0)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -118,6 +118,10 @@ windows = { version = "0.62", features = [
     # Visual GPU Phase 4 capability detection (ADR-005)
     "Win32_System_SystemInformation",
     "Wdk_System_SystemServices",
+    # ADR-014 v2 R3 L3-4 W-0: NtQueryInformationProcess(ProcessCommandLineInformation)
+    # for getProcessCommandLineByPid (read an ssh child's argv to classify
+    # interactive-login vs tunnel/one-shot in the key-locker wrong-target fail-safe).
+    "Wdk_System_Threading",
     # ADR-007 P1: hot-path window APIs (koffi → windows-rs).
     # EnumWindows / GetWindowText / GetWindowRect / GetForegroundWindow /
     # IsWindowVisible / IsIconic / IsZoomed / GetClassNameW /
--- a/index.d.ts
+++ b/index.d.ts
@@ -369,6 +369,7 @@ export declare function win32ForegroundFlashInject(targetHwnd: bigint, targetPid
 export declare function win32GetFocusedChildHwnd(targetHwnd: bigint): bigint | null
 export declare function win32BuildProcessParentMap(): NativeProcessParentEntry[]
 export declare function win32GetProcessIdentity(pid: number): NativeProcessIdentity
+export declare function win32GetProcessCommandLine(pid: number): string[] | null
 export declare function win32GetScrollInfo(hwnd: bigint, axis: string): NativeScrollInfo | null
 export declare function win32PostMessage(hwnd: bigint, msg: number, wParam: bigint, lParam: bigint): boolean
 export declare function win32ConsolePasteNoFocus(hwnd: bigint, text: string): NativeConsolePasteResult
--- a/index.js
+++ b/index.js
@@ -104,6 +104,7 @@ export const win32ForceSetForegroundWindow    = nativeBinding.win32ForceSetForeg
 export const win32GetFocusedChildHwnd         = nativeBinding.win32GetFocusedChildHwnd;
 export const win32BuildProcessParentMap       = nativeBinding.win32BuildProcessParentMap;
 export const win32GetProcessIdentity          = nativeBinding.win32GetProcessIdentity;
+export const win32GetProcessCommandLine       = nativeBinding.win32GetProcessCommandLine;
 export const win32GetScrollInfo               = nativeBinding.win32GetScrollInfo;
 export const win32PostMessage                 = nativeBinding.win32PostMessage;
 export const win32GetFocus                    = nativeBinding.win32GetFocus;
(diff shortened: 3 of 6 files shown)


Harusame64:
@codex review

Focus: the `win32_get_process_command_line` unsafe block — UNICODE_STRING buffer bounds (Length is a byte count; Buffer points inside our local buffer), the CommandLineToArgvW / LocalFree pointer lifetime, the STATUS_INFO_LENGTH_MISMATCH grow path + the 128 KiB alloc cap, and the fail-safe-to-None contract on ACCESS_DENIED / non-success NTSTATUS.

Harusame64:
@codex review

Re-review at new HEAD. Since the last pass (`c78eef4a`) this branch added Round 2 (bound `wlen` by `ret_len` + `LocalFree` on the `argc<=0` edge) and a build-guard fix (added `win32GetProcessCommandLine` to index.d.ts to clear check:native-types).

Focus: the `win32_get_process_command_line` unsafe block — UNICODE `CommandLineToArgvW` lifetime / `LocalFree` on every return path, and the `ret_len` bound.

Harusame64:
@codex review

Round 3 applied your 3 P2 findings in win32_get_process_command_line:
- Alignment: header now read via `ptr::read_unaligned` into an aligned local (no `&UNICODE_STRING` over a byte-aligned Vec<u8>).
- Bounds: validate the whole [Buffer, Buffer+Length) span lies within the returned bytes + Length is even, before `from_raw_parts`.
- argv UTF-16: `to_string()` failure now frees argv and returns None (fail-safe) instead of `unwrap_or_default()`.

Please confirm the fixes and that the new bounds arithmetic can't under/over-flow.

Harusame64:
@codex review

Round 4 (final): added `(str_start & 1) != 0` to the bounds guard so an odd (misaligned) UNICODE_STRING.Buffer fails safe to None before the u16 `from_raw_parts` — closing the last alignment gap in win32_get_process_command_line. This supersedes the earlier rounds; please confirm no P1/P2 remain.

Harusame64:
@codex review

R4 (`094c4d18`) added an Opus P3 hardening on top of the R3 commit you last reviewed (`af931b8a`): the Buffer-pointer bounds guard now also rejects an odd `Buffer` base pointer before building the u16 slice (2-byte alignment). Please re-review the current HEAD.

Harusame64:
@codex review

Round 5 (`18d76da3`) resolves the index.d.ts↔index.js drift you flagged: it adds `export const win32GetProcessCommandLine = nativeBinding.win32GetProcessCommandLine` to index.js so the declared named export is real.

Note on runtime impact: this codebase consumes every win32 fn via the `requireNativeWin32()` probe on `addon.default` (single load point in native-engine.ts; `index.js` ends with `export default nativeBinding`), and nothing in src/ named-imports from index.js — so the runtime path (`win32.ts:421`) already worked. The re-export closes the latent `.d.ts` trap at zero cost. The broader `.d.ts`→`.js` parity gap (7 sibling win32 fns) is tracked as a follow-up.

The three `src/win32/process.rs` inline threads are your R2/R3 findings (unaligned header read, slice bounds + Length-even, argv-conversion fail-safe), all fixed in R3 `af931b8a`/R4 `094c4d18` and mechanically re-pinned to the new HEAD.

Harusame64:
@codex review

Round 6 (`1f5aa201`) addresses your P2 on the buffer-retry: `win32_get_process_command_line` now grows+retries on STATUS_INFO_LENGTH_MISMATCH, STATUS_BUFFER_OVERFLOW, and STATUS_BUFFER_TOO_SMALL (folded into one predicate), so a long-but-readable ssh command line is no longer misread as unreadable. Still fail-safe: a second non-success status falls through to None. `cargo test process` green. Please re-review the current HEAD.

PR #511, opened 2026-07-06
ADR-014 v2 R3 L3-4 W-1: terminal dispatch hook + read-path export

Harusame64:
## What

Slice **W-1** of the ADR-014 Key Locker live-wiring. Adds a fire-and-forget **dispatch hook seam** to `src/tools/terminal.ts` so an external module (the Key Locker wiring) can be notified of every command dispatched into a terminal pane, and makes the pane-tail read path importable. This mirrors the existing `setTerminalReadHook` pattern.

## Changes (additive only)

- **Seam** (top of module, near the schemas): `TerminalDispatchEvent { paneId, command }`, `setTerminalDispatchHook(fn|null)` register/clear, and `fireTerminalDispatch(paneId, command)`. The fire path is `try/catch`-isolated — a hook throw can never break a dispatch. Null observer is the default (no behaviour change when unset).
- **SEND handler** (`terminalSendHandler`): fires **once** after the null/valid-window guard with `String(win.hwnd)` + the **user's original `input`** (never a rewritten/quoted form).
- **RUN handler** (`terminalRunHandler`): fires **once** after the window is resolved, **before** `buildExitCommand` rewrites the input into `sendInput`, with the **user's original `input`**.
- **Read path**: `readTerminalRaw` is now `export`ed (behaviour unchanged) so the wiring can pull a pane's tail text.
- `fireTerminalDispatch` is exported so the new unit test can exercise the try/catch isolation directly (the in-handler fire path needs a live Win32 window).

## Invariants

- `paneId === String(win.hwnd)` — a **bare decimal** string (`win.hwnd` is `bigint`; `String(bigint)` === `bigint.toString(10)`, no `0x`/`n`/separators). Asserted in the test.
- Fires **exactly once** per resolved-window send and per resolved-window run.
- No existing dispatch behaviour, control flow, or test was changed.

## Tests

New `tests/unit/terminal-dispatch-hook.test.ts` (6 cases, no real terminal): registration + delivery, default no-op, clear-on-null, throwing-hook isolation, verbatim pass-through, and the `String(bigint)` bare-decimal paneId contract. All 6 pass; `npm run build` (tsc) passes.

Additive / fire-and-forget. Do **not** merge yet.

### DIFF
--- a/src/tools/terminal.ts
+++ b/src/tools/terminal.ts
@@ -862,7 +862,7 @@ export const terminalReadHandler = async ({
 export const terminalSendHandler = async ({
   windowTitle, input, method: inputMethod = "auto", chunkSize = 100,
   pressEnter, focusFirst, restoreFocus, preferClipboard, pasteKey,
-  forceFocus: forceFocusArg, trackFocus, settleMs,
+  forceFocus: forceFocusArg, trackFocus, settleMs, notifyDispatch = true,
 }: {
   windowTitle: string;
   input: string;
@@ -876,6 +876,11 @@ export const terminalSendHandler = async ({
   forceFocus?: boolean;
   trackFocus: boolean;
   settleMs: number;
+  /** ADR-014 v2 R3 L3-4 S-A: fire the dispatch hook on successful delivery.
+   *  terminalRunHandler passes false when it delegates here — it owns the
+   *  single notification (with the user's ORIGINAL input) so this nested send
+   *  must not double-fire. Defaults true for direct action='send'. */
+  notifyDispatch?: boolean;
 }): Promise<ToolResult> => {
   const force = forceFocusArg ?? (process.env.DESKTOP_TOUCH_FORCE_FOCUS === "1");
   const startedAt = Date.now();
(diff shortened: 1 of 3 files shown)


Harusame64:
@codex review

Focus: the two `fireTerminalDispatch` call sites — are they after the resolved-window guard (never fire for an unresolved window), exactly once per send/run, and do they pass the USER's original `input` (not the buildExitCommand exit-mode rewrite in the run path)? Also confirm the hook is truly fire-and-forget (a throwing hook cannot break a dispatch) and additive (zero behaviour change when no hook is set).

Harusame64:
@codex review

Re-review at new HEAD. The only change since the last pass (`113e4afc`) is a build-guard fix: regenerated tests/fixtures/failwith-callsite-shapes.json for the line drift the dispatch hook introduced (line numbers only — tool / errKind / contextShape unchanged). No production-code change in this delta.

Focus (unchanged from prior request): the two `fireTerminalDispatch` call sites — are they after the state mutation / on every early-return path?

Harusame64:
@codex review

Round 2 applied your 2 P1 findings on the S-A dispatch hook:
- Double-fire: run now passes `notifyDispatch:false` when delegating to send and is the single notifier for its path.
- Premature fire: send fires via `markDispatched()` only at the 5 successful-delivery returns; run fires once after delivery is confirmed (`sendPayload?.ok === true`) with the USER's original input (never the exit-mode epilogue).

Please confirm no path double-fires and none fires before bytes are delivered.

Harusame64:
@codex review

Re-review at HEAD aedc0813 (final). Round 2 fixed both P1s on the S-A dispatch hook (single fire after delivery + notifyDispatch suppression on run→send). Please confirm no P1/P2 remain.

Harusame64:
@codex review

Round 3 (`60a9ce18`) addresses your P2 on `fireTerminalDispatch`: the fire-and-forget hook is typed `=> void` but TypeScript accepts an async function, so an async rejection escaped the synchronous try/catch. The fix now detects a thenable return value and attaches a rejection handler; a regression test registers an async observer whose returned promise rejects and asserts no `unhandledRejection` fires. Please re-review the current HEAD.

PR #512, opened 2026-07-06
ADR-014 v2 R3 L3-4 W-2b: tick argv-classified unregistered-ssh fail-safe

Harusame64:
## What

Closes a **credential-disclosure** hole Opus found while reviewing the L3-4 plan (R2 Q1 / R3 Q1a): in the cooperative model a user can hand-`ssh host` out of the SAME local pane the assistant launched. The wiring never dispatched it, so no session frame is pushed and no pid is registered — the tracker stays confidently **local**, and a later assistant `sudo` would fill a **LOCAL secret into the REMOTE prompt**.

## How

`SshSessionWatch.tick()`, for a **LOCAL-anchored** pane (`session===null` AND `remoteDepth===0`), now walks the shell's process **subtree** (`parentMap`) and classifies each live `ssh`-named **descendant**'s argv (new `commandLine(pid)` seam on `ProcessSnapshot`, fed in production by W-0 `getProcessCommandLineByPid`) with the **existing** `interactiveSshTarget` rule:

- interactive in-bound login (`ssh user@host`) → `markUnknown` (decline, never disclose)
- argv **unreadable** (`null` — ACCESS_DENIED on an elevated/cross-user ssh) → `markUnknown` (fail-safe)
- **tunnel** (`-N`/`-f`/`-L`) / one-shot (`ssh host cmd`) / `scp` / `sftp` / git-over-ssh → **NOT** flagged, so those panes keep autofilling (**OQ-W-7 = option B**, the chosen behaviour — decline only real in-bound logins, not every ssh-named descendant)

Design points:
- **Descendant walk** (not direct-children only) closes the `sudo ssh host` grandchild case (Opus R3 Q1a).
- **Gated on `remoteDepth===0`**: a legit assistant-dispatched interactive ssh has already pushed a frame (`depth>0`) and is handled by the existing unwatched-frame backstop — so this scan never fights the normal push→register flow. That's the key correctness point.
- Strictly a **new fail-safe** (`markUnknown` on doubt), never a new pop. Short-circuits on the first interactive/unreadable ssh.

## Tests
35 watch tests (9 new W-2b: interactive direct-child / `sudo ssh` grandchild / tunnel / one-shot / scp-inner-ssh / unreadable-argv / non-ssh-only / depth>0-gating / registered-session-not-rescanned) + full 317 key-locker suite green; `tsc` clean. One existing test (`bad-pid-on-local-pane`) updated to give its incidental ssh child an explicit **tunnel** argv — legitimately local under the new semantics; the assertion is unchanged.

## Review focus
The `remoteDepth===0` gating (does it ever let a legit dispatched ssh get double-markUnknown'd, or let an in-bound ssh through?), the subtree-walk cycle guard, and whether `interactiveSshTarget` over a raw process argv classifies every disclosure case correctly (esp. `scp`/git-over-ssh inner ssh, ProxyJump).

Plan: `desktop-touch-mcp-internal@afcd0bb:docs/adr-014-v2-r3-l3-4-live-wiring-plan.md` (W-2b). Depends conceptually on #510 (W-0) for the production `commandLine` adapter (wired in W-3); this PR uses only the seam interface + fakes.

### DIFF
--- a/src/engine/key-locker/ssh-session-watch.ts
+++ b/src/engine/key-locker/ssh-session-watch.ts
@@ -310,8 +310,10 @@ export class SshSessionWatch {
    * `shellPid` and, for each LIVE `ssh`-named DESCENDANT (subtree, not just direct — a `sudo ssh`/wrapper/
    * tmux-reparented login is a grandchild), classify its argv with `interactiveSshTarget` (the SAME rule that
    * decides whether an ssh opens a session). Returns true (⇒ caller `markUnknown`s, decline-not-disclose) iff
-   * any descendant is an interactive in-bound login OR its argv is UNREADABLE (`commandLine` null — an
-   * elevated/cross-user ssh we cannot introspect: fail-safe to "possibly interactive"). A tunnel (`-N`/`-f`/
+   * any LIVE descendant is itself UNREADABLE (identify() name "" — an elevated/cross-user process we cannot
+   * even name, e.g. the ssh in `sudo ssh`; Codex PR#512 P1), OR is an interactive in-bound `ssh` login, OR is
+   * an `ssh` whose argv is UNREADABLE (`commandLine` null — an elevated/cross-user ssh we cannot introspect:
+   * fail-safe to "possibly interactive"). A tunnel (`-N`/`-f`/
    * `-L`), a one-shot (`ssh host cmd`), or scp/sftp/git-over-ssh (whose inner ssh carries a trailing remote
    * command) all classify null ⇒ NOT flagged, so those panes keep autofilling (OQ-W-7 = option B). Short-
    * circuits on the first interactive/unreadable ssh, so the common no-ssh subtree is one adjacency walk.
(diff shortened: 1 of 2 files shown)


Harusame64:
@codex review

Focus: the `remoteDepth===0` gating in the session===null branch (can a legit dispatched ssh ever get double-markUnknown'd, or can an in-bound interactive ssh slip through?); the subtree BFS cycle guard (`seen`); and whether `interactiveSshTarget` over a RAW process argv (from CommandLineToArgvW) correctly classifies scp/sftp/git-over-ssh inner ssh and ProxyJump as non-interactive while catching `ssh user@host` / `sudo ssh`.

Harusame64:
@codex review

Re-review at new HEAD (`54b1cbb8`). Since the last pass (`3c4c849`) this branch applied Round 2 (Opus P2+P3): empty-argv fail-safe + per-tick children map + coverage.

Focus: the `remoteDepth===0` gating in the `session===null` branch, and the empty-argv fail-safe — does an empty/unclassifiable argv correctly fall to decline (disclosure guard) rather than trusted-continue?

Harusame64:
@codex review

Round 3 applied your P1 in the W-2b subtree scan: an empty identify() name at that point is a LIVE-but-UNREADABLE descendant (every BFS pid is a parentMap KEY = alive), so it now fails closed (return true ⇒ markUnknown) before the name!=='ssh' skip — a `sudo ssh` with an elevated ssh no longer leaves the depth-0 pane trusted-local.

Please confirm the fail-closed gate and that a readable non-ssh descendant still skips.

Harusame64:
@codex review

Re-review at HEAD 62d39f12 (final). Round 3 fixed the P1 by failing closed on an unreadable live ssh descendant. Please confirm no P1/P2 remain.

PR #513, opened 2026-07-07
ADR-014 v2 R3 Key Locker — L3-4 W-2: capture driver (RECONCILE-then-FREEZE)

Harusame64:
## What this is

The **live-wiring capture driver** (L3-4 slice W-2, THE CRUX) — the first production owner that turns the merged, unit-tested, *pure* L3 pipeline into a running "the locker learns and fills credentials" loop, while honoring the three atomicity obligations (`W1`/`W2`/`W3`) the `ssh-session-watch.ts` header pinned as **wiring** responsibilities that a pure module cannot self-close.

No public tool change — the autofill loop is internal (`key_locker` management surface already shipped in L4).

## Files

- `src/engine/key-locker/key-locker-capture-driver.ts` — engine, **pure over injected seams** (no Win32 / host / terminal import). Owns the live `SessionTracker` + `SshSessionWatch` (constructed by the manager, W-3) so tests drive the REAL reconciliation against a fake process tree.
- `tests/unit/key-locker-capture-driver.test.ts` — 20 tests.

## The four events

| Event | Contract |
|---|---|
| `onLocalPaneLaunched` | **W1**: `beginLocalSession` + `watchPane` in ONE turn — **LAUNCHED panes only**, so a pre-existing pane stays UNKNOWN → declines (**P1-B**; anchoring on "no ssh child seen" would false-anchor a plink/mosh/renamed-ssh pane local → disclose a LOCAL secret to a REMOTE prompt). |
| `onDispatch` | **S-A / W3 RECONCILE-then-FREEZE**: `watch.tick()` reconcile → freeze `tracker.get` → `recordDispatch` (derive-then-record) → arm the poller for a credential command in a known session (OQ-W-3). Reconcile + freeze + record are **synchronous** (no `await` between) so an interleaved async tick can't reorder them; only the pure, idempotent arm-derive is awaited AFTER. **Single-flight (P2-E)** also drops the re-entrant onDispatch the Mode-A landed re-run fires (**W4-O2**). |
| `poll` | **§3.1 before/after ssh-child DIFF**: register the NEW interactive child, not a pre-existing tunnel (**Q2**); `>1` interactive OR argv-unreadable ⇒ `markUnknown`, never guess (**R3 Q3**). Then `runCaptureLoop` bound to the **FROZEN** session on a credential prompt. |
| `onPaneClosed` | **W2**: `unwatchPane` + `tracker.forget` same-turn. |

## Why RECONCILE-then-FREEZE (not freeze alone)

The tracker is mutated not only by dispatches but by `tick()` reacting to an EXTERNAL ssh child exiting. Counterexample freeze-alone misses: `ssh host-a` → the ssh exits (pane now LOCAL) → before the next tick, `sudo x` dispatched → a freeze of `tracker.get()` captures the stale `{host-a, remote}` → autofills host-a's secret into a LOCAL prompt (L2 re-verify can't catch it — window/title unchanged). So `onDispatch` reconciles FIRST (pops the dead ssh), then freezes → correct-or-UNKNOWN.

## Tests (20, real tracker + watch, fake tree)

W1 anchor / P1-B decline / OQ-W-3 arm filter / derive-then-record (`ssh b@host-b` derives host-a) / **P1-A** reconcile-pops-dead-ssh-so-`sudo x`-freezes-localhost / frozen-survives-interleaved-tick / **single-flight** / **§3.1 DIFF** (tunnel-sibling registered-correctly, argv-unreadable ⇒ sink, >1 ⇒ sink) / every `CaptureLoopOutcome` pass-through / poller timeout / **W2** close.

## Verification

```
npm run build && npm run lint            # clean
npx vitest run --project unit tests/unit/key-locker-*.test.ts   # 341 passed
npm run check:stub-catalog               # unchanged (no public tool)
npm run check:failwith-fixtures          # unchanged
```

Next: **W-3** (manager Win32 snapshot adapter + tick timer) → **W-4** (impure wiring root binding the terminal hooks) → live dogfood → R3 release.

Plan: `desktop-touch-mcp-internal@b69c582:docs/adr-014-v2-r3-l3-4-live-wiring-plan.md` (§2.2, §3, §4 W-2)

### DIFF
--- a/src/engine/key-locker/key-locker-capture-driver.ts
+++ b/src/engine/key-locker/key-locker-capture-driver.ts
@@ -253,12 +253,15 @@ export class KeyLockerCaptureDriver {
     this.tracker.recordDispatch(paneId, command);
     if (this.tracker.remoteDepth(paneId) > depthBefore) rec.pendingSsh = this.snapshotSshBaseline(rec.shellPid);
 
-    // (3) ARM the poller iff this is a credential command in a KNOWN session (OQ-W-3 — don't poll after
-    //     `ls`). The derive is pure/idempotent; the loop re-derives from the SAME frozen frame (a benign
-    //     double-derive). An UNKNOWN pane (reconcile sank it, or a conditional/ambiguous record) stays
-    //     disarmed (rec.armed already null from 0a) — but `pendingSsh` (session tracking) is INDEPENDENT of
-    //     arming, so it stays. Both pre-await returns leave the null from 0a; nothing interleaved (sync).
-    if (!isKnownSession(frozen)) return;
+    // (3) ARM the poller iff this is a credential command in a session that is KNOWN both BEFORE the record
+    //     (the derive-then-record context) AND AFTER it. `recordDispatch` can SINK the pane to UNKNOWN for a
+    //     conditional/ambiguous command — `false && ssh host ; sudo -v`, `cd x && ssh host` — whose effect it
+    //     cannot predict; the pre-record `frozen` still looks local, so without the post-record check the
+    //     trailing `sudo` prompt would fill under the (mis-derived) skipped-ssh binding (Codex L3-4-W2 R5 P1).
+    //     Declining a sunk pane is the fail-safe (never wrong-target). The derive is pure/idempotent; the loop
+    //     re-derives from the SAME frozen frame (a benign double-derive). `pendingSsh` (session tracking) is
+    //     INDEPENDENT of arming, so it stays. Both pre-await returns leave the null from 0a (nothing async yet).
+    if (!isKnownSession(frozen) || !isKnownSession(this.tracker.get(paneId))) return;
     const binding = await this.deps.deriveBinding(command, frozen);
     // Publish ONLY if this dispatch is still the latest for the pane — a newer onDispatch (which bumped
     // `dispatchSeq` and cleared `rec.armed`) owns the arm now; an out-of-order older derive must not clobber
--- a/src/engine/key-locker/key-locker-capture-driver.ts
+++ b/src/engine/key-locker/key-locker-capture-driver.ts
@@ -167,12 +167,11 @@ interface PaneRecord {
   /** A `poll()` is in progress for this pane — serializes polls (a re-entrant poll returns `busy`) so a
    *  slow prompt read cannot let two polls run two capture loops for one dispatch (Codex L3-4-W2 R2 P2). */
   pollBusy: boolean;
-  /** The CAPTURE LOOP (prompt found → fill → landed → save) is running — a SUBSET of `pollBusy`. onDispatch
-   *  drops ONLY a re-dispatch of `loopCommand` (the Mode-A landed re-run, P2-E / W4-O2), not a real dispatch. */
-  loopInFlight: boolean;
-  /** The command the in-flight capture loop is processing (null when no loop) — lets onDispatch tell the
-   *  Mode-A re-run (same command) from a REAL post-auth dispatch (different command) that must be processed. */
-  loopCommand: string | null;
+  /** True ONLY while the Mode-A `runToExit` landed seam is executing — the exact window that seam re-fires
+   *  onDispatch for the same command (W4-O2). onDispatch drops a re-entrant ONLY on this flag (NOT a command
+   *  string, which is fragile to hook-emission normalization — Opus R7 P2). Mode B never enters runToExit, so
+   *  a genuine post-auth dispatch during readPaneAfterAuth/offerSave is admitted (recorded/reconciled). */
+  inRunToExit: boolean;
 }
 
 const SSH_PROGRAM = "ssh";
@@ -199,7 +198,7 @@ export class KeyLockerCaptureDriver {
   onLocalPaneLaunched(paneId: string, shellPid: number, cwd?: string): void {
     this.tracker.beginLocalSession(paneId, cwd);
     this.watch.watchPane(paneId, shellPid);
-    this.panes.set(paneId, { shellPid, armed: null, dispatchSeq: 0, pendingSsh: undefined, pollBusy: false, loopInFlight: false, loopCommand: null });
+    this.panes.set(paneId, { shellPid, armed: null, dispatchSeq: 0, pendingSsh: undefined, pollBusy: false, inRunToExit: false });
   }
 
   /**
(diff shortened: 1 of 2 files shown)


Harusame64:
@codex review

Harusame64:
Thanks — real window. Resolved in `a488cf41`, but **not** by deferring the push.

**Why the push stays at dispatch (deferring it is unsafe).** `recordDispatch`'s push-at-dispatch is a deliberate *fail-safe*: an unwatched remote frame (`session===null && remoteDepth>0`) is markUnknown'd → the pane **declines**, never wrong-targets. If we deferred the push until the child is correlated, a *missed* correlation (Toolhelp miss / the child reparents) would leave the tracker **under**-reporting remote → a later command derives `sudo://localhost` for a pane that is actually remote → a **LOCAL secret typed into a REMOTE prompt** (disclosure). So deferral trades a safe decline for an unsafe disclosure.

**What closes the window instead.** `watch.tick()` is driven **only** from the driver (`onDispatch` step 0 + `tickWatch`). So the driver now `correlateAllPending()` — registers every armed pane's freshly-spawned interactive-ssh child via `noteSshOpened` — **before every tick it issues**. A just-pushed frame whose child is already visible is registered before the backstop can observe it, so a normal ssh login keeps its tracking. The only residual is the sub-ms gap before the child appears in Toolhelp (**OQ-W-9**, already an accepted P3) — and that path is still fail-safe (declines, never wrong-target).

3 new tests pin it: a periodic `tickWatch` keeps a correlated frame; another pane's `onDispatch` tick keeps it; a child-not-yet-visible tick markUnknowns (the honest fail-safe boundary).

@codex review

Harusame64:
Good catch — fixed in `0fa9bbdc`.

The single-flight flag (`loopInFlight`) was set only *after* the `readPromptTail` await, so a poll-timer re-entry during a slow UIA prompt read passed the guard. Added a per-pane **`pollBusy`** flag set *before* the await (cleared in a `finally`), so a re-entrant `poll()` returns `busy`. `loopInFlight` (the capture-loop window) is now a strict subset of `pollBusy` and stays — `onDispatch` checks it to drop the Mode-A landed re-run (W4-O2). New test: two concurrent polls with a blocked prompt read ⇒ the second is dropped, `readPromptTail` is called once, exactly one capture loop runs.

@codex review

Harusame64:
Both real — fixed in `22d25b84`.

**P1 (stale arm):** after the `readPromptTail` await, `poll()` now checks `rec.armed !== armed` (the arm captured before the await). If a newer dispatch overwrote it mid-read, the poll aborts with a new `superseded` status and runs no loop — so a newer prompt is never filled/saved under the older binding. The newer arm is left intact and polled next tick.

**P2 (correlation published too late):** the interactive-ssh correlation was stored inside `rec.armed`, set only *after* the `deriveBinding` await, so `correlateAllPending` couldn't see it during a slow `ssh -G`. Decoupled it to `rec.pendingSsh`, **published the instant `recordDispatch` pushes the frame** (before the arm-derive await). Bonus correctness: `pendingSsh` now also survives a not-armed ssh (binding `null`), so the session-end watch still tracks it.

2 new tests: `pendingSsh` visible to `tickWatch` during a blocked `deriveBinding` (frame kept); a poll whose arm was overwritten mid-read returns `superseded`, runs no loop, leaves the newer arm.

@codex review

Harusame64:
Real — the last onDispatch-derive-window hole. Fixed in `af0bc702`.

The previous `rec.armed` was left in place across a *newer* dispatch's `deriveBinding` await, so a poll could see the new command's ssh prompt while `rec.armed` still held the OLD cached-sudo arm — the R4 `rec.armed !== armed` guard doesn't fire (the arm is unchanged), so the ssh prompt fills under sudo.

Fix (completes the single-slot latest-wins arm discipline):
- `onDispatch` now **clears `rec.armed` at its start** (before the derive await), so no stale arm is pollable while a newer command derives.
- a per-pane monotonic **`dispatchSeq`** token gates the post-derive publish: only the LATEST dispatch arms, so two overlapping derives resolving out of order can't clobber the newer arm.

This closes the whole onDispatch-derive-window class — R2's push/register race, R3's late correlation, and R4's stale arm all stemmed from that one await: the arm is cleared + token-gated, `pendingSsh` is published pre-await, and `poll` re-validates the arm after its own read. (A per-pane async mutex was ruled out: the Mode-A landed re-run fires `onDispatch` from *within* the in-flight loop, so a queue would deadlock — drop/abort flags are the only deadlock-free shape.)

2 tests: a cached-sudo arm cleared the instant a newer ssh dispatches; an out-of-order older derive doesn't clobber the newer arm.

@codex review

Harusame64:
Real — a different class from the await-window wave (this one is record-then-arm). Fixed in `5f71b9f3`.

`recordDispatch` SINKS the pane to UNKNOWN for a conditional/ambiguous command (`false && ssh host ; sudo -v`) because it can't predict whether the conditional ssh runs — but the arm decision used only the PRE-record `frozen` (still local), so the trailing sudo prompt could fill under the mis-derived skipped-ssh binding.

Fix: arm only if the session is known **both** before the record (the derive-then-record context, unchanged) **and after** it — decline when `recordDispatch` markUnknown'd the pane. Fail-safe (declines, never wrong-target). The legitimate nested `ssh b@host-b` still arms (recordDispatch PUSHES a known host-b frame; the derive stays from the pre-push frozen host-a).

Test: `false && ssh deploy@host-a ; sudo -v` ⇒ pane UNKNOWN ⇒ not armed ⇒ the sudo prompt declines, capture never called.

Separately: the independent reviewer flagged that this single `onDispatch` `await` has now produced several distinct re-entrancy bugs, and recommended a structural fold (move `deriveBinding` out of `onDispatch`, arm on a cheap synchronous pre-filter, make `onDispatch` run-to-completion atomic). I've filed that as a tracked follow-up PR (it changes the arm-filter and needs its own over-arm tests) rather than expanding this one — the current targeted discipline is complete and well-tested for W-2.

@codex review

Harusame64:
Real — a Mode-B lifecycle hole, distinct from the await-window and record-then-arm classes. Fixed in `1e74c3ba`.

The `loopInFlight` guard dropped EVERY onDispatch during a capture loop, but a Mode-B login succeeds mid-loop — during `readPaneAfterAuth`/`offerSave` the shell is back at an interactive prompt — so a genuine `ssh host-b` there was dropped instead of recorded, leaving the tracker on host-a → a later host-b `sudo` filled the host-a binding.

Fix: onDispatch drops ONLY the Mode-A landed re-run (the SAME command re-fired by `runToExit`, W4-O2; W4-O1 guarantees the hook emits the user input = `loopCommand`), NOT a real dispatch (a different command), which is recorded/reconciled normally. The loop stays insulated (local captured `armed` + immutable `frozen`), and its `finally` now disarms ONLY its own arm (`rec.armed === armed`), so a real dispatch that armed a newer command mid-loop survives. A same-command real re-dispatch (rare) is bounded-safe to drop (nested `ssh host-a` from host-a still resolves to host-a).

Tests: a real nested `ssh host-b` during a Mode-B post-auth window is recorded (tracker → host-b) and its arm survives; the Mode-A same-command re-run is still dropped.

@codex review

Harusame64:
Applied the more robust fix for the re-run drop (independent-reviewer recommendation), in `9c0291a6` — superseding the R6 string-match.

The R6 same-command drop was a string match (fragile to any hook-emission normalization → a duplicate loop, though never a wrong-target). The driver now **brackets its own `runToExit` seam with a per-pane `inRunToExit` flag**, and onDispatch drops a re-entrant ONLY on that flag — the exact window the Mode-A re-run fires. Mode B never enters `runToExit`, so the flag stays false through `readPaneAfterAuth`/`offerSave`, cleanly admitting real post-auth dispatches. Net simplification: `loopInFlight` + `loopCommand` removed for one `inRunToExit`.

@codex review

Harusame64:
## On hold — foundational timing correction pending

Both R8 P1 findings are **valid and confirmed**, and together they surface a foundational issue: this driver was designed assuming the S-A dispatch hook fires **before** delivery (plan W4-O1), but the merged W-1 (#511) fires `fireTerminalDispatch` **after successful delivery** (verified: `terminal.ts:903-913` send, `terminal.ts:2137-2145` run). That breaks three things the unit tests (which model fire-before) don't catch:

1. **P1 (line 251)** — the pre-freeze W-2b scan runs before `recordDispatch`, so a fast just-dispatched `ssh` child is already visible and gets classified as an *unregistered user ssh* → the pane is markUnknown'd → the assistant's own login prompt is never armed/filled and the session watch is lost.
2. The §3.1 correlation baseline is snapshotted *after* the child has already spawned, so the before/after diff can never register it.
3. **P1 (line 228)** — the loop admits real dispatches during the *pre-auth* capture/inject window (shell blocked at the password prompt), recording commands ahead of what actually ran → a later fill can stale/wrong-target if the original login is rejected.

Rather than patch these on the wrong foundation (this is round 8), the plan's timing model is being corrected first (fire-after-delivery), reviewed, and the driver redesigned against it. This PR is marked **draft / on hold** and kept as reference for the security-converged parts (reconcile-then-freeze + post-record gate + the wrong-target invariants) and the test suite.

Thanks — the multi-round adversarial pass caught a real design-assumption error the unit tests structurally couldn't.

Harusame64:
Superseded by #515 (W-2 REDO — the synchronous fire-after-delivery capture driver). This draft was built on the fire-BEFORE-delivery assumption that Codex's 8-round review surfaced as wrong (the S-A hook fires only after a confirmed delivery). The redesign per plan §0-CORR shipped in #515. Closing.

PR #514, opened 2026-07-07
ADR-014 v2 R3 Key Locker — L3-4 W-0.5: SshSessionWatch surgical W-2b exempt (fire-after-delivery prep)

Harusame64:
## What this is

**W-0.5** — a small, surgical prep change to the 6-round-hardened `ssh-session-watch.ts`, ahead of the fire-AFTER-delivery driver redesign (plan §0-CORR).

## Why

The merged W-1 (#511) fires the S-A dispatch hook **only after a confirmed delivery** (`terminal.ts:903-913` / `2137-2145`). So an `ssh host-a` the **assistant** dispatched is already in the process tree when the driver drives its reconcile tick — and the W-2b unregistered-ssh scan would mis-flag the assistant's **own** login as a lurking *user* ssh and `markUnknown` the pane (this is Codex PR #513 R8 P1, which surfaced that the whole W-2 driver assumed *fire-before-delivery*).

## The change

`tick(exempt?: { paneId, pid, startTimeMs })` — **conditional + surgical** (per the Opus plan review, then hardened to a PID identity over two Codex rounds — see below):
- The driver passes `exempt` **only when the dispatched command itself opens an interactive ssh** and it has **proven** (via a before/after ssh-descendant delta) the exact child pid THIS dispatch spawned; a `sudo`/non-ssh dispatch, the periodic timer, and an ambiguous/unproven delta pass nothing → **full scan everywhere**.
- When present, inside the W-2b scan for `exempt.paneId`, a descendant is skipped **only if its pid === `exempt.pid` AND its creation time === `exempt.startTimeMs`** (and a zero creation time never matches — the module's doubt sentinel). Any **other** interactive ssh — including a real user `ssh` to the **same host** (a **different pid**) — **still flags**; an **unreadable** ssh descendant (name `""` / argv `null`) **still flags** (fail-safe); the exempt pid's **own subtree is still walked** (a lurking child under it still flags); every **other pane** scans in full.
- The exempt is a **strict narrowing** of the opportunistic W-2b scan — it never touches the hardened wrong-target pop / markUnknown core.

## Why identity is pid+creation-time, not host (Codex W-0.5 P1 → P2)

- **P1 (host is not identity):** an earlier draft matched on the argv *host*. Codex flagged that a user's own `ssh host-a` to the **same host** the assistant just dispatched would inherit the exemption → a lurking user login goes unflagged → disclosure. Fixed by matching the **proven pid** the driver correlated, not the host.
- **P2 (pid reuse):** a bare pid match can be fooled if the dispatched ssh exits and Windows reassigns its pid to a user's ssh before the snapshot. Fixed by also matching the **creation time** (mirrors this module's existing `session.startedAt` reuse discipline) — a reused pid reads a different creation time ⇒ NOT exempt ⇒ still flags.

## Why it's safe (Opus plan-review 2 rounds + HEAD review, CONVERGED)

- The naive **unconditional** exempt was rejected as a **P1 disclosure** (a `sudo` dispatch over a lurking user ssh would skip the scan → local secret into the remote prompt). The conditional gate closes it.
- **Same-host collision** (`ssh host-a` dispatched while a user also has `ssh host-a` open): the two are **distinct pids** — only the proven-pid child is skipped, the user's is a different pid and still flags. (And the driver's `correlateDispatchedSsh` seeing `>1` host-matching child → `markUnknown`, with the post-record-sink gate re-reading `isKnownSession` **after** that, is a second layer.) No gap.

## Tests (8 exempt-specific; 47 ssh-watch total)

proven-pid exempt not flagged / same-host DIFFERENT-pid still flags (P1) / different-host still flags under exempt / unreadable descendant still flags (fail-safe) / exempt scoped to its named pane only / lurking child UNDER the exempt pid still flags (subtree walked) / REUSED exempt pid — same pid, different creation time — still flags (P2) / no-exempt full scan (backward-compat).

## Verification

```
npm run build && npm run lint            # clean
npx vitest run --project unit tests/unit/key-locker-*.test.ts   # green
```

No public tool change. The redesigned synchronous driver (W-2 REDO) consumes this exempt.

Plan: `desktop-touch-mcp-internal@fad05e8:docs/adr-014-v2-r3-l3-4-live-wiring-plan.md` (§0-CORR.2, §0-CORR.5 W-0.5)

### DIFF
--- a/src/engine/key-locker/ssh-session-watch.ts
+++ b/src/engine/key-locker/ssh-session-watch.ts
@@ -233,13 +233,16 @@ export class SshSessionWatch {
    * `exempt` (L3-4 §0-CORR.2 — the fire-AFTER-delivery correction): when the driver drives this tick from
    * `onDispatch` for a command that ITSELF opens an interactive ssh, that ssh child is ALREADY in the tree
    * (the S-A hook fires post-delivery), so the W-2b unregistered-ssh scan would mis-flag the assistant's OWN
-   * just-dispatched login as a lurking USER ssh and `markUnknown` the pane. `exempt = { paneId, host }`
-   * SURGICALLY removes ONLY the `host`-matching interactive descendant of `exempt.paneId` from the W-2b scan —
-   * every OTHER unregistered interactive ssh (a real user `ssh host-evil` on the same shared pane), and every
-   * other pane, is scanned in full. Omit `exempt` (a `sudo`/non-ssh dispatch, or the periodic timer) ⇒ the
-   * full scan runs everywhere. The exempt NEVER touches the wrong-target pop/markUnknown core.
+   * just-dispatched login as a lurking USER ssh and `markUnknown` the pane. `exempt = { paneId, pid }` skips
+   * EXACTLY that one process pid in the W-2b scan for `exempt.paneId` — the driver PROVED (via a before/after
+   * ssh-descendant delta) it is the child THIS dispatch spawned. Every OTHER interactive-ssh descendant — a
+   * user's ssh to the SAME host is a DIFFERENT pid — STILL flags; every other pane scans in full. A PID (not a
+   * host) is REQUIRED: a host match cannot prove the skipped process is the newly-dispatched one, so a user's
+   * pre-existing `ssh host-a` would be wrongly skipped and leak local→remote (Codex L3-4 W-0.5 P1). Omit
+   * `exempt` (a `sudo`/non-ssh dispatch, the periodic timer, or an ambiguous/unproven delta) ⇒ the full scan
+   * runs everywhere. The exempt NEVER touches the wrong-target pop/markUnknown core.
    */
-  tick(exempt?: { paneId: string; host: string }): void {
+  tick(exempt?: { paneId: string; pid: number }): void {
     if (this.panes.size === 0) return;
     const snap = this.deps.snapshot();
     // A degenerate (empty) snapshot means the native call FAILED, not that every process died — skip the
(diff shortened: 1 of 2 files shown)


Harusame64:
@codex review

Harusame64:
Correct — host is not process identity. Fixed to PID-based in `023bcc24`.

`tick(exempt?: { paneId, pid })` now skips **exactly the driver-proven pid** in the W-2b scan; a user's `ssh host-a` with a **different** pid still flags, an unreadable descendant still flags, and every other pane scans in full. The exempt pid's own subtree is still walked.

The driver (W-2) proves the pid via a **before/after ssh-descendant delta** (plan §0-CORR.2): the NEW host-matching ssh since the pane's last reconcile. Your exact counterexample is closed at the source — a **nested** `ssh host-a` with no new local child yields **0** new candidates ⇒ no exempt ⇒ the full scan flags the user's pre-existing `ssh host-a` ⇒ markUnknown ⇒ decline. `>1` new host-matching (assistant + a racing user ssh) ⇒ no exempt ⇒ markUnknown. So it never leans on a downstream check.

Headline new test: "a SAME-HOST user ssh with a DIFFERENT pid STILL flags".

@codex review

Harusame64:
Good catch — pid reuse. Fixed in `566895a5` by matching creation time too, mirroring this module's own `session.startedAt` reuse discipline.

`tick(exempt?: { paneId, pid, startTimeMs })` now skips a descendant only when **both** the pid **and** the creation time match (and a zero creation time never matches — the doubt sentinel this file already uses). A reused pid reads a different creation time ⇒ NOT exempt ⇒ still flags via the existing checks. `identify()` is read once per descendant. New test: a reused exempt pid (exempt time 20 vs the live pid's time 99) flags.

@codex review

PR #515, opened 2026-07-07
ADR-014 v2 R3 Key Locker — L3-4 W-2 REDO: synchronous capture driver (fire-after-delivery)

Harusame64:
## What this is

**W-2 REDO** — the capture DRIVER, the CRUX of the Key Locker live wiring, redesigned for the **fire-AFTER-delivery** model (plan §0-CORR). Supersedes PR #513's fire-before driver (now draft/on-hold): Codex's 8-round review of #513 surfaced that the merged W-1 (#511) fires the S-A dispatch hook **only on a confirmed-successful delivery**, so `onDispatch` reacts to a command that has ALREADY run and whose ssh child can ALREADY be in the process tree — a foundational timing assumption the fire-before design got wrong.

Engine-pure over injected seams (no Win32/host/terminal import), so the whole wrong-target story is unit-tested with a fake process tree. The thin impure composition root (W-4) binds the seams to real hooks + Win32.

## The three fire-after consequences (§0-CORR)

- **(A) `onDispatch` is FULLY SYNCHRONOUS (OQ-W-13 fold).** The old `await deriveBinding` inside the handler produced the entire await-window bug class (#513 R2–R5). With the sync fold the handler runs atomically — no interleaving — so `dispatchSeq` + clear-at-start + the token gate all **collapse and are deleted**. It arms on a cheap `looksLikeCredential` pre-filter (`programOf(first real token of any segment) ∈ {sudo,doas,ssh,su}`, a generous superset); the poll's capture loop runs the authoritative `deriveBinding`. Over-arming just declines `not_a_credential`.
- **(B) The W-2b scan EXEMPTS the assistant's OWN just-dispatched ssh (§0-CORR.2).** With the child already visible, the reconcile's W-2b scan would classify the assistant's own `ssh host-a` as a lurking USER ssh → markUnknown → its login never arms. The driver proves the exact child pid via a per-pane **`sshBaseline` delta** (host-matched to the dispatched host) and passes `watch.tick({ paneId, pid, startTimeMs })` (the W-0.5 exempt, #514) to skip only that one process. A user's **same-host ssh is a DIFFERENT pid ⇒ still flags**; **0 or >1 candidates ⇒ NO exempt** (fail-safe). Host is not identity (Codex W-0.5 P1); pid+creation-time closes reuse (P2).
- **(C) Post-LANDED loop-admit gate (§0-CORR.3, `loopPhase`).** During the pre-landed window the shell is BLOCKED at the credential prompt, so a delivered command has NOT run — recording it would put the tracker ahead of reality. `onDispatch` DROPS on `pre-landed`; the `awaitLanded` wrapper flips to `post-landed` on acceptance → subsequent real dispatches admitted. This REPLACES the `inRunToExit` / command-match gates (both admitted the pre-auth window). The Mode-A `runToExit` landed re-run fires while still pre-landed ⇒ dropped (W4-O2).

## What SURVIVES from #513 (unchanged security spine)

Reconcile-then-freeze (P1-A) + the post-record-sink gate (R5) + the frozen-derive W3 closure + `correlateAllPending` straggler backstop (OQ-W-5) + `pollBusy` single-flight + the `superseded` re-validation + `onPaneClosed` W2 + the wrong-target invariants. **Deleted:** `dispatchSeq`/clear/token, `inRunToExit`, the §3.1 dispatch-time baseline (replaced by the continuous `sshBaseline`).

## Design notes for review

- **`dispatchedInteractiveSshHost` is AVAILABILITY-only, not security.** It host-matches the exempt delta. Any divergence from `recordDispatch`'s ssh detection only biases toward NO exempt → the W-2b scan then flags the assistant's own login → it declines (a bounded availability regression), never a wrong-target. So it mirrors `recordDispatch` without needing the hardened module's full rigor.
- **Nested logins decline by design.** A nested `ssh @host-b` from a host-a session pushes depth-2 → `noteSshOpened` markUnknowns (the inner login runs on the remote host, locally unobservable). The tests pin this (they do NOT expect a trusted host-b frame).

## Tests (35, fire-after)

W1 anchoring (launched-only, P1-B) / arm pre-filter (env-skip, unknown-pane decline, conditional-ssh sink) / **child ALREADY visible at dispatch ⇒ armed + correlated, NOT markUnknown'd** / **exempt is load-bearing (negative demo)** / **user hand-ssh + sudo ⇒ markUnknown (disclosure guard)** / **pre-existing same-host user ssh still flags under an exempt** / straggler correlation + fail-safe residual / argv-unreadable & >1 ⇒ markUnknown / tunnel not mis-registered / derive-then-record (pre-push frame) / reconcile-then-freeze (P1-A) + frozen survives interleaved tick / **loopPhase: pre-landed DROP, post-landed ADMIT, Mode-A re-run dropped** / single-flight + superseded / all CaptureLoopOutcomes / multi-pane isolation / W2 close.

## Verification

```
npm run build && npm run lint                                      # clean
npx vitest run --project unit tests/unit/key-locker-*.test.ts      # 364 passed
check:stub-catalog / check:failwith-fixtures                       # no drift (no new tool / failWith callsite)
```

No public tool change. Next: W-3 (manager tracker/watch ownership + Win32 snapshot adapter) → W-4 (wiring root) → live dogfood → release.

Plan: `desktop-touch-mcp-internal@fad05e8:docs/adr-014-v2-r3-l3-4-live-wiring-plan.md` (§0-CORR, §2.2, §3, §4 W-2)

### DIFF
--- a/src/engine/key-locker/key-locker-capture-driver.ts
+++ b/src/engine/key-locker/key-locker-capture-driver.ts
@@ -1,0 +1,579 @@
+// ADR-014 v2 R3 Key Locker — L3-4 W-2 (REDO): the capture DRIVER (the live-wiring CRUX).
+//
+// Plan: desktop-touch-mcp-internal:docs/adr-014-v2-r3-l3-4-live-wiring-plan.md (§0-CORR, §2.2, §3)
+//
+// Every L3 pipeline module is merged, pure, and unit-tested in ISOLATION, but NOTHING wires them into a
+// running loop. This driver is that first production owner — the orchestration that turns the terminal's
+// dispatch/prompt events into "the locker learns and fills credentials", while honoring the three
+// atomicity obligations the pure `ssh-session-watch.ts` header pinned as WIRING responsibilities (W1/W2/W3).
+//
+// ⚠ FIRE-AFTER-DELIVERY (§0-CORR — SUPERSEDES the fire-before design of PR #513): the merged W-1 (#511)
+// fires the S-A dispatch hook ONLY on a CONFIRMED-successful delivery (`terminal.ts:903-913` send /
+// `:2137-2145` run). So when `onDispatch(paneId, command)` fires, the command has ALREADY been delivered and
+// is running, and a fast interactive-ssh child CAN ALREADY be in the process tree. Three structural
+// consequences vs the old fire-before design:
+//   (A) `onDispatch` is FULLY SYNCHRONOUS (OQ-W-13 fold). The old `await deriveBinding` inside the handler
+//       produced the entire await-window bug class (push→register race, late correlation, stale-arm-during-
+//       derive/read). With the sync fold the handler runs atomically — no interleaving — so `dispatchSeq` +
+//       clear-at-start + the token gate all COLLAPSE and are DELETED. Arm on a CHEAP synchronous pre-filter
+//       (`looksLikeCredential` = `programOf(first real token of any segment) ∈ {sudo,doas,ssh,su}`, a generous
+//       superset) and let the poll's capture loop run the AUTHORITATIVE `deriveBinding` it already runs. A
+//       generous pre-filter only OVER-polls (bounded-safe, OQ-W-3), never wrong-targets.
+//   (B) The W-2b scan must EXEMPT the assistant's OWN just-dispatched ssh (§0-CORR.2): with the child already
+//       visible, the reconcile's W-2b unregistered-ssh scan would classify the assistant's own `ssh host-a`
+//       as a lurking USER ssh → markUnknown → the assistant's own login never arms. The driver proves the
+//       exact child pid THIS dispatch spawned via a before/after `sshBaseline` delta (host-matched to the
+//       dispatched host) and passes `watch.tick({ paneId, pid, startTimeMs })` to skip exactly it. A user's
+//       SAME-host ssh is a DIFFERENT pid ⇒ still flags; ambiguous (>1) or unprovable (0) ⇒ NO exempt.
+//   (C) The loop admits a real `onDispatch` ONLY in its POST-LANDED window (§0-CORR.3, `loopPhase`): during
+//       the pre-landed window the shell is BLOCKED at the credential prompt, so a delivered command has NOT
+//       run — recording it would put the tracker ahead of reality. This REPLACES the old command-match /
+//       `inRunToExit` gates (both admitted the pre-auth window).
+//
+// It stays ENGINE-PURE over injected seams (no Win32 / host / terminal import) so the whole wrong-target
+// story is unit-testable with a fake process tree; the thin IMPURE composition root (W-4,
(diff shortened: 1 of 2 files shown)


Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
## Review status after R5 (`f4a0dc4e`)

Thanks @codex — the incremental sweep across R1–R5 surfaced a genuine disclosure class (a frozen credential binding filled into a pane whose session changed after the freeze) at **every** point the binding is trusted. All are now closed by one structural pattern — **reconcile-then-check the live session against the arm's `expected` frame** — applied at dispatch, at the poll pre-loop, and (the authoritative point) at the **injection instant** inside the `injectPane` wrapper, plus `tickWatch` arm-hygiene.

Two remaining Codex R5 inline comments:

**`:445` (clear stale arms after unknowning panes)** — resolved in R5. `tickWatch` now disarms any pane whose freshly-reconciled live session no longer matches its arm's `expected` (the arm-hygiene loop right below the flagged `watch.tick()`), and the injection-instant guard makes even a lingering arm unable to fill a mismatched pane. This comment is a re-pin of the earlier thread; the fix is in place.

**`:299` (require a fresh proof before exempting same-host ssh)** — this is a real imprecision, but after a full consequence trace it is **same-host self-consistent, not a cross-host disclosure** (recorded as plan OQ-W-14, deferred P3):

1. The exempt is **host-keyed** to the dispatched host `dHost` — a candidate pid is exempted only when its argv host === `dHost`. So a mis-exempt happens **only when the user's pre-existing ssh and the assistant's dispatch target the same host**.
2. A same-host mislabel yields only same-host derived bindings (`sudo://host-a`, `ssh://host-a`) — the secret's target host is correct.
3. A **cross-host** case (user → host-EVIL, assistant → host-a) does not match (`host-EVIL ≠ host-a`) → not exempted → the W-2b scan flags it → `markUnknown` → decline.
4. A nested dispatch to a **different** host pushes `remoteDepth ≥ 2`, which `tick` markUnknowns before it can go cross-host.
5. When the mis-registered outer ssh exits, its nested inner login dies with it, so the pop-to-local is correct.
6. The R5 injection-instant reconcile+check re-verifies the live session at the secret-delivery instant, so even a mis-registration cannot deliver a secret into a session-mismatched prompt.

No clean fix distinguishes "assistant's fresh child" from "user's pre-existing same-host ssh" under a stale baseline in the fire-after model (creation-time thresholding is defeated by the process-epoch vs wall-clock gap), and declining every single-candidate exempt would decline the assistant's own login on every ssh (a real availability regression). So it is deferred as a same-host-safe P3 in OQ-W-14 rather than fixed.

40 driver + 369 key-locker unit tests green; build + lint clean.

PR #522, opened 2026-07-08
fix(key-locker): DF-5 console-buffer injector (AttachConsole + WriteConsoleInput)

Harusame64:
## DF-5 (BLOCKER) — SendInput returned `sent=0`; switch to a console-buffer injector

Found in the live dogfood (2026-07-08). The SendInput injector returned `sent=0/N, GetLastError()==0` — a foreground-**ownership** gate (input refused because the window-less helper is not the foreground-input owner; **not** UIPI, which returns the full count with events silently dropped). A window-less helper cannot acquire foreground ownership without a fragile cross-process `AllowSetForegroundWindow` + foreground-steal ladder.

### Fix — Option B (adjudicated by Fable, design-reviewed by Opus)
Replace SendInput with a **console-buffer injector**: `AttachConsole(consolePid)` + `WriteConsoleInputW(CONIN$)` writing `KEY_EVENT` records that the shell cooked-reads (`ReadConsole`). Foreground- and UIPI-immune, self-contained in the helper, and a natural fit since the Key Locker only ever targets a classic conhost (`ConsoleWindowClass`).

The old *"inject only when the target is FOREGROUND"* property is **re-expressed** as *"inject only into the verified `consolePid`'s console buffer"*: the write is **addressed** to that console (not broadcast to whatever is foreground), and a **two-stage identity re-verify** (before attach + after the CONIN$ handle opens) pins that hwnd/pid/class/title still name the same console — so mis-delivery to a wrong window is structurally impossible.

### Key details
- Keystrokes ride in each record's **`UnicodeChar`** (a cooked-read cooks `UnicodeChar`, **not** the SendInput-style `wScan`) with `wRepeatCount=1`; 1 UTF-16 code unit = 1 down/up pair (surrogate pairs handled).
- The 3 std handles `AttachConsole`/`FreeConsole` rebind are **saved/restored** around the bracket; **no diagnostic may write** between `AttachConsole` and the final `FreeConsole` (would leak onto the user's console).
- `not_foreground` is **no longer emitted** (kept in the wire enum for back-compat).
- **No wire-protocol change**; change confined to `tools/key-locker/Injection.cs` + `Program.cs` wiring.

### Testing
- **`-SelfTestInjectConsole`** (new, headless): runs the **real** `Win32Input.ReVerifyAndType` against a self-spawned child console (echo-off cooked-read) and asserts a unicode+surrogate secret round-trips — deterministic, no live ssh. Added as an e2e case.
- Pre-impl spike (`scratchpad`, 4/4 PASS) proved the mechanism + that a `wScan`-only mapping types zero chars.
- All green: `dotnet build`, `-SelfTest` / `-SelfTestL2` / `-SelfTestInjectConsole`, `tsc`, key-locker unit suite (378/378), e2e self-test cases.

### Not in this PR
Phase 6 live dogfood (native OpenSSH cooked-read leg — OQ-DF-6; Git-MSYS ConPTY — OQ-DF-8) needs the real desktop and runs after review.

### Notes
- Stacked on #521 (DF-2/DF-3) so the rebuilt `bin/key-locker.exe` inherits DF-2's `InvariantGlobalization=false`. Merge #521 first.

Design / adjudication / Opus review / spike: `desktop-touch-mcp-internal@main:docs/adr-014-v2-r3-l3-4-live-wiring-followups.md` (DF-5)
Impl plan: `desktop-touch-mcp-internal@main:docs/adr-014-v2-r3-df5-writeconsoleinput-impl-plan.md`

### DIFF
--- a/src/engine/key-locker-host.ts
+++ b/src/engine/key-locker-host.ts
@@ -8,7 +8,7 @@
 // NamedPipeClientStream). What differs from S1 is the CONTROL VERBS and the trust story:
 //
 //   * The locker opens a WPF PasswordBox secure dialog on `capture`, DPAPI-encrypts the secret
-//     at rest, and serves it to consumers (askpass / SendInput — L2) — never over THIS pipe.
+//     at rest, and serves it to consumers (askpass / console-buffer inject — L2) — never over THIS pipe.
 //   * The SECRET NEVER CROSSES THIS PIPE. The wire carries opaque ids + control only; `capture`
 //     replies with {captured, rt} booleans, not the secret. So the pipe's integrity guarantee
 //     (kernel client-verify, below) protects the CONTROL channel — a squatter-server is inert
@@ -105,7 +105,7 @@ export interface LockerReply {
 
 // L2 wire contracts (the locker owns these frame shapes; the injector orchestrator consumes them).
 
-/** The dedicated-console target of a SendInput (`inject`) — §2.1 of the L2 plan. */
+/** The dedicated-console target of a console-buffer injection (`inject`) — §2.1 of the L2 plan. */
 export interface InjectTarget {
   /** The console window HWND (decimal string). */
   hwnd: string;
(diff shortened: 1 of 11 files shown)


Harusame64:
@codex review

Harusame64:
@codex review (updated: Round 1 findings applied — stale-comment sweep + InjectorKind sendinput→console rename)

Harusame64:
@codex review (P2 addressed in b227894: TerminateProcess the -InjectConsoleChild in the finally so failure paths don't orphan a blocked key-locker.exe/conhost)

PR #523, opened 2026-07-08
fix(key-locker): DF-5 console-buffer injector — re-target to main (#522 didn't reach main)

Harusame64:
## Why this exists

**#522 (DF-5) never reached `main`.** It was stacked on #521's branch (`fix/key-locker-invariant-globalization-wpf-crash`) and merged into it at 10:01 — but #521 had already merged to `main` at 09:58, so #522's merge landed in a now-dead branch. `main` today has DF-1/DF-2/DF-3 but **not** DF-5.

This PR brings the **identical, fully-reviewed** DF-5 change onto current `main` and rebuilds `bin/key-locker.exe` against `main`'s csproj (so it keeps DF-2's `InvariantGlobalization=false`).

## What DF-5 fixes

The SendInput injector returned `sent=0/N, GetLastError()==0` on the live desktop — a foreground-**ownership** gate (input refused because the window-less helper is not the foreground-input owner; **not** UIPI). Replace it with a **console-buffer injector**: `AttachConsole(consolePid)` + `WriteConsoleInputW(CONIN$)` writing `KEY_EVENT` records the shell cooked-reads (`ReadConsole`) — foreground- and UIPI-immune, self-contained in the helper.

- *"inject only when FOREGROUND"* → *"inject only into the verified `consolePid`'s console buffer"* (addressed write + two-stage identity re-verify before attach and after CONIN$ open).
- Keystrokes ride in each record's `UnicodeChar` (a cooked-read cooks `UnicodeChar`, not `wScan`), `wRepeatCount=1`; std handles saved/restored around the attach bracket; no diagnostic writes inside it.
- `not_foreground` no longer emitted (kept in the wire enum for back-compat). Internal `InjectorKind` `"sendinput"` → `"console"`.
- New **`-SelfTestInjectConsole`**: headless deterministic proof (no live ssh) via the real `Win32Input.ReVerifyAndType`; terminates the child on failure paths so a failed self-test never orphans a blocked key-locker.exe. Added an e2e case.

## Review provenance (from #522)

- **Opus 2 rounds** → P1/P2 zero (struct ABI, `UnicodeChar` mapping, std-handle save/restore, stage-2 re-verify ordering, zeroize, wire non-impact all verified).
- **Codex 1 round** → one P2 (self-test child could orphan on failure) — applied (`TerminateProcess` in the finally).
- `tsc` clean; key-locker unit **378/378**; `-SelfTest` / `-SelfTestL2` / `-SelfTestInjectConsole` all `{ok:true}`.

## Not in this PR

Phase 6 live dogfood (native OpenSSH cooked-read leg — OQ-DF-6; Git-MSYS ConPTY — OQ-DF-8) needs the real desktop.

### DIFF
--- a/src/engine/key-locker-host.ts
+++ b/src/engine/key-locker-host.ts
@@ -8,7 +8,7 @@
 // NamedPipeClientStream). What differs from S1 is the CONTROL VERBS and the trust story:
 //
 //   * The locker opens a WPF PasswordBox secure dialog on `capture`, DPAPI-encrypts the secret
-//     at rest, and serves it to consumers (askpass / SendInput — L2) — never over THIS pipe.
+//     at rest, and serves it to consumers (askpass / console-buffer inject — L2) — never over THIS pipe.
 //   * The SECRET NEVER CROSSES THIS PIPE. The wire carries opaque ids + control only; `capture`
 //     replies with {captured, rt} booleans, not the secret. So the pipe's integrity guarantee
 //     (kernel client-verify, below) protects the CONTROL channel — a squatter-server is inert
@@ -105,7 +105,7 @@ export interface LockerReply {
 
 // L2 wire contracts (the locker owns these frame shapes; the injector orchestrator consumes them).
 
-/** The dedicated-console target of a SendInput (`inject`) — §2.1 of the L2 plan. */
+/** The dedicated-console target of a console-buffer injection (`inject`) — §2.1 of the L2 plan. */
 export interface InjectTarget {
   /** The console window HWND (decimal string). */
   hwnd: string;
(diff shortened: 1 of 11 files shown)


Harusame64:
@codex review — this is the byte-identical DF-5 change from #522 (Opus 2 rounds + Codex 1 round, all resolved), re-targeted to main because #522 merged into the already-merged #521 branch and never reached main. Rebuilt bin/key-locker.exe against main's csproj.

Harusame64:
@codex review — P2 addressed in ed23bf1: -SelfTestInjectConsole now detects a ConPTY handoff (GetClassName != ConsoleWindowClass) and reports {ok:false,skipped:true} + exit 0; the e2e treats skipped as a pass. Classic-conhost (CI + default desktop) still runs the full assertion.

PR #524, opened 2026-07-08
feat(key-locker): on-demand anchored console — launch_console + terminal paneId (ADR-014 R3 OQ-W-16-bis)

Harusame64:
## What & why

Makes **Key Locker autofill usable in a real environment**. Until now autofill only worked on the single console a dogfood env flag (`DESKTOP_TOUCH_KEY_LOCKER_DOGFOOD=1`) auto-launched at startup — a console window popped up on app start and there was no path for a real user. This adds an **on-demand** anchoring surface: the assistant opens an autofill-capable anchored console when it is about to run a credential command for the user, then drives `ssh`/`sudo` into it.

The anchored pane is a classic `ConsoleWindowClass` console (the only injectable class — Windows Terminal is multiplexed and not injectable) that the human can **also** see and type into (cooperative handoff).

Decision (Fable-adjudicated **Option X**; Opus 3-round plan review to P1/P2-zero): a new `key_locker` action + an optional `terminal` `paneId` target. Rejected Y (`workspace_launch` — would punch a hole in its shell-interpreter blocklist), Z (new tool — catalogue cascade), and implicit auto-anchor (architecturally impossible under the fire-after dispatch model).

## Changes

- **`terminal` read/send gain optional `paneId`** (decimal hwnd). **SEND** binds directly by hwnd (`findTerminalWindowByHwnd`) so it survives the post-login title drift (`user@host: ~`) **and** a same-title sibling. **READ** resolves via the uniqueness-guarded `resolveTitleByHwnd` and **safely declines** on a non-unique title (never a wrong-pane read — a correctness bound, not a disclosure; the internal inject path is unchanged).
- **`key_locker` action `launch_console`** → opens (or reuses) an anchored console, returns `{paneId, windowTitle}`. Consent is an acquire path (like `save`). Routes through the live wiring's `ensureAnchoredConsole` so the pane is actually anchored (`onLocalPaneLaunched`); reuse is hwnd-direct; a live-pane cap (3) declines a `fresh` spray with `KeyLockerConsoleLimit`.
- **Discoverability advisory** — a credential command sent to a non-anchored pane (no `paneId`) while the locker is enabled nudges toward `launch_console`. Wired via a generic `_advisory` hook the wiring fills, so `terminal.ts` / `_advisory.ts` never import the locker.
- **Dogfood env auto-launch** now routes through the same `ensureAnchoredConsole` so its startup pane is tracked + reused (kept as a dogfood convenience; a normal install opens panes on demand).

## Tests

20 new unit tests: `findTerminalWindowByHwnd`, paneId routing + decline paths, `ensureAnchoredConsole` reuse/cap/prune, credential-nudge gating/dedup, advisor registration. Full unit suite green (4856 passed); stub-catalog regenerated; no tool-count change.

Plan: `desktop-touch-mcp-internal@6321fe5:docs/adr-014-v2-r3-oq-w-16-bis-anchoring-plan.md`


Harusame64:
@codex review

PR #525, opened 2026-07-08
fix(key-locker): self-healing launch_console reuse — skip a pane that can no longer arm (OQ-8)

Harusame64:
## What & why

Live dogfood of the new `key_locker launch_console` flow surfaced **OQ-8**: a pane returned by `launch_console`'s DEFAULT reuse could be in a state where it can never autofill. Observed cleanly — a console launched at server startup and reused after sitting idle **arm-missed deterministically**, while a **fresh** pane launched right before use **filled end-to-end**.

Diagnosis (instrumented, since reverted): the arm gate itself is **healthy** — a pane arms and fills fine when its driver record is present, even after minutes idle. The real failure is a pane whose driver record was **torn down while its window stayed alive** (a spurious event-bus `window_disappeared` → `onPaneClosed`, most likely under window churn), or whose session drifted to `UNKNOWN`. Reuse only checked window-liveness, so it handed back a pane that can no longer arm.

Both Fable and Opus design-reviewed the fix approach and independently **rejected a blind "re-anchor on reuse"**: re-anchoring a pane that is mid-remote-session as local would disclose a LOCAL secret to a REMOTE prompt (the P1-B / W-2b hole).

## Fix — self-healing reuse

`ensureAnchoredConsole` reuse now requires a pane be **REUSABLE**:
- window alive (hwnd-direct `findTerminalWindowByHwnd`), AND
- still driver-anchored (`driver.hasPane`), AND
- session `KNOWN` (`isKnownSession(tracker.get(paneId))`).

A non-reusable pane is dropped from tracking and a **fresh, re-anchored pane is launched** — self-healing, robust regardless of what caused the record loss/drift. No blind re-anchor of the stale pane (avoids the wrong-target disclosure).

## Changes
- `driver.hasPane(paneId)` accessor.
- `wiring.isReusable(paneId)` + the reuse gate + cap now counts reusable panes.
- 2 self-heal tests (record torn down / session drifted → relaunches). `key-locker-*` suite: 404 green.
- No wire / tool-surface change (no stub / failwith-fixture drift).

Plan: `desktop-touch-mcp-internal:docs/adr-014-v2-r3-oq-w-16-bis-anchoring-plan.md` (OQ-8)

### DIFF
--- a/src/engine/key-locker/key-locker-capture-driver.ts
+++ b/src/engine/key-locker/key-locker-capture-driver.ts
@@ -410,6 +410,13 @@ export class KeyLockerCaptureDriver {
     }
   }
 
+  /** Is this pane still ANCHORED (a driver record exists)? The wiring's `launch_console` reuse checks this: a
+   *  pane whose window is alive but whose record was torn down (a spurious `window_disappeared` → `onPaneClosed`,
+   *  OQ-8) can never arm, so reuse must NOT return it — it launches a fresh, re-anchored pane instead. */
+  hasPane(paneId: string): boolean {
+    return this.panes.has(paneId);
+  }
+
   /** Panes with a credential dispatch awaiting a prompt (the wiring polls exactly these). */
   armedPaneIds(): string[] {
     const out: string[] = [];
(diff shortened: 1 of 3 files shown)


Harusame64:
@codex review

PR #536, opened 2026-07-16
fix(key-locker): decode UTF-8 program output in launched consoles (CJK mojibake)

Harusame64:
## Summary

Fixes garbled (mojibake) CJK/Japanese output in the console opened by `key_locker({action:'launch_console'})`.

`launchAnchoredConsole` spawns a classic conhost whose output codepage defaults to the system OEM CP (e.g. **932** on a Japanese Windows install). A child tool that writes UTF-8 bytes (a database CLI, `git`, `ssh`) is then mis-decoded as Shift-JIS into the console's UTF-16 buffer, so **both the human's display and `terminal({action:'read'})` return mojibake** — the assistant can't read Japanese/CJK output. This was hit in practice analysing a Japanese-data DB through the locker console.

## Change

Set the console output codepage to UTF-8 at startup:

```
[Console]::OutputEncoding = [System.Text.UTF8Encoding]::new()   # no BOM
```

delivered via `-EncodedCommand <base64-utf16le>` (not `-Command`) to sidestep the node → cmd → `start` three-layer quoting of the `(` `)` `::` tokens. One file changed (`key-locker-manager.ts`); the title-claim nonce is interpolated before encoding, so the claim path is unchanged.

## Why it's safe for credential injection

`[Console]::OutputEncoding` calls **only** `SetConsoleOutputCP` — it does **not** touch the input codepage or the `WriteConsoleInputW`/UTF-16 cooked-read path the L2 secret injector rides on. Deliberately **not** `chcp 65001`, which would also change the input CP (and smuggle a `>` metachar). No change to `Injection.cs`.

## Verification (on-hardware)

- Launched the exact built spawn form → title claim resolves, `OutputCP=65001 BOMlen=0`, a child's UTF-8 Japanese output renders and `terminal(read)` returns it clean.
- A cp932 console with `[Console]::OutputEncoding` alone → `InputCP=932 OutputCP=65001` (input CP untouched), clean output.
- **Injection regression guard**: `key-locker.exe -SelfTestInjectConsole` → `{"ok":true,"skipped":false}` (full console-buffer injection verified, unicode+surrogate secret round-trips).
- Unit: `key-locker-anchoring` 14/14 pass; full suite 4861 pass (the 3 reds are pre-existing/flaky and unrelated — 2 known e2e, 1 unit that passes in isolation).

Plan: `desktop-touch-mcp-internal@4e0158e:docs/adr-014-v2-r3x-slice-plan.md` (slice La-0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [1.12.2] - 2026-07-16 — Fix: garbled (mojibake) CJK / Japanese output in Key Locker consoles
+
+### Fixed
+
+- **The Key Locker console now decodes UTF-8 program output correctly.** On a system whose
+  default console code page is not UTF-8 — for example a Japanese Windows install (code page
+  932) — a command that prints UTF-8 text (a database CLI, `git`, or any modern tool) showed up
+  as garbled characters (mojibake) in the console opened by
+  `key_locker({action:'launch_console'})`, and `terminal({action:'read'})` returned that same
+  garbled text — so the assistant could not read Japanese/CJK output either. The launched console
+  now sets its output encoding to UTF-8 at startup, so UTF-8 output (including CJK / Japanese)
+  both renders on screen and reads back correctly. This changes only how program output is
+  decoded for display; credential autofill and all other behavior are unchanged.
+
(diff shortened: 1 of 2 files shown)


Harusame64:
@codex review

Harusame64:
Added a title-claim hardening commit (498d0dc, Opus R1 P3-1). @codex review

PR #537, opened 2026-07-16
feat(key-locker): shell-anchor creation-time reuse detection (S-pid PR1)

Harusame64:
## What & why

First slice of the **S-pid identity re-base** (the groundwork for Windows Terminal autofill). It closes the shell-anchor **liveness-only residual** the ssh session-end watch has carried since L3-3 (documented at `ssh-session-watch.ts:55`).

Before: the watch anchored a pane on its shell pid and treated the pid merely being *alive* as "same shell". A reused pid (the anchored shell dies, Windows reassigns its pid to an unrelated process) could mask shell death, and the subtree walked from a foreign pid could mis-derive a binding.

After: the watch captures the shell's **process creation time** at `watchPane` (from its own snapshot — the same source `noteSshOpened` already uses for the session ssh, so there is one source of truth and no cross-layer plumbing to drift). `tick` case (a2) declines a pane whose live `shellPid` reads a **different non-zero creation time** (a reused pid) via `markUnknown` + unwatch.

Fail-safe fallbacks (no new spurious declines):
- **0 baseline** — the shell was not snapshot-visible at registration (never happens in production; `launchAnchoredConsole` polls until the shell owns the console window) ⇒ liveness-only, the pre-hardening behavior.
- **transient 0 current read** — a `GetProcessTimes` glitch on a still-live shell ⇒ keep watching (0 is the doubt sentinel); a real reuse re-reads non-zero next tick.

This mirrors the existing `session.startedAt` non-zero reuse discipline, applied to the shell anchor.

## Scope

- No signature changes, no launch-side changes, fully back-compat. The creation-time is sourced internally, so **existing tests are untouched** and every consumer keeps its current call shape.
- The launch-side `PaneAnchor` capture + the L2 `InjectTarget` re-verify (pid+creation-time) land in **PR2**; WT autofill in **PR3**.

## Tests

`tests/unit/key-locker-ssh-session-watch.test.ts` + 4 cases: reused pid (different non-zero time) ⇒ markUnknown+unwatch; same time ⇒ steady state undisturbed; transient 0 read ⇒ keep watching; 0 baseline ⇒ liveness-only. Full key-locker unit suite green (396/396).

## Plan

`desktop-touch-mcp-internal@55d99f0:docs/adr-014-v2-r3x-s-pid-gate.md` (E5) — Opus-GO design gate, OQ-S-3 read-path spike cleared on hardware.

### DIFF
--- a/src/engine/key-locker/ssh-session-watch.ts
+++ b/src/engine/key-locker/ssh-session-watch.ts
@@ -51,10 +51,14 @@
 // `markUnknown` sets the tracker stack to null and does NOT auto-recover on the session's later exit —
 // `recordDispatch` early-returns on a null stack — so an unknown pane re-anchors only when the wiring calls
 // `beginLocalSession` at a fresh local prompt (SP-L3-OQ-8 territory), not via exit observation. Safe
-// (declines), just degraded longer. RESIDUAL (P3, plan §Risks): the shell anchor is liveness-only
-// (`!parentMap.has(shellPid)`) with no startTime check — a reused shell pid can mask shell death; it
-// self-heals (the dead shell's ssh child dies/reparents ⇒ the session check catches it) and is tracked in
-// the plan, not gated on.
+// (declines), just degraded longer. SHELL-ANCHOR REUSE (S-pid gate PR1, formerly the liveness-only P3
+// residual, NOW CLOSED): the shell anchor carries the shell's CREATION TIME (`WatchedPane.shellStartTimeMs`,
+// captured at `watchPane`), so `tick` case (a2) declines a pane whose live `shellPid` reads a DIFFERENT
+// non-zero creation time (a reused pid = the anchored shell died) instead of trusting the foreign process.
+// A 0 baseline (shell not snapshot-visible at registration — never in production) or a transient 0 current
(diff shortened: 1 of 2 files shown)


Harusame64:
@codex review

Harusame64:
**Review round 1 — both reviewers GREEN.**

- **Opus** phase-boundary review: **GO (P1 zero)**. Verified a2 reuse-detection correctness, both 0-fallbacks fail-safe, the `watchPane` registration snapshot read does not disturb the R6 re-anchor decline / W-2b / depth-pivot invariants, and the 47 existing tests stay green because each fake tree holds the shell's creation time constant (a2 is a no-op) — not a 強制命令-4 test rewrite. One P2 was a private-plan SSOT drift (gate E5 sketched a plumbed `shellStartTimeMs` param; PR1 ships a 2-arg signature that reads creation-time from the watch's own snapshot — the single-source-of-truth pattern `noteSshOpened` uses). Resolved by syncing the gate doc (`desktop-touch-mcp-internal@1c2988b`); no code change to this PR.
- **Codex**: clean ("Didn't find any major issues").

No code findings ⇒ no fix commit. Ready for merge review.

PR #538, opened 2026-07-16
feat(key-locker): PaneAnchor spawn capture + pid+creation-time inject re-verify (S-pid PR2)

Harusame64:
## What & why

Second slice of the **S-pid identity re-base** (the groundwork for Windows Terminal autofill): the launch-side `PaneAnchor` spawn capture + the L2 `InjectTarget` **pid+creation-time re-verify**.

Before: the inject target was named by a console window hwnd plus whatever pid owned it — pid liveness alone could not detect a reused pid, and the window was the identity source.

After: target identity is **the shell process the locker itself spawned**, named by `{shellPid, shellStartTimeMs}` (a reused pid reads a different creation time, so the pair is unforgeable). The window is demoted to an *additional* classic-only anchor.

## Scope (gate E1-classic / E3 / E3b / E4 / §4 / §5)

- **Launch (E1)**: `launchAnchoredConsole` freezes the anchor at spawn. The classic spawn command, title claim, and pid derivation are byte-identical; the ONE addition is the creation-time read (0 keeps polling within the existing deadline; a deadline without a non-zero read fails the launch — 0 is the doubt sentinel and can never anchor).
- **Wire (E3)**: `InjectTarget` gains required `shellPid`/`shellStartMs` for both hosts; `hwnd`/`consolePid`/`titleFp` become classic-only. Additive — the L0 reader ignores unknown props.
- **C# re-verify (E3/§4)**: a NEW PRIMARY `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)` + `GetProcessTimes` **exact integer equality** check for both hosts, fail-closed on every failure mode (`target_gone` on open-failure, `target_mismatch` on unreadable/0/different time). Classic ADDITIONALLY keeps the entire previous predicate unchanged, plus the new `ConsolePid == ShellPid` cross-check — strictly strengthened, never relaxed. The wt branch (`Hwnd==0`: no window predicate, `AttachConsole(ShellPid)`) ships now but is unreachable until PR3 constructs a wt anchor. **Local uint-field FILETIME** (the BCL ComTypes FILETIME is signed and would sign-extend ~half of all reads — gate Opus P1-2).
- **Parser (E3b)**: `HandleInject` materializes the new fields via the extracted `ParseInjectTarget`; `-SelfTestL2` now pins the wire round-trip (number + string forms, wt-shape absent-field tolerance) AND the FILETIME sign-extension vector (`(1<<32)|0x80000000 → 644245`, the same value `process.rs filetime_to_ms` yields).
- **Seam (E4)**: driver `PaneRecord` carries the full anchor; `CaptureDriverDeps.injectPane` takes the anchor (gate Opus P2-2 — the wiring never re-derives identity from the paneId string).
- **Self-test (§5)**: `-SelfTestInjectConsole` child now publishes its pid+time beside the hwnd; the parent threads them into the target, so the PRIMARY check + cross-check run end-to-end. Verified locally: `{"ok":true,"skipped":false}`.
- `bin/key-locker.exe` rebuilt in this PR (the W-3.5 lesson).

## Out of scope (PR3)

WT launch (`wt -w 0 new-tab` + WMI pid discovery), the `wt:<pid>:<startMs>` paneId form + schema bump, the WT read/send seams (E6), and the `launch_console host:` tool surface (E7). Classic public paneId stays `String(hwnd)` — zero back-compat break.

## Tests

- key-locker unit suites green. **Call-shape-only mechanical updates** for the E4 signature changes (`onLocalPaneLaunched`/`injectPane` take the anchor; a `classicAnchor()`/`anchorOf()` helper) — no assertion rewrites.
- `key-locker-inject-target` suite extended: classic carries the anchor identity verbatim; wt assembles pid+time only and early-declines a dead/reused/transient-0 shell.
- key-locker e2e 9/9. The `-SelfTestL2` assertion now pins the richer `{ok, serving, wireParse}` output (strict toEqual kept — the shape change is this PR's designed contract extension, called out for review).
- Full suite: unit green; two unrelated live-desktop e2e blips (`terminal-hidden-input` passes on pinpoint rerun; `browser-cdp-verification` fails identically on unmodified main — pre-existing).

## Version-skew note

Old exe + new Node: the old reader ignores the new props and classic still passes its full old predicate; a wt target presents Hwnd==0 and fails closed. New exe + old Node: the missing fields parse to 0 and the PRIMARY check fails closed (no unverified inject). The exe and dist ship atomically in the release zip, so the second skew exists only in a mixed dev tree.

## Plan

`desktop-touch-mcp-internal@1c2988b:docs/adr-014-v2-r3x-s-pid-gate.md` (E1 classic, E3, E3b, E4, §4, §5) — Opus-GO design gate. PR1 = #537 (watch-side creation-time reuse detection).

### DIFF
--- a/src/engine/key-locker-host.ts
+++ b/src/engine/key-locker-host.ts
@@ -107,20 +107,27 @@ export interface LockerReply {
 
 // L2 wire contracts (the locker owns these frame shapes; the injector orchestrator consumes them).
 
-/** The dedicated-console target of a console-buffer injection (`inject`) — §2.1 of the L2 plan. */
+/** The anchored-pane target of a console-buffer injection (`inject`) — §2.1 of the L2 plan, re-based on
+ *  the S-pid identity (S-pid gate E3). `shellPid`/`shellStartMs` are REQUIRED for both hosts (the
+ *  locker's PRIMARY pid+creation-time re-verify); the window-derived trio is CLASSIC-ONLY (a WT pane has
+ *  no per-pane window). Additive on the wire — an un-rebuilt L0 reader ignores unknown props. */
 export interface InjectTarget {
-  /** The console window HWND (decimal string). */
-  hwnd: string;
+  /** The console window HWND (decimal string) — classic only (absent for wt). */
+  hwnd?: string;
   /**
    * The pid that OWNS the console window — whatever `GetWindowThreadProcessId(hwnd)` returns for the
    * `ConsoleWindowClass` window. L3 supplies `getWindowProcessId(hwnd)` and the locker re-verifies
    * with the SAME API on the same hwnd, so the value matches by CONSTRUCTION (L3 plan §4). The
    * `ConsoleWindowClass` allowlist is what excludes a WT multiplexer; this is the window-owning pid,
-   * not asserted to be a specific conhost-vs-shell process (Opus R1 P3-1).
+   * not asserted to be a specific conhost-vs-shell process (Opus R1 P3-1). Classic only.
    */
-  consolePid: number;
-  /** Opaque hash of the expected pane identity (secondary anchor). */
-  titleFp: string;
+  consolePid?: number;
+  /** Opaque hash of the expected pane identity (secondary anchor). Classic only. */
+  titleFp?: string;
+  /** The spawn-tracked SHELL pid — the AttachConsole target for wt, the identity root for both hosts. */
+  shellPid: number;
+  /** The shell's creation time, ms since the 1601 epoch = floor(FILETIME/10000) (gate §4). Never 0. */
+  shellStartMs: number;
   /** Append Enter after the secret. */
   submit?: boolean;
 }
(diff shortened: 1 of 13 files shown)


Harusame64:
@codex review

Harusame64:
**Review round 1 — both reviewers GREEN.**

- **Opus** phase-boundary review: **GO (P1/P2 zero)**. Verified gate-verbatim conformance (E1-classic/E3/E3b/E4/§4/§5), classic strictly strengthened (new reject paths only; old-Node frames fail closed BEFORE the wt branch so no skew path reaches a wt ACCEPT), the §4 cross-language FILETIME contract exact vs `process.rs` (uint zero-extension both sides; the sign-extension vector arithmetically checked), fail-closed mapping per gate wording, and test changes call-shape-mechanical only (強制命令 4 clean). §3.1/§3.2/Lesson sweeps clean (public docs describe only the unchanged classic paneId; no other consumers of the changed seams). Two P3s, non-blocking: (1) the dropped `bad_target` branch is structurally unreachable — no hole; (2) the gate §5 live classic fill (DF-5 regression guard) + WT dogfood remain as pre-release dogfood steps outside this PR (already mandated by the gate).
- **Codex**: clean ("Didn't find any major issues"), no inline comments.

No code findings ⇒ no fix commit. Ready for merge review.

PR #539, opened 2026-07-16
feat(key-locker): Windows Terminal autofill — the La slice (S-pid PR3)

Harusame64:
## What & why

**The La slice — Windows Terminal autofill.** `key_locker({action:'launch_console'})` now opens a new tab in the user's current Windows Terminal window by DEFAULT (`host:'windows-terminal'`); `host:'classic'` is the explicit escape hatch to the previous dedicated conhost. This completes the S-pid identity re-base: a WT pane has no per-pane window, so its identity is the spawn-tracked shell `{shellPid, shellStartTimeMs}` alone — which PR2 (#538) already carries through the L2 pid+creation-time re-verify.

Implements gate **E1-wt / E2 / E5-wt / E6 / E7 / §3** verbatim (E1-classic/E3/E3b/E4/§4/§5 shipped in PR2; E5 in PR1).

## Scope

- **Engine `pane-id.ts` (E2)**: the one paneId codec over both public forms — classic `String(hwnd)` decimal (unchanged, zero back-compat break) and the new self-describing `wt:<shellPid>:<shellStartTimeMs>`. `parsePaneId` replaces the scattered `BigInt(paneId)` sites; a process-lifetime wt tab-title registry lets the terminal layer resolve a wt pane's window with NO locker import (the rejected "terminal fold" coupling stays rejected). Unit test pins the gate LENGTH INVARIANT (`formatWtPaneId(max).length <= 48`, Opus P3-1).
- **`launchAnchoredConsole({host})` (E1-wt)**: `wt -w 0 new-tab --title <nonce> --suppressApplicationTitle -- powershell -NoProfile -NoExit -EncodedCommand <b64>`. wt.exe resolved at its EXPLICIT `%LOCALAPPDATA%\Microsoft\WindowsApps\wt.exe` alias path (never bare PATH — OQ-S-1). Shell discovered by the `-EncodedCommand` argv marker via the existing native snapshot seam (exactly-1 discipline: 0 keeps polling, **>1 fails the launch** and kills the marked shells; name==powershell + non-zero creation time verified). **Any wt-path failure throws `KeyLockerWtUnavailable` — never a silent classic fallback (E1).** Impl note: reads argv through `snapshotProcessTree().commandLine`, not a live WMI query — same marker/discipline, no WMI dependency, unit-fakeable (gate doc synced, `internal@4632c35`).
- **Driver (E5-wt)**: `tickWatch` prunes a wt `PaneRecord` whose shell exited (a wt pane has no `window_disappeared` signal), full W2 close bundle; an empty snapshot never mass-prunes.
- **`terminal.ts` (E6)**: `resolvePaneTitle` / `findTerminalWindowByPaneId` branch on the pane kind — classic keeps its hwnd-direct / title-round-trip seams; a wt pane resolves to its WT window ONLY while its nonce tab title is showing (active tab), else declines (the honest contract: autofill/reads operate on the active tab; switching away pauses the loop fail-safe — injection re-verifies pid+time regardless). paneId schema `max(32)` → `max(48)` (additive). The dispatch hook now keys on the PUBLIC paneId so a wt pane's arm fires (`String(win.hwnd)` would be the WT host window).
- **Wiring + tool (E7)**: `ensureAnchoredConsole({host})` defaults to `windows-terminal` (OQ-R3x-2 — a bare launch_console opens a new tab in the user's CURRENT window; that IS the feature). Reuse is host-aware; a wt pane's reusability is its ANCHOR-IDENTITY liveness (shell alive + exact creation-time), no window needed (an inactive tab is still a valid pane). `KeyLockerWtUnavailable` SUGGESTS names `host:'classic'` as the retry. Tool description + stub catalog regenerated (tool count unchanged). CHANGELOG `[Unreleased]` added.

## Tests

Full unit suite green:
- pane-id codec + length invariant + malformed/0-sentinel declines;
- **gate §3 host-shape equivalence pair** (conhost vs WindowsTerminal — interactive-ssh dispatch, sudo→ssh grandchild, unreadable descendant, and WT sibling-tab invisibility) proving identical descendant-correlation classification under both hosts;
- wt close-prune (shell exit + empty-snapshot guard);
- host-aware reuse + wt anchor-identity reuse/reuse-after-pid-reuse/cap.

No C# / exe change in this PR (verified — the identity contract landed in PR2). Two unrelated live-desktop e2e blips: `browser-cdp-verification` fails identically on unmodified main (pre-existing); `terminal-console-paste-load` passes on pinpoint rerun (latency flaky).

## Out of scope

The gate §5 **live WT dogfood** (fill end-to-end + CJK) is a pre-release step (tracked for the R3.x release), not a PR gate — it needs a real desktop.

## Plan

`desktop-touch-mcp-internal@4632c35:docs/adr-014-v2-r3x-s-pid-gate.md` (E1-wt, E2, E5-wt, E6, E7, §3) — Opus-GO design gate. PR1 = #537, PR2 = #538.


Harusame64:
@codex review

Harusame64:
@codex review

Harusame64:
**Review round 1 — Opus GO (P1/P2 zero); findings applied.**

- **Opus** phase-boundary review (production bar): **GO — no P1/P2.** Verified gate-verbatim conformance (E1-wt/E2/E5-wt/E6/E7/§3), the native-snapshot discovery deviation matches the synced gate note and preserves the exactly-1 identity contract (exact argv-token equality, stricter than WMI LIKE), and the security invariants end-to-end: I2 traced launch→wt paneId dispatch key→arm→AttachConsole(shellPid); the E6 honest contract (inactive/switched-away tab → null → decline, no wrong-target read); the secret never rides the WT send path; wt.exe explicit-alias PATH hardening + >1 fail-loud. E5-wt prune (empty-snapshot guard, classic not pruned by shell exit, full W2 bundle), E2 codec totality, E7 host-aware reuse + anchor-identity liveness + registry lifecycle all confirmed. Sweeps clean (§3.1 catalog/CHANGELOG synced + user-facing English; §3.2 classic path byte-unchanged; 強制命令 4 test legitimacy).
  - **P3-1 (applied):** the wt startMs regex accepted 16 digits, but 16-digit values exceed Number.MAX_SAFE_INTEGER (parsing would round). Tightened to 14 digits (a real FILETIME-ms is ~14 digits; 14 stays inside the safe-integer range), and the length-invariant test now bounds the true parser ceiling. Commit `7e9f57e`.
  - **P3-2 (no action — gate-acknowledged):** the Mode-A exit probe uses WM_CHAR, which WT does not process → a probe to a WT pane may no-op → bounded `runToExit` timeout. This is gate OQ-S-5 (deferred to dogfood, bounded-safe, never identity-relevant; the fill itself is unaffected). Confirm at live dogfood.
- **CI:** the first run failed only on the ADR-021 failwith-callsite-shape gate (a generated fixture — the E6 terminal:send branch dropped one failWith callsite + shifted line numbers). Regenerated in `7e9f57e`.
- **Gate §5 live-WT dogfood:** a pre-RELEASE step (needs a real desktop; OQ-S-3 already spike-cleared read/discovery on hardware), not a merge gate.

Re-triggered Codex on the new HEAD.

Harusame64:
@codex review

PR #540, opened 2026-07-16
fix(key-locker): WT-capable Mode-A exit probe channel (Codex P1 follow-up to #539)

Harusame64:
## What & why

**Follow-up to #539 — a Codex-P1 fix that landed on the branch AFTER #539 was merged, so it is NOT in `main`.** #539 merged at commit `7e9f57e`; the fix commit `a12f5a6` was pushed afterward and was left behind. This PR carries it to `main` unchanged (cherry-pick).

### The bug (present in `main` right now)

Since #539, `key_locker({action:'launch_console'})` defaults to a Windows Terminal tab. But the Mode-A exit probe in `runToExit` (the completion detector for a ONE-SHOT credential command — `sudo <cmd>`, `git push`, `ssh host cmd`) was hardcoded to `method:"background"`, which Windows Terminal **rejects** (`wt_xaml_pipeline` — WT's XAML pipeline drops WM_CHAR). So in the now-default WT flow:

1. the secret is captured and filled correctly (the fill rides `AttachConsole`, pid-addressed — **unaffected**);
2. the exit probe is never delivered → the echo-immune sentinel never writes;
3. `runToExit` polls to its full timeout → `awaitLanded` returns not-accepted;
4. the capture loop **discards the just-captured, successfully-filled secret** (the D5 save-gate `finally` deletes it).

Net: a one-shot credential in the default WT flow is captured, filled, and then thrown away — the user is re-prompted next time. Interactive logins (`ssh user@host`, `sudo -i`) are unaffected (Mode-B is read-only, no send).

### The fix (gate E6 intent — "the Mode-A exit probe uses the WT-capable send machinery")

- `exitProbeMethod(paneId)` — a pure, exported, unit-pinned policy: a **wt** pane uses `foreground_flash` (resolves to `clipboard_flash` on the WT XAML window — a ~60ms focus flash of a **benign non-secret sentinel**; the secret never rides this path), a **classic** pane keeps `background` (resolves to no-steal `wm_char`; byte-equivalent to the shipped probe — the classic regression-guard path is bit-identical).
- `runToExit` now checks the `terminalSendHandler` return via `sendSucceeded()`: a typed send FAILURE (a `failWith`, not a throw — previously ignored) is treated as a prompt `probe_error` (fail-safe discard) instead of polling to the full `EXIT_PROBE_MS` timeout. Same fail-safe outcome, promptly and with the real cause.

## Review status (carried from #539)

Both reviewers verified this exact commit ON the #539 branch before it was orphaned:
- **Opus Round 2: GO** — foreground_flash confirmed the correct WT-capable channel (clipboard_flash for WT / no-steal wm_char for classic); classic byte-equivalent; no new disclosure path (the probe is `buildExitProbe`, a non-secret sentinel; the secret rides AttachConsole); `sendSucceeded` fail-closed. Two P3s, both gate-accepted (the ~60ms focus flash + the clipboard use are OQ-S-5, dogfood-confirmed).
- **Codex Round 1: P1** (this is the finding being fixed — Codex's API-contract axis caught the discard consequence that Opus had classed P3 "bounded timeout").

## Tests

`key-locker-anchoring` 23/23 (adds the `exitProbeMethod` pin: wt→foreground_flash, classic→background, malformed→background). Build green.

## Plan

`desktop-touch-mcp-internal@7a39bbc:docs/adr-014-v2-r3x-s-pid-gate.md` (E6, OQ-S-5 — updated to "resolved in La").

### DIFF
--- a/src/tools/key-locker-wiring.ts
+++ b/src/tools/key-locker-wiring.ts
@@ -19,6 +19,7 @@
 //   gap6 landed  — `runToExit` OBSERVES the already-run command's exit via an epilogue-only probe
 //                  (`buildExitProbe`), NEVER re-running it.
 
+import type { ToolResult } from "./_types.js";
 import type { BindingMeta } from "../engine/key-locker/binding-store.js";
 import { BindingStore } from "../engine/key-locker/binding-store.js";
 import type { BindingUri } from "../engine/key-locker/binding.js";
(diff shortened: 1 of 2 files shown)


Harusame64:
@codex review

Harusame64:
**Reviewers GREEN — ready for merge.**

- **Codex** (commit `6446267`): clean, no inline comments.
- **Opus**: GO — this exact fix (as `a12f5a6` on the #539 branch) was verified in Opus Round 2 before it was orphaned: foreground_flash confirmed the correct WT-capable channel (clipboard_flash for WT / no-steal wm_char for classic), classic byte-equivalent, no new disclosure path, `sendSucceeded` fail-closed.
- **CI**: build / CodeQL / docker-build / Analyze all green.

This is the P1 fix that #539 merged without (it landed on the branch after #539 was cut at `7e9f57e`). Merging this closes the default-WT-flow one-shot-credential discard bug currently in `main`.

PR #541, opened 2026-07-17
fix(key-locker): honor ssh options after the destination (F-3, P1 security)

Harusame64:
## What

`ssh user@host -p 2222` silently resolved as **port 22**. The argv parser stopped at the destination, so every option behind it was dropped. Real OpenSSH accepts them — `ssh.c` runs getopt **again** after the destination (`goto again`) — which is why, in the live dogfood, the pane genuinely reached the `:2222` sshd while we derived `:22`.

## Why it is P1, not "autofill didn't fire"

The derived canonical key carried the **port-22 known_hosts fingerprints**, so it bit-matched a stored port-22 binding. Nothing downstream can catch that:

| guard | what it actually checks |
|---|---|
| `BindingStore.resolve` | a plain string-key map lookup |
| the fp-set | derived **from the wrong endpoint** — so it equals the stored one. Broken at derivation, not comparison |
| `sessionMatches` | `execHost` + `isRemote`, where `execHost` is a **portless, userless** bare host |
| injection-instant re-verify | shell **pid + creation time** — process identity, never endpoint identity |

⇒ the port-22 secret could be typed into the port-2222 server's prompt.

`ssh host -l otheruser` is the same bug and easier to hit: `ssh -G host` yields the default user with a **bit-identical** fp-set, so the secret reaches a **different principal on the same host**.

**The same parser fed the fail-safe meant to catch this**, so both halves broke together. Any trailing option (`ssh host -v`) counted as a "remote command" ⇒ pane classified one-shot ⇒ no remote frame pushed ⇒ the W-2b scan did not flag the login either ⇒ a later `sudo` fills a **LOCAL** secret into the **REMOTE** prompt. (`ssh-session-watch.ts` already fail-safed an *unreadable* argv; a readable-but-misparsed one fell straight through to trust-local.)

Reachable on the shipped classic path — independent of the Windows Terminal work.

## How

- **Parse with ssh's own two-pass rule.** Options are options *wherever* they appear; the first non-option token after the destination begins the remote command. One guard condition is the root fix.
- **`ParsedSshCommand` gains `remoteCommand` + `undecidable`.** A letter outside both allow-lists — or no destination outside a query mode — makes the parse untrustworthy.
- **`interactiveSshTarget` → `classifySshLogin`.** Host-or-null could not say "I don't know", and treating doubt as "opens nothing" *is* the inversion. Now `interactive` / `none` / `undecidable`; every consumer declines on `undecidable`, and only a **proven** `none` is trusted.
- **`ssh -G` receives the whole argv**, so the endpoint decision never rides on the completeness of our own flag table. Verified against the real binary: `ssh -G h sudo apt update` prints the same config as `ssh -G h` (only the stderr pty warning differs; we parse stdout).
- **`1` is deliberately in neither allow-list** — real ssh is fatal on it (`SSH protocol v.1 is no longer supported`), so a pushed frame would track a still-local pane as remote: the mirror of this bug. **`2` is no-arg** — undocumented in the synopsis but accepted by every build, so `ssh -2 host` opens a real session.

## Verification

Measured on this machine against **two** independent builds (Windows OpenSSH 9.5p2 and Git/MSYS 10.3p1 — the two the DF-3 comment warns about):

```
before:  ssh dtmdogfood@localhost -p 2222  →  host-not-known, port 22  →  deriveBinding = null
after :  ssh dtmdogfood@localhost -p 2222  →  ssh://dtmdogfood@localhost:2222
         both option orders now converge on the identical canonical URI
```

Tests (4605 unit passed, 0 failed):
- the **26-row argv decision table** (pure)
- the **three-state classifier** + `recordDispatch` arms
- **W-2b** flags post-destination and unclassifiable forms; still does not flag a proven one-shot
- real-`ssh -G` post-destination cases (`-p` / `-l` / `-J` / trailing command)
- a **security regression pin**: a port-22 binding never resolves for `-p 2222`
- a **flag-table canary** that re-derives both allow-lists from the local ssh's usage synopsis, so an OpenSSH upgrade that moves a letter fails loudly instead of silently costing autofill

No existing assertion changed. The derivation suite's `ssh -G` stub was made faithful to the new call shape (it assumed the destination was the last token, so `ssh h sudo apt update` read as a connection to `update`); it now reuses `parseSshCommand` deliberately — the parser is pinned independently by the pure table and the real-`ssh -G` suite, so the stub is scaffolding, not a second implementation.

## Notes for review

- Found by the **live Windows Terminal dogfood**, not by a test.
- `landed-detection.ts` was a fourth caller the plan's table missed (the compiler caught it). It was a victim of the same bug: `ssh h -p 2222` classified one-shot ⇒ Mode A waited for an exit that never comes ⇒ timeout ⇒ **a correct secret was discarded**. It now falls to Mode B correctly.
- Follow-ups (Windows Terminal launch + the arm race, `terminal(action='run')` paneId, locker observability) are separate PRs; this one is first because it is the only shipped one.

### DIFF
--- a/src/engine/key-locker/command-derivation.ts
+++ b/src/engine/key-locker/command-derivation.ts
@@ -229,6 +229,11 @@ async function deriveSsh(args: string[], exec: ExecFn): Promise<BindingUri | nul
   // Query / no-login modes never prompt (`-G` config query, `-Q` algorithm query, `-V` version).
   // Modes that still authenticate (-T no-pty, -N no-remote-cmd) derive normally.
   const parsed = parseSshCommand(args);
+  // An argv we cannot classify is one whose destination AND remote-command boundary are both guesses — so
+  // which endpoint this secret would be typed at is a guess too. Fail closed: no binding, no lookup, no
+  // save. (Doubt is checked BEFORE queryMode for the same reason it is in `classifySshLogin`: with an
+  // unknown letter present we cannot even trust that `-G` is a query rather than that letter's value.)
+  if (parsed.undecidable) return null;
   if (parsed.queryMode || parsed.destination === undefined) return null;
   const resolved = await resolveCanonicalForSshCommand(args, exec);
   // host-not-known / unresolvable ⇒ fail closed: no binding, no lookup, no save.
--- a/src/engine/key-locker/key-locker-capture-driver.ts
+++ b/src/engine/key-locker/key-locker-capture-driver.ts
@@ -602,8 +602,9 @@ function sessionMatches(live: PaneSession, expected: SessionFrame): boolean {
 /**
  * The ssh-named DESCENDANTS of `shellPid` in a snapshot (subtree, not just direct children — the exempt
  * delta must see a `sudo ssh`/wrapper/tmux-reparented login the same way the W-2b scan does). For each,
- * report its interactive-login HOST (`commandLine` → `interactiveSshTarget`, or null for a tunnel/one-shot/
- * unreadable ssh) + its creation time (for the pid+time exempt). A `seen` set guards pid-reuse cycles.
+ * report its interactive-login HOST (`commandLine` → `classifySshLogin`, or null for a tunnel/one-shot/
+ * unclassifiable/unreadable ssh) + its creation time (for the pid+time exempt). A `seen` set guards
+ * pid-reuse cycles.
  */
 function sshDescendants(snap: ProcessSnapshot, shellPid: number): Map<number, { host: string | null; startTimeMs: number }> {
   const out = new Map<number, { host: string | null; startTimeMs: number }>();
--- a/src/engine/key-locker/key-locker-capture-driver.ts
+++ b/src/engine/key-locker/key-locker-capture-driver.ts
@@ -67,7 +67,7 @@ import type { PaneAnchor } from "./inject-target.js";
 import type { InjectResult } from "./injector.js";
 import {
   ENV_ASSIGN_RE,
-  interactiveSshTarget,
+  classifySshLogin,
   isKnownSession,
   programOf,
   type PaneSession,
@@ -624,7 +624,8 @@ function sshDescendants(snap: ProcessSnapshot, shellPid: number): Map<number, { 
       const id = snap.identify(pid);
       if (id.name !== SSH_PROGRAM) continue;
       const argv = snap.commandLine(pid);
-      const host = argv === null ? null : interactiveSshTarget(argv.slice(1));
+      const cls = argv === null ? ({ kind: "none" } as const) : classifySshLogin(argv.slice(1));
+      const host = cls.kind === "interactive" ? cls.host : null;
       out.set(pid, { host, startTimeMs: id.startTimeMs });
     }
   }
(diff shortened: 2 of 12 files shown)


Harusame64:
@codex review

Harusame64:
@codex review

Round 1 findings applied in `cdf6814`. Two changes worth re-checking, both from the Opus review:

1. **Doubt now outranks a query** in `classifySshLogin`. With a future with-arg letter, real ssh consumes the query flag as that letter's value and opens a session (`ssh -z -G h`), so answering `none` there left the pane trusted-local while a remote login was open. Real queries are unaffected — the parser's post-loop rule already exempts them, so `-V` / `-Q cipher` / `-G h` still answer `none`.

2. **The whole-argv `ssh -G` call is withdrawn.** It published dispatched secrets on ssh.exe's command line (`ssh h mysql -pSECRET` reaches the resolver, because `ssh` alone arms the poller). Resolution now passes options + destination only; the endpoint is identical because `optionArgs` carries post-destination options after the two-pass fix. `deriveSsh` also declines on `undecidable` — it was the one consumer that never mapped doubt to a decline, which is what made the whole-argv detour look necessary.

On your P2 (keep tunnel SSH on the auth-prompt landed path): confirmed as a real defect, and thank you — but it is **pre-existing, not a regression here**. The old classifier also returned null for `-N`, so `ssh -N -L … host` selected Mode A before this PR too, and `tests/unit/key-locker-landed-detection.test.ts` deliberately pins that invocation to one-shot today. Fixing it inverts that pin's intent, so it is tracked as a separate change — landed-mode has to ask "does this exit on its own?" rather than "does it open a shell" — and gated before release rather than folded into this security fix.

One correction to the obvious next step, in case it saves you a round: plain Mode B does not fix it either. `ssh -N` prints nothing after a successful auth, so the last non-empty line stays `…password:`, `stillHiddenPrompt` remains true, and the bounded poll still ends in a discard. The fix needs a liveness-based accept (the ssh child is alive and no denial line appeared), which is why it is being measured on real hardware first rather than designed from the fake.

Harusame64:
@codex review

Your P2 was right, and it was a **regression this PR introduced** — thank you. Fixed in `d39b956`.

Verified before fixing: `main` ignored post-destination options entirely, so it happened to answer "no session" for `ssh h -p`; the two-pass fix made it an interactive login and the tracker pushed a remote frame for a pane that never left local (measured: `{"execHost":"prod.example.com","isRemote":true}`). Real ssh exits first — `option requires an argument -- p`, status 255.

The fix follows your suggestion but lands on **"proven non-login" rather than "undecidable"**, because we know exactly what ssh does here: `undecidable` routes to `markUnknown`, and an UNKNOWN pane does not recover on its own, so a typo would cost the pane its autofill entirely. `malformed` is therefore a separate channel from `undecidable` — "we know it fails" vs "we don't know what happens" — and `undecidable` still wins when both are set, since an unknown letter may be the one that swallowed the "missing" value.

The verdict rests on getopt's contract (an optstring letter declared with `:` and given no argument makes getopt error and ssh exit), not on `-p` specifically, so the real-ssh test pins the contract with one representative letter instead of one case per letter. The per-letter fact that matters — whether a letter takes a value at all — is `SSH_FLAGS_WITH_ARG`, which the synopsis canary already guards.

Pinned: `h -p` / `h -l` / `h -o` / `h -4p` (the cluster tail reaches the same branch — easy to miss), plus `-p h` which is NOT malformed (the value is available; it eats the destination, so it stays undecidable), the pane staying local and usable afterwards, a later local `sudo` still deriving locally, and the getopt contract itself. All mutation-checked: reverting the fix turns nine of them red.

Also in this round, from the Opus review: `classifySshLogin` now checks doubt before a query (`ssh -z -G h` — a future with-arg `-z` would eat `-G`, so real ssh opens a session while our flag scan sees a query), and the whole-argv `ssh -G` call is withdrawn because it published dispatched secrets on ssh.exe's command line.

Harusame64:
@codex review

Round 4 applied in `8d475d8`. This addresses your three Round 3 P2s:

1. **`-O` (control-master command)** and **`-s` (subsystem)** — both correct: after the two-pass fix they reach the classifier as post-destination options, but step 7 only tested `has("N")||has("f")||has("W")`, so both were misread as interactive logins and pushed a remote frame for a pane that never left local. Fixed structurally: the inline chain is now a named `SSH_FLAGS_NO_LOGIN_SHELL` set (`NfWOs`), each letter carrying its reason, checked **before** the remote-command step. `-s sftp` now lands on `none` because a subsystem is not a login shell — not by the coincidence that "sftp" is a remote-command token. Flag-before twins (`-O check h` / `-s h`), which were broken on `main` too, are fixed by the same letter-based rule and pinned.

2. **`ssh h -p -v` (`-v` eaten as the port value)** — this one stays `interactive` **deliberately**. It is a bad *value*, not a bad *letter*: the parse is neither `malformed` (a value is present) nor `undecidable` (no unknown letter is scanned). Telling a bad port from a good one means per-option value semantics = re-implementing ssh's config parser in the sync module, which the plan forbids. The resolver still fails closed because the real `ssh -G` rejects it; the spurious frame is sunk by the watch's unwatched-frame backstop on the next tick. Pinned as a visible F-7 known-inaccuracy (same class as `h -p 2222x`) so the gap is documented, not silently absent.

Verified: build clean, all key-locker unit suites green (session-tracker 100/100).

Harusame64:
@codex review

Follow-up pushed in `75dfbd5` (HEAD). This folds `-n` into `SSH_FLAGS_NO_LOGIN_SHELL` (now `NfWOsn`) on a measurement rather than a reading.

`-n` redirects the remote stdin from `/dev/null`, so the remote shell hits EOF and exits at once — the pane returns to local. Measured against a throwaway local sshd (portable Windows OpenSSH 10.0.0.0p2, pubkey auth):

- `ssh -n <dest>` (no remote command) → **exit 0 in ~400 ms** (357/386/488 over 3 runs): the shell prints its prompt then EOF-exits.
- plain `ssh <dest>` (stdin still open) → held open (killed at 120 s): the shell persists.

So `-n` opens a shell that does not take the pane over (the mirror of `-N`, which persists but opens no shell). Classifying it `interactive` would push a remote frame for a pane that is local again within ~400 ms, and the vanishing child would trip the watch's unwatched-frame backstop into `markUnknown` — wrong twice. It is letter-only decidable, so it sits beside `N`/`f`/`W`/`O`/`s`. Pinned: `h -n` / `-n h` → `none` + a recordDispatch stays-local test.

Please review HEAD `75dfbd5` (Round 4 `-O`/`-s` set + this `-n` follow-up together).

PR #546, opened 2026-07-22
fix(terminal/key_locker): guide LLMs away from paneId/windowTitle mixup + accept paneId in terminal run

Harusame64:
## Why

An assistant dogfooding `key_locker({action:'launch_console'})` passed the returned **windowTitle** (`dtm-locker-console-<hex>`) into terminal's **`paneId`** slot. That value never parses as a paneId, so the generic `TerminalWindowNotFound` suggest ("run desktop_discover / try a partial title") pointed the *opposite* way — while the very same value, passed as `windowTitle`, would have resolved. The assistant then hallucinated a non-existent `terminal({action:'list_panes'})` trying to re-discover the pane.

Root cause: the failure message misdirected, and `run` (the most convenient one-shot action) did not accept the pane handle that `launch_console` returns, pushing callers back onto title-based targeting — the exact drift-prone path `paneId` exists to avoid.

## What changed (all additive / non-breaking — error code stays `TerminalWindowNotFound`)

- **`paneIdMissSuggest()`** — shape-aware recovery hints branched on the failure:
  - a launch_console **windowTitle in the paneId slot** → name the `paneId` field (and offer passing it as `windowTitle`);
  - any other **malformed** handle → point at launch_console reuse (`fresh:false`);
  - **well-formed-but-gone** → explain the wt inactive-tab case.
  Wired into terminal `read`/`send`/`run` paneId-miss paths via `failCode`.
- **`terminal(action='run')` now accepts `paneId`** (parity with read/send). The dispatcher resolves it to the pane's live-unique title up front **and** passes the public paneId through to `terminalRunHandler`, so the autofill dispatch keys by the paneId — never `String(hwnd)` (a wt pane's host-window hwnd is not the driver's pane key; mirrors `terminalSendHandler.markDispatched`).
- **`launch_console` result carries a `hint`** field naming the paneId field; `key_locker` + `terminal` descriptions clarify paneId-vs-windowTitle, note there is no pane-listing action, and that `fresh:false` reuse returns the paneId.

## Tests

- `tests/unit/terminal-paneid-target.test.ts`: added `paneIdMissSuggest` branch coverage + `run` paneId parity + neither-arg rejection.
- Full unit suite green. The 5 failing e2e tests (`context-consistency` hasModal / `browser-cdp` headless) are pre-existing and environmental — they read the live desktop / headless Chrome and touch neither terminal paneId nor key_locker.

## Not in this PR (follow-up)

`hasModal` over-reporting (desktop_state) — the dogfood's "modal false detection" symptom reproduces in the environmental `hasModal` e2e failures here. Tracked as a desktop_state follow-up; needs investigation before a fix.


Harusame64:
@codex review

Harusame64:
@codex review

Round 2 applied your P2 ("Keep paneId targeting during run polling"): the run handler now re-resolves the pane title via `currentReadTitle()` before every post-send read (initial post-send / poll loop / final read), so a classic console that retitles to `user@host` mid-run is tracked. The pre-send baseline read intentionally keeps the original title. Please re-review at HEAD (24b77fc).

Harusame64:
@codex review

Round 3 applied your R1 P2 (run polling drift) and then Opus R3 flagged that the naive re-resolve could fall back to a stale title and read a same-title sibling pane (disclosure). Final state at HEAD:

- `currentReadTitle()` returns `string | null` — a fresh `resolvePaneTitle(paneId)`, no stale fallback.
- On null the post-send reads SKIP: initial post-send reads are skipped, the poll loop `continue`s (genuine closure owned by the `isWindowStillAlive(hwnd)` check), and the final read passes the paneId to `terminalReadHandler` which declines safely instead of reading a sibling.

Please re-review the latest HEAD.

Harusame64:
@codex review

Addressed your #3 (closed WT pane detection): added `isPaneShellAlive(paneId)` — on a null re-resolve the poll loop now verifies the wt paneId's shell pid/startTime identity (same idiom as `inject-target.ts`) and ends with `window_closed` when the shell is gone, instead of skipping until timeout; an inactive (alive) tab keeps polling (pause). Your #1 (re-resolve during polling) and #2 (stale-title fallback) were already fixed by the earlier re-resolve + null-skip commits. Please re-review the latest HEAD.

Harusame64:
### Review status — the two Codex comments re-pinned on the latest commit are already-fixed findings

Codex re-pins prior inline comments onto the new HEAD's lines; both current comments describe behaviour that has since been fixed (verified in code):

- **"Keep paneId targeting during run polling"** (pinned at `terminal.ts:2929`) — 2929 is the dispatcher's one-time resolve for the *send target + baseline*. The **poll loop re-resolves the pane title every read** via `currentReadTitle()` (`terminal.ts:2560-2578`) and the **final read passes the paneId** straight to `terminalReadHandler` (safe re-resolve). The stale-title premise no longer holds.
- **"Detect closed WT panes before skipping reads"** (pinned at `terminal.ts:2578`) — the `isPaneShellAlive(paneId)` check sits right above that line (`terminal.ts:2574-2577`): a closed wt tab (shell pid/startTime gone) now ends the run with `window_closed`; an inactive tab keeps polling.

Codex's substantive findings across rounds — #1 (re-resolve during polling), #2 (no stale-title fallback), #3 (closed-wt detection) — are all addressed. Opus (5 rounds) and Fable (2 rounds) both report P1/P2 = 0. CI green.

PR #547, opened 2026-07-22
fix(deps): bump hono to 4.12.31 and fast-uri to 3.1.4 (Dependabot #29-33)

Harusame64:
Resolves 5 of the 6 open Dependabot alerts, all transitive via `@modelcontextprotocol/sdk`.

| Alert | Package | Fix | How |
|---|---|---|---|
| #29 #30 #31 | hono | 4.12.26 → **4.12.31** | within SDK's `^4.11.4` — lockfile bump |
| #32 #33 | fast-uri | 3.1.2 → **3.1.4** | within ajv's `^3.0.1` — lockfile bump |

**Not fixed here — #28 (`@hono/node-server` serve-static path traversal):** the fix (2.0.5) is a major bump the SDK does not yet allow (it pins `^1.19.9`). The vulnerable `serve-static` path is not exercised by the MCP JSON-RPC transport, so it is not exploitable in this server. Tracked upstream; recommend dismissing the alert with that rationale.

`package.json` unchanged (transitive bumps only). Build passes; CI runs the full suite. Ships to users with the next release (the runtime zip rebuilds from this lockfile at tag time).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4871,9 +4871,9 @@
       }
     },
     "node_modules/fast-uri": {
-      "version": "3.1.2",
-      "resolved": "https://registry.npmjs.org/fast-uri/-/fast-uri-3.1.2.tgz",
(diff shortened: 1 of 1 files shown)


Harusame64:
@codex review

PR #548, opened 2026-07-25
fix(desktop_act): compare the viewport gate against the window an element came from

Harusame64:
ADR-029 Phase 1. Plan: `desktop-touch-mcp-internal@f201fdc:docs/adr-029-multimonitor-visual-entity-gate.md`; the two decisions taken while implementing are in Harusame64/desktop-touch-mcp-internal#33.

## What was broken

Visually discovered elements (OCR / screen recognition — targets with no accessibility tree: remote desktop, games, canvas apps) were checked against whatever window happened to be in the foreground. On a multi-monitor desktop that is usually a *different* window, so `desktop_act` refused plainly visible elements with `entity_outside_viewport`, and the advice it gave ("scroll it into view") could never fix it.

## The gate

`UiEntity.origin` now carries the discovery-time target of the primary candidate, and the gate compares the element rect against **that window's current rect**. Comparing against the containing window would be tautological — the element was observed inside it — which is why the identity has to be carried from discovery time.

The gate returns a verdict instead of a boolean, so the block cases give different advice:

| verdict | meaning | recovery |
|---|---|---|
| `null` | reachable | proceed |
| `entity_outside_viewport` | scrolled out of its own window, or the window moved / closed | `scroll(action='to_element')` or re-`desktop_discover` |
| `origin_window_not_visible` (new) | origin window minimised / DWM-cloaked — nothing is drawn there | `focus_window(windowTitle)` then `desktop_discover` |

A virtual-screen fallback for the minimised case was rejected: it turns the gate into a no-op and lets the click land on an unrelated window. Non-HWND origins (title / `@active` / browser tabs) still fall back to a virtual-screen bounds check, which blocks coordinates off every monitor rather than passing everything.

The three conservative passes (no rect / structured source / Win32 failure) are unchanged.

`isInViewport` was renamed to `checkViewport` deliberately: leaving the `is…` name while returning `null`-means-ok would let `!check(...)` silently invert the meaning at any future call site.

## The coordinate guard (why it ships in the same PR)

Opening the gate exposes the second half of the bug: nut.js clamps any point outside the primary monitor into it, so a click aimed at a second monitor reports success while landing somewhere else — a worse failure than the block it replaces. `assertCoordinateReachable` refuses such a point **before the cursor moves** (after libnut clamps, the damage is done, and a negative coordinate can even trip the failsafe corner).

Guarded at every site where a coordinate is final:

- `moveTo` in `mouse.ts` — covers `mouse_click` / `mouse_move` / `scroll`
- both endpoints of `mouse_drag`, up front, so it cannot abort mid-drag with the button held
- the executor's mouse routes — **`desktop_act` does not go through `mouse.ts`**; it drives nut.js directly, including the UIA→mouse downgrade that the gate passes unconditionally, and `cdpClick`

`browser_click` keeps its own nut.js move and is left for Phase 2 (recorded as R8) — it is a different tool family with its own envelope and needs its own blast-radius call.

## Why a new `TouchFailReason`

Folding the refusal into `executor_failed` would hand the caller that reason's advice — "fall back to `mouse_click`" — which walks straight back into the guard. The original exception is also discarded twice before reaching the caller (touch loop catch, then a fresh `ExecutorFailedError` in the raw handler), so steering by message alone is structurally impossible. The touch loop matches the error by `name` and reports `coordinate_outside_reachable_bounds`; the raw handler emits an envelope whose `try_next` names what works today.

## Behaviour change

Coordinates on a monitor placed left / right / above / below the primary one now fail fast with `CoordinateOutsideReachableBounds` where they previously returned success and clicked the wrong place. CHANGELOG says so under Unreleased. Full multi-monitor mouse input (native `SendInput` + `MOUSEEVENTF_VIRTUALDESK`) is Phase 2, which also widens the reachable region to the whole virtual screen.

## Tests

- gate: passes an element whose window is not focused; blocks when the element centre left its origin window **and now sits over a different window** (a tautological implementation would pass — this pins the rejected design out); closed origin → stale; minimised / cloaked → `origin_window_not_visible`; non-HWND origin on / off the virtual screen; the three conservative passes
- resolver: `origin` set from the primary candidate, and from the most recent one in a merged group
- guard: bounds arithmetic incl. the half-open edges, negative coordinates, unknown region → allow
- executor: both mouse routes refuse without calling `mouseClick`, and the UIA→mouse downgrade case specifically
- touch loop: the refusal is promoted, a generic error still reports `executor_failed`
- envelope: `try_next` for the new reason is coordinate-specific, not executor_failed's

`tests/unit/issue-211-classify-branch-producer-pin.test.ts` gained one recognised producer pattern — `new <Code>Error(...)` from the typed-error family. Without it the first classify branch fed by a typed error class reads as dead code.

Full suite run once: unit green. Two e2e failures are unrelated to this diff (`browser-cdp-verification` — headless Chrome rescue-branch classification, no code in this diff; `terminal-hidden-input` — background WM_CHAR delivery, passed on re-run).

@codex review


Harusame64:
@codex review

Harusame64:
## Round 1 findings applied (`9833e69`)

Two P1s, three P2s and three P3s from the review round.

**P1-1 — stale recovery wording.** Two of the four README recovery lines still said "scroll, then re-discover" for `entity_outside_viewport`, so the repo gave two different answers in two headed lists. Both now describe the same two branches.

**P1-2 — filtered-out windows were misread as closed.** `enumWindowsInZOrder` drops invisible, **untitled** and sub-50px windows, so "absent from the enumeration" is not "closed". An untitled canvas / game / remote-session window — precisely the accessibility-blind target this gate exists for — would have been blocked as stale permanently, a regression of the same shape as the bug being fixed. The gate now probes such an HWND directly (`getWindowRenderState`, new in `win32.ts`): a live-but-filtered window is compared against its real rect, a hidden / minimised / cloaked one reports `origin_window_not_visible`, and only a handle that is genuinely gone reports a stale view.

**P2-1 — classify arm placement** matched its comment's claim (it sat below `window not found` / `timeout`, not above).

**P2-2 — the single-monitor end-to-end reproduction is in** (`tests/e2e/adr-029-viewport-gate.test.ts`). It stages the real bug with two real windows, asserts the precondition that the element lies outside the foreground window so it cannot pass for the wrong reason, and runs the production gate **with no injected dependencies** — the unit tests inject fakes and would not catch a Win32-level regression. Also added a source-level pin on the facade wiring: nothing else in the suite would notice `checkViewport` being reverted to a constant pass.

**P2-3 — blast-radius sweep done.** No caller in the repo passes off-primary coordinates to the guarded mouse handlers; the negative coordinates that exist are in geometry / ROI / dispatch unit tests that never move the cursor. The macro replay path forwards whatever a user recorded, which is what the CHANGELOG warns about.

**P3-1** — an all-numeric window *title* is indistinguishable from an HWND string and providers fall back to the title; when a live window carries that title the origin is now treated as title-based instead of being probed as a handle and blocked as stale.
**P3-3** — unknown monitor bounds still allow everything (the machine stays usable) but now say so once instead of silently restoring the misclick the guard exists to prevent.
**P3-4** — the two historical v2 rollout documents whose recovery tables predate this change now point at the current sources of truth.

**P3-5 (`## [Unreleased]` heading)** — kept deliberately: this is not a release commit, and the release process renames the section when the version is cut.

Full suite: **5030 unit tests pass**. Two e2e failures are unrelated to this diff — `browser-cdp-verification` (headless Chrome rescue-branch classification; no file in this diff, and main's CI is green) and `terminal-hidden-input` (background WM_CHAR delivery; passes on individual re-run).

@codex review

Harusame64:
## Round 2 findings applied (`975ee70`)

Round 2 found **no P1**. Two P2s and the P3 set are addressed.

**P2-1 — the ordering fix was pinned by nothing.** The classify arm now sits above the generic arms and says the placement is load-bearing, but no test enforced it. Pinned the same way the SpawnFailed ordering already is: a message that *also* contains "window not found" / "timed out" / "element not found" must still classify as the coordinate refusal.

**P2-2 — the e2e overstated its coverage.** Every window it staged had a title, so all three tests took the *enumerated* branch and the probe was only ever hit with a dead handle. The live-but-filtered path — the entire point of Round 1's P1-2 fix — was still verified against injected fakes only, leaving the real question (do `GetWindowRect` / `IsWindowVisible` behave as assumed for an untitled window?) unanswered. Now there is a fixture that spawns a real visible, borderless, **untitled** window, located by process id because the title-based enumeration cannot see it. The test asserts the premise (absent from the enumeration) before checking its elements are still compared against its real rect, so it cannot pass vacuously.

**P3 set:**
- one enumeration snapshot now serves both lookups on the miss path — two snapshots could disagree about the same desktop
- HWND match beating an identical window title is now documented as deliberate
- corrected the doc claim that the gate passes conservatively on *any* Win32 error: the per-HWND probe reports a stale view instead, which costs a re-discovery rather than a wrong click
- the wiring pin is anchored to line start so a commented-out line cannot satisfy it
- the two remaining historical documents carrying superseded recovery advice / the old gate API now say so

**Recorded in the ADR instead of fixed here** (Harusame64/desktop-touch-mcp-internal#33, `ddde570`):
- **R9** — `origin_window_not_visible` tells you to restore the window by title, which an untitled window does not have. Only bites where the two cases intersect (hidden **and** untitled).
- **R10** — a recycled HWND makes any origin comparison point at an unrelated window. Pre-existing (the foreground comparison had the same weakness) and bounded by the lease TTL; the durable fix is a composite window identity.

**`describe.skipIf`** on the e2e is kept: it is the project's convention for live-desktop staging, and every staged precondition inside the tests is asserted rather than skipped.

Full suite: unit green. Unrelated failures — `browser-cdp-verification` (headless Chrome rescue classification; no file from this diff, main CI green), and `ui-pattern-store-persistence` / `terminal-console-paste-load`, both of which pass on individual re-run.

@codex review

Harusame64:
## Round 3 findings applied (`0160d13`) — both Codex P1s

Both were the same shape: a recovery path that quietly reintroduced the failure this change removes.

**P1 — title-based origins were treated as valid anywhere on screen.** Discovery with a plain `windowTitle` passes no handle to the providers, so the recorded origin is the *title* — the common case, not an edge case. Checking only the virtual screen meant a stale element stayed clickable at its old coordinates after its window moved, hid or closed, landing the click on whatever occupies that area now. The origin is now resolved in identity order — HWND → title → direct probe of a handle the enumeration filtered out — and same-titled windows pass when any visible one still covers the element. Only `"@active"` (no identity was available at discovery) and browser tabs fall back to the virtual screen.

**P1 — the error recommended `browser_click` as monitor-independent. It is not.** All three of its paths funnel into one OS-cursor click; CDP only resolves the coordinates and verifies delivery afterwards. On a browser window on a secondary monitor — the exact scenario raising this error — following that advice would have reproduced the silent clamped click the guard exists to prevent, and a clamped click that happened to mutate a DOM node would even have been reported back as `delivered`.

I deferred guarding `browser_click` in Round 1 on the grounds that it needed its own blast-radius assessment. **That premise did not survive inspection**: the funnel is a single function, and the surrounding `catch → failWith` already shapes a typed error into the same code and suggestions — one line, no new envelope work. So it is guarded here rather than deferred, and every surface that named it as an escape hatch now says the opposite. `click_element` (accessibility invoke, no cursor) is the one route that genuinely works on any monitor; the CHANGELOG and both READMEs say so.

**Structural follow-up:** manual enumeration of cursor-moving paths missed `browser_click` once — in a review that specifically asked for that enumeration. So it is a test now: any source file that moves the nut.js cursor must also assert reachability. That is what would have caught this.

The earlier Codex P2 (hidden origin windows classified as closed) was already fixed in Round 1 by the direct HWND probe.

ADR updated (Harusame64/desktop-touch-mcp-internal#33): R8 is closed rather than carried forward — what remains for Phase 2 is capability (reaching a secondary monitor at all), not safety.

Full suite: unit green (5039). Unrelated failures — `browser-cdp-verification` (headless Chrome rescue classification, no file from this diff, main CI green) and `desktop-facade` (10s timeout under full-suite load; 75/75 pass individually).

@codex review

Harusame64:
## Round 3 (Opus) findings applied (`269ad39`) — 2 P1

**P1 — the title resolution I added for Codex's P1 was itself a 100%-block regression.** I matched titles with equality, but `windowTitle` is a case-insensitive **substring query** everywhere else in this codebase, and the recorded origin is the raw query, not a resolved full title: `desktop_discover({target:{windowTitle:"Notepad"}})` records `"Notepad"` while the live title is `"Untitled - Notepad"`. Equality missed, and every visual-only element discovered that way — the most common shape for accessibility-blind apps — was blocked. That is precisely the symptom this PR exists to remove, reintroduced by the fix for it. Titles are now matched the way the discovery path matches them (case-insensitive substring, first hit in Z-order).

Resolving to **one** window rather than "any match that covers the element" is deliberate and load-bearing under substring matching: with a short query like `"Chrome"` many windows match, and accepting any of them would drift back toward the containment check this design rejected as tautological. The topmost drawn match is the window discovery would resolve the same query to now.

**Why nothing caught it:** every unit fixture used a byte-exact full title, and the e2e built HWND origins by hand. Both are fixed — the unit suite pins the substring case, and the e2e now exercises the title lane against a real window using a **prefix of its title**, the way a user would type it.

**P1 — one surface still advertised `browser_click`** as the escape hatch (`docs/system-overview.md`). Corrected; it is refused by the same guard.

**P3** — documented that the cursor-move invariant is file-level and therefore coarse: it catches a *new cursor-moving file*, which is the miss that actually happened, not a new unguarded call site inside a file that already guards elsewhere, and it does not see an aliased import.

Also checked: `terminal-provider`'s `origin.id` cannot be undefined (guarded by an early return above it), so the typed-shape concern does not bite.

Full suite: unit green. Unrelated failures — `browser-cdp-verification` (headless Chrome rescue classification; unchanged by the browser guard added in the previous commit — same assertion, same value, and no `CoordinateOutsideReachableBounds` anywhere in the run) and `terminal-console-paste-load` (live-desktop load test).

@codex review

Harusame64:
## Round 4 findings applied (`d4b569c`) — P1 + P2, fixed structurally

**P1 — resolving the title origin by "first match that isn't minimised or cloaked" retargeted the gate to a window discovery never picked.** Discovery takes the topmost substring match *whatever its state*; skipping a minimised or cloaked match to reach the next one lets a second window that merely shares the substring become the comparison rect, and a stale coordinate inside it passes — a click in the wrong window. With a query like `"Chrome"` and a virtual-desktop switch (which cloaks windows) that is ordinary, not exotic. Correct verdict there is `origin_window_not_visible`: **resolve first, judge that window's visibility second.**

**P2 — the gate did not mirror discovery's exclusion of dialog-class and owned windows**, so an owned dialog whose title contains the query and sits above its owner became the comparison rect.

Both are one failure mode: **the gate matching by a slightly different rule than the discovery it is checking.** That has now bitten three times in this PR — equality vs substring, visibility skipping, dialog filtering — so this round fixes it structurally rather than with another one-off correction. `pickPlainTopLevelWindowByTitle` is the predicate half of the existing resolver helper, applied to a window list the caller already has; both the discovery path and the gate call it with the same flags. The rule cannot drift again without changing one shared function.

The first cut of that refactor broke a contract its own test pins — `findPlainTopLevelWindowByTitle("")` must return null *without enumerating* — caught by `find-plain-top-level-window.test.ts`. The early return is back.

New tests: the minimised-match case now pins "does not retarget", and a dialog + owned-popup fixture pins the exclusion.

Full suite: unit green. Unrelated failures — `browser-cdp-verification` (headless Chrome rescue classification), plus `replay-backend` and `terminal-hidden-input`, both timing-sensitive and passing individually (13/13 and full file respectively).

@codex review

Harusame64:
## Round 5 findings applied (`de3faf2`) — P1

**P1 — the gate was mirroring the wrong matcher.** Round 4 made it follow `resolveWindowTarget`'s Case 3 flags, but that function is not what picks the window a visual-only entity comes from: it only *probes* for a match and then lets the raw query flow through to the providers. The window whose pixels became the entity is chosen inside the OCR pipeline, by a plain case-insensitive substring find over the Z-ordered list — **no** minimised, dialog or owned filtering. So excluding dialogs and owned windows in the gate skipped a dialog that OCR had actually captured and compared against its owner instead: a false block, or a pass against the wrong window.

The gate now uses that unfiltered rule, and the comment names the real authority (`engine/ocr-bridge.ts`) so the next reader does not have to rediscover it.

This is the third variant of one mistake — the gate matching by a slightly different rule than the code it is checking. Worth noting that the shared predicate from Round 4 is what made this round a **flag change** rather than another reimplementation; the structural fix paid off immediately.

**P2 — ADR/implementation drift.** §7.1.2 still described the title lane as falling back to the virtual screen and named the wrong matcher; §7.3-1 branch (b) was worded on that basis. Both are synced (Harusame64/desktop-touch-mcp-internal#33, `dcfb239`), and the Phase 1 checklist is now ticked off with what actually shipped. The durable fix stays recorded as OQ7: have the pipeline return the handle it already resolved and the title lane disappears entirely.

Full suite: unit green. Unrelated failures — `browser-cdp-verification` (headless Chrome rescue classification) and `terminal-console-paste-load` (live-desktop load test).

@codex review

Harusame64:
## Round 6 (Codex) findings applied (`b3d5680`) — the durable fix, not another patch

Codex found two more ways the title lane can click the wrong window, and both have the same root: **the gate re-resolves a query at act time instead of checking an identity.**

- Several windows match the substring → a Z-order change between discovery and act makes the same predicate pick a *different* window, and a stale coordinate inside that replacement passes.
- An entity discovered by explicit handle whose window has since closed could be matched against an unrelated live window whose title happens to contain the same digits.

Both disappear once the handle is recorded at discovery — the fix the ADR already named as OQ7. The capture pipeline resolves that handle anyway, so it now returns it; the OCR and visual lanes stamp it on their candidates; the resolver carries it to `UiEntity.origin.hwnd`; and the gate prefers it and, having one, **never** falls back to title matching.

**Why a field beside `target` rather than replacing `target.id`** (this was the design question, and Fable's read of the code decided it): `target` feeds `candidateKey`, so replacing it shifts every entityId and evidence digest — and the fold path's carry-forward depends on the discover lane and the post-touch lane agreeing on that id, so parity would break and entities would read as *disappeared*. The new field is outside the key: identity, leases and the fold are untouched, pinned by a resolver test asserting the entityId is unchanged. The fold's carried-forward candidates get the same handle too, since that path already knows the window it verified.

**Occlusion (the other P1) is deliberately not fixed here.** The finding is correct — mouse input is delivered by hit-test and this route does not focus first, so a covering window receives the click — but the class **predates this change**: the old gate equally passed a background-origin entity that the foreground window covered. A rectangle-based check cannot see per-pixel alpha or non-rectangular windows, so it would produce exactly the **false blocks this PR exists to remove**. The correct fix is a native hit-test, which belongs with the native mouse work in Phase 2; the fold's repaint verification already gives an after-the-fact signal. Recorded as R11 with that plan (Harusame64/desktop-touch-mcp-internal#33, `6d746d3`), along with OQ7's resolution.

Full suite: **5048 pass**. Sole failure is `browser-cdp-verification` (headless Chrome rescue classification) — unrelated and unchanged throughout this PR.

@codex review

Harusame64:
## Round 7 applied (`fc36455`) — and the occlusion finding, answered with evidence

**Opus Round 7: P1 and P2 empty.** It verified the identity/parity question directly — `originHwnd` is outside `candidateKey`, so entityId, evidence digest, lease validation and the fold's pre/post parity are untouched, and it could not construct a divergent case. Two P3s fixed: a browser-tab target no longer picks up a window handle (only a window-kind candidate has a rectangle worth judging), and the ADR now names the right remaining un-stamped lane.

### On `Reject coordinates occluded by higher Z-order windows`

This is the one finding I am **deliberately not implementing in this PR**, and it has now been checked against the real desktop rather than argued from first principles. An enumeration-plus-rectangle occlusion check would **miss the true occluder and fire on a harmless one**:

- **False negative:** the Key Locker's topmost dialog — which genuinely takes clicks — is removed from `enumWindowsInZOrder` by the tool exclusion. So is the untitled topmost window this PR's own e2e helper spawns (untitled filter). The real occluders are not even in the scan's input.
- **False positive:** with auto-dock configured, the server itself pins the host terminal window on top (`pin` defaults to true). That is this user's normal setup, so every visual element underneath would be blocked — reintroducing the exact 100%-block symptom this PR removes.
- **Precedent:** the modal sensor already tried "windows above the target in Z-order" and needed a hand-maintained allowlist of eight system-resident classes (CoreWindow/IME, ApplicationFrameWindow, InputSite, Shell_TrayWnd, …) to suppress false positives. Doing it here means growing that list from dogfood reports — not one PR's scope.
- The class also **predates this change**: the old gate equally passed a background-origin entity that the foreground window covered.

The correct mechanism is **focus-then-verify** (reusing the existing focus ladder), with a native hit-test as confirmation — no Win32 hit-test exists in the repo yet, so it is a new native surface that belongs with the native mouse work. It also sidesteps `GetWindowRect` including the invisible DWM border. Recorded as R11 with that plan; the fold's repaint verification already gives an after-the-fact signal today.

Full suite: 5048 pass; sole failure `browser-cdp-verification` (headless Chrome rescue classification), unrelated and unchanged throughout.

Handing to the user for final review — the deferral above is theirs to confirm.

PR #549, opened 2026-07-25
feat(mouse): reach every monitor with a native cursor path (ADR-029 Phase 2a)

Harusame64:
Implements **ADR-029 Phase 2a — the native cursor path**, the phase that actually restores the original symptom: a visual-only element on a second monitor could not be acted on. Phase 1 (#548) only made the failure safe and typed.

Plan (private repo): `desktop-touch-mcp-internal@12d89a6:docs/adr-029-phase2a-plan.md` — six review rounds (Opus + Codex) before implementation started.

## What changes

**`src/win32/mouse.rs` (new).** `SetCursorPos` places the cursor in physical coordinates, `GetCursorPos` reads it back, and only a mismatch falls through to `SendInput` with `MOUSEEVENTF_VIRTUALDESK` — the flag libnut omits, which is why every non-primary monitor was unreachable. A click hit-tests wherever the cursor is, so exact placement is the entire job; the 0..65535 mapping used by the correction path is a pure function with unit tests over the negative-origin case.

**`src/engine/cursor.ts` (new) — the only file that moves the cursor.** It owns the guard, the native/fallback decision and the interpolation. Interpolation stays in TS because a Rust sleep loop would hold the libuv thread for the whole gesture; points are planned by distance (1px, as nut.js does) so applications watching the pointer stream see the same gesture, and each tick ships its batch in one napi call rather than one per pixel.

**Reachable bounds widen by capability.** Every monitor when the native path is present, the primary monitor when it is not — widening unconditionally would restore the silent misclick on a build that cannot actually reach the other monitors. Containment is per monitor, not the virtual screen's bounding box: a point in the gap of a staggered layout is in the box but on no monitor, and Windows pulls it to the nearest one.

**New typed error `CursorPlacementBlocked`** (`desktop_act` reports `reason:"cursor_placement_blocked"`) for a point that is on a monitor but could not be reached — another app holding the cursor with `ClipCursor`, a disconnected or locked session, another program repositioning the pointer, or a monitor added or removed mid-move. Calling that "outside the reachable bounds" would be untrue and none of that error's recovery applies.

**The drag path gained a `try/finally`.** The move can now throw, and an exception between press and release would leave the left button held down at OS level.

## Review notes

- **Not covered, deliberately:** with an elevated window in front, UIPI discards the click but not `SetCursorPos`, so the move succeeds and verifies — that failure stays silent. Out of scope for this phase, recorded in the ADR.
- **Path planning skips monitor gaps** rather than walking through them; with a button held those clipped intermediate points would fire `dragover` and drop-target highlighting at positions nobody asked for.
- **CHANGELOG rewrites the Phase 1 entry** instead of appending: both phases ship in one release, so users never see the primary-monitor-only behaviour, and leaving that entry would describe a state that never shipped.
- The invariant test is now "only `cursor.ts` may move the cursor", with a separate allowlist for declaring the bindings versus calling them.

## Verification

- `cargo test -p desktop-touch-engine` — 6 new tests over the coordinate mapping
- Unit: `cursor.test.ts` (path planning, gap skipping, exactness), `reachable-bounds.test.ts` (per-monitor containment, both message variants, classification and ordering pins for both codes), promotion test in `guarded-touch.test.ts`, `FAIL_REASONS` updated
- Full suite: 5054 passed. Four files fail and all four fail the same way on a clean tree: `browser-cdp-verification` (verified by stashing), `context-consistency` (the known `hasModal` over-reporting), `terminal-console-paste-load` (timing/stress), and `ui-pattern-store-persistence` (passes in isolation, flakes under full-suite load)
- **Live desktop, freshly built addon:** exact landing via `set_cursor_pos`, batch path, and `ok:false` carrying the clamped position for an unreachable point
- `tests/e2e/adr-029-cursor-move.test.ts` skips until the addon is rebuilt (the running MCP server holds the `.node`), so CI is where it first executes

**Still needs dogfood:** this machine currently reports a single monitor, so the negative-coordinate case — the original symptom — has not been exercised on real hardware. Checklist for a two-monitor session: `mouse_click` and `mouse_drag` onto the left monitor, `desktop_act` on a visual-only entity there, an RDP session, and whether `get_cursor_position` reports negative coordinates correctly (libnut's read path is untouched and its behaviour there is unverified).


Harusame64:
@codex review

Harusame64:
@codex review

Round 1 findings applied in `72ea0d2`. Your P1 (read-back failure fabricating a successful position) matched the Opus review independently and is fixed: an unreadable position is now `ok:false` with `method:"readback_failed"`, the `SendInput` return value is no longer discarded, and the TypeScript side raises `CursorPlacementBlocked` instead of letting a bare error fold into `executor_failed`.

Please re-review the new commit. Areas worth attention: the five `method` values and whether any path still reports success without verifying; the 250ms monitor-layout memo (does it widen the window where the guard and the move disagree?); and whether the invariant test still holds now that the binding-declaration file is exempt from its runtime rule.

PR #550, opened 2026-07-25
fix(lint): let lint see unused imports in tests, and clear what it finds

Harusame64:
Closes the two CodeQL alerts and the Dependabot alert raised against v1.13.0, and removes the reason lint could not have caught the first two.

## Why lint missed it

`eslint.config.mjs` turned `@typescript-eslint/no-unused-vars` **off** for `scripts/` and `tests/`. So a stale import in a test was invisible to `npm run lint` and only surfaced later as a CodeQL alert — which is exactly how #153 and #154 reached `main`: lint reported zero errors on the very files that carried them.

The blanket "off" is really about the argument and caught-error shapes that tests produce, so those are exempted individually (`args: "none"`, `caughtErrors: "none"`, `^_` patterns) and unused variables and imports are an error again.

## What that surfaced

14 existing violations, all stale imports or one unused destructured binding, removed here:

| File | Symbol |
|---|---|
| `tests/e2e/browser-cdp.test.ts` | `readFileSync` |
| `tests/e2e/wait-until.test.ts` | `sleep` |
| `tests/e2e/workspace-chain.test.ts` | `beforeAll`, `launchNotepad`, `sleep`, `enumWindowsInZOrder` |
| `tests/unit/desktop-facade.test.ts` | `DesktopSeeInput` |
| `tests/unit/desktop-state-causal-include.test.ts` | `CommitL1Emitter`, `ToolCallEvent` |
| `tests/unit/homing-snapshot-delta.test.ts` | `nutjs` ← CodeQL #154 |
| `tests/unit/homing-snapshot-integration.test.ts` | `nutjs` ← CodeQL #153 |
| `tests/unit/path-class-contract/e-uia-fallback-ladder.test.ts` | `ExecutorKind` |
| `tests/unit/session-registry.test.ts` | `TargetSpec` |
| `tests/unit/working-memory-b1.test.ts` | `WorkingMemoryProjection` |

The two `nutjs` imports are mine: leftovers from moving the homing tests' spy to the cursor choke point in #549.

## Also here

- **Dependabot #34 (postcss, high)** — pinned to `^8.5.18` through `overrides`, resolving to 8.5.23. It arrives via `vitest → vite`, is dev-only, and ships in neither the npm package nor the release zip, so this is hygiene rather than exposure.
- **`docs/release-process.md`** claimed `npx ... --help` prints a version banner and usage. Neither the launcher nor `src/` implements `--help`, and 1.12.4 behaves identically — the step silently proved nothing during the 1.13.0 release. The doc now states what that command does demonstrate (download, SHA verification, extraction) and points at the stdio initialize smoke test as the check to trust.

## Verification

Unit suite green (293 files, 4716 tests). `npx eslint src/ tests/ scripts/` reports 0 errors. No production code is touched.

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -91,9 +91,13 @@ Run:
 ```bash
 node --check bin/launcher.js
 npm run check:launcher-manifest
+npm run lint            # stale eslint-disable directives are errors, so CI would fail on them
 npm run build
 npm publish --dry-run
 ```
+
+`npm run lint` is here so the guard is local: `.githooks/pre-push` only blocks
+direct pushes to `main`, and lint otherwise runs for the first time in CI.
 
 ### Dogfood Pass (Required, v1.3 lesson)
 
@@ -454,7 +458,7 @@ wrong version, etc.) that the stdio smoke test above cannot detect.
 rm -rf "$USERPROFILE/.desktop-touch-mcp/releases/vX.Y.Z"
 npm cache clean --force
 
-# 2. Download via npx and verify --help
+# 2. Download via npx (proves download + SHA verification + extraction)
 #    ⚠️ Run this from a cwd OUTSIDE the project source dir first
 #    (PowerShell: `cd $env:TEMP`  |  bash: `cd /tmp`).
 #    The project's own package.json has name=@harusame64/desktop-touch-mcp; once its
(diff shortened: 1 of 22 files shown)


Harusame64:
@codex review

Lint-scope change plus a dependency follow-up. Three things worth focused attention:

1. **Scope of the config change.** `eslint.config.mjs` now lets lint see unused imports in `tests/**`. Please check the override does not weaken anything for `src/**`, and that the ordering of the flat-config blocks makes the intended rule win where the globs overlap.

2. **The clearing edits.** Ten test files lose an import or an `eslint-disable`. Each one should be dead, not load-bearing: a side-effect import, a binding used only in a type position, or a disable that suppresses a rule under a different environment would all look unused here and break silently. Flag any you cannot prove is dead.

3. **Dependency override.** `package.json` gains `"postcss": "^8.5.18"` in overrides with the lockfile at 8.5.23 (Dependabot #34, dev-only via vitest → vite), and nanoid moves to 3.3.16. Please check the override cannot conflict with what vite actually requires, and that no other transitive copy of postcss remains pinned below the patched version.

Harusame64:
@codex review

`0cba01b` answers the round-1 review of this PR: `linterOptions.reportUnusedDisableDirectives: "error"` plus a sweep of the 15 stale directives that were passing CI as warnings (`npm run lint` is bare `eslint .`, so nothing warning-level ever failed). Worth checking:

1. Whether promoting unused-disable to `error` can fire on a directive that is load-bearing in an environment CI does not exercise — a rule that is off for `tests/**` but on elsewhere, or a type-aware rule that reports only with a project service.
2. The `docs/release-process.md` `--help` step: it now documents that stdin must be at EOF (`< /dev/null` / `cmd /c "... < NUL"`), because `wireLauncherStdio` ends the child's stdin only when the parent's ends, so the interactive form never returns. Please sanity-check that claim against `bin/launcher.js`.

Harusame64:
@codex review

`d7e03f3` — you were right on both counts, and the previous commit made it worse rather than better.

Measured: `src/server-windows.ts:527` handles `--help` / `-h` before the stdio transport starts and exits 0, `bin/launcher.js:410` forwards argv, and `node dist/index.js --help` prints `desktop-touch-mcp v1.13.0` plus the usage block and exits 0 — no stdin redirect involved. So both the pre-existing "no usage text" expectation and the stdin claim I added on top of it are gone. The step now asserts the usage output and the version matching X.Y.Z, and says that silence or a version mismatch is a regression in argument forwarding, extraction, or startup. The label and the closing sentence were adjusted to match what the step now proves.

PR #551, opened 2026-07-25
fix(errors): route recovery advice to the root suggest (ADR-029 OQ8)

Harusame64:
Fixes ADR-029 OQ8, found while reviewing #549: recovery advice written at the call site was landing where nothing reads it.

## The defect

`failWith(err, toolName, context)` takes a context **record**, not an options bag. Everything in it except the three `ROOT_HOISTED_KEYS` is nested under `context`, and the root `suggest` is whatever `classify(message)` produced. So `failWith(err, "mouse_drag", { suggest: [...] })` files the advice at `context.suggest`.

Fourteen call sites had drifted into that shape. Nine were harmless duplicates — `classify` supplied good advice at the root anyway. **Five were not**: they classify to the generic `ToolError`, which has no `SUGGESTS` entry, so the failure shipped with *no* advice in the place the server instructions tell the model to read.

| Site | Code buried in the message | Root `suggest` before |
|---|---|---|
| `keyboard.ts:812` | `ForegroundFlashRequiresTarget` | 0 |
| `keyboard.ts:851` | `ForegroundFlashUnsupported` | 0 |
| `mouse.ts:888` | `TabDragBlocked` | 0 |
| `mouse.ts:908` | `CrossWindowDragBlocked` | 0 |
| `terminal.ts:1179` | `ForegroundFlashUnsupported` | 0 |

## The fix

The same move `SpawnFailed` made in this file: classify the code the producer already names in its message, and let `SUGGESTS` carry the advice. Call sites just stop passing it. The new arms sit after the longer `ForegroundFlashNotApplicableTo*` ones so the shorter stem cannot poach them; a test pins that ordering.

**Rejected alternative:** adding `suggest` to `ROOT_HOISTED_KEYS`. The presenter spreads `rootExtras` last, so a hand-written line would *override* the dictionary entry — a regression for the nine sites where `classify` already supplies richer advice (`WindowNotFound` has five).

## What the new lint rule then found

The rule added to stop this recurring flagged a second defect of the same family: **23 sites pass `{ context: { … } }` to that argument**, which renders as `context.context.x`. The codebase already knew about this trap — `keyboard.ts` comments and two regression tests state the keys belong flat — but only some sites had been converted. All 23 are flattened here (recounted in round 3 — the first count of 21 was low), so callers read `context.x` uniformly.

The rule is scoped to `src/**` minus `_errors.ts`, and bans both keys in the third argument of `failWith` / `errorFromMessage`, pointing at `SUGGESTS` or `failCode` instead.

## Verification

Envelopes rendered before and after: the five sites go from `code:"ToolError", suggest:0` to their own code with 2–3 root suggestions, and `context` comes back one level deep. Ordering pins hold for all four foreground-flash codes.

Unit suite green (293 files, 4716 tests). ESLint 0 errors. `failwith-callsite-shapes.json` regenerated.

## Note on the wire

`code` changes from `ToolError` to a specific code at those five sites, and `context.context.x` becomes `context.x` at the 23. Both are the direction ADR-021 set (typed codes, single-nest context) and nothing in docs or README documents the old shapes as contract.

## Review rounds

**Round 2** (`3e8cee8`): `keyboard` `method:'background'` now returns `WindowNotFound` when no window matched the title (the press-side guard is compound, so its else branch splits on `!target`); `ForegroundFlashFailed` became a typed code — the sixth hole of this bug class, in the files this PR already touches; `SUGGESTS.BackgroundInputIncomplete` regained the elevation line the call-site cleanup had dropped.

**Round 3** (`3cd08ee`, `cf6ab50`): the lint rule resolves its callee back to an `_errors` import binding, so an aliased import cannot slip past, and peels TS casts around the third argument — both were real escapes. `SUGGESTS.ForegroundFlashFailed` is now keyed to `context.reason`: two of the eight native reasons are pre-flight rejects where nothing was sent, and `foreground_restore_failed` happens after the paste succeeded, so the old blanket "failed partway — retry once" was actively wrong for three of them. `elevated_target` was cited as an example reason in three places and is produced by no code path. The flattened-site count was 21 and is 23. The `_errors.ts` lint exemption rested on a false rationale and is gone.


Harusame64:
@codex review

Production-code change to the failure envelope path. Two things worth focused attention:

1. **classify ordering.** Four new arms are added after the longer `ForegroundFlashNotApplicableTo*` ones because all four codes share the `ForegroundFlash` stem. Please check the whole cascade for a substring collision I may have missed, in either direction.

2. **21 call sites flattened.** `{ context: { … } }` in `failWith`'s third argument renders as `context.context.x`; those are now flat. If any consumer — production, test, or docs — still reads the doubled shape, it would break silently.

Also worth a look: the new ESLint rule's AST matching (alias imports, re-exports), and whether the merged `ForegroundFlashUnsupported` suggestion reads correctly for both the keyboard and terminal producers, whose original advice differed.

Harusame64:
@codex review

Round 2 answered your P2 (alias tracking) with documentation only; `3cd08ee` now closes it properly.

Callee matching is the union of the textual name and ESLint scope resolution back to a binding imported from `_errors`, so `import { failWith as f }` and a local `const g = f` re-alias are caught at any scope depth, with the alias chain depth-bounded (`const f = f` terminates). The name signal is kept on purpose — a same-named helper from an unrelated module is still flagged, which the header documents as a deliberate over-approximation for a guard rule. A computed literal key (`["suggest"]`) is flagged too; a computed identifier is not guessed.

New: `tests/unit/eslint-no-suggest-in-failwith-context.test.ts`, 13 cases through the ESLint `Linter` API covering both directions plus the termination guard.

Worth focused attention:

1. **Scope resolution correctness.** `resolvesToHelper` reads `variable.defs` and treats `def.type === "ImportBinding"` with `def.parent.source.value` matching `/(^|[\/])_errors(\.js|\.ts)?$/`. Please check the def-shape assumptions hold under the TS parser (the repo lints `.ts` via typescript-eslint, while the new tests drive plain `.js` fixtures), and that walking `scope.upper` for the variable cannot resolve the wrong binding when a local shadows an import.

2. **Anything still reachable.** Third argument as a variable or a spread is knowingly out of scope. If you see another shape in `src/` that reaches these helpers and is not matched, name it.

Also still open from round 2, if you want a second opinion: `ForegroundFlashFailed` became a typed code in this PR rather than being deferred, and its classify arm is placed before the generic arms because reasons like `focus_wait_timeout` would otherwise be poached by the UiaTimeout arm.

Harusame64:
@codex review

`bd4769a` — round 4 landed. The advice for `ForegroundFlashFailed` was wrong for three of the reasons and I have rewritten it against the native path; this is the part worth an independent trace:

1. **`send_input_failed` is emitted at two steps** — the Ctrl+V of step 6 (nothing pasted) and the Enter of step 8 (pasted, Enter missing, and `pressEnter` defaults to true for `terminal:send`) — plus the addon-missing shortcut in `bg-input.ts` (nothing sent). The reason cannot distinguish them, so the advice now says the state is ambiguous and to read the target rather than resending. Please confirm there is no fourth producer.
2. **`foreground_restore_failed` can mask an intercepted paste**: `inner?` propagates before the `paste_warning_detected` check, so "the text was pasted" is not unconditional. Check that reading of the control flow.
3. **`context.reason` value range**: the eight `ForegroundFlashErrorReason` values, plus the four from `ClipboardError::as_reason()`, plus any unknown native message passed through raw by `KNOWN_FLASH_REASONS`. If there is a fifth source I missed, name it.

Also in this round: `wt_paste_warning_intercepted` no longer advises "send fewer lines" (`validate_input` rejects every newline, so the input is always single-line) and warns the dialog may still be up (`escape_sent` is discarded); `pressEnter` is no longer suggested for `keyboard:type`, which hardcodes it false; `BackgroundInputIncomplete`'s retry line is scoped to the tools where retrying does not replay a combo; and the matrix's stale 35-entry SUGGESTS list is replaced by the dictionary as SSOT plus a grep recipe.

Harusame64:
@codex review

`395140d` acts on your last round. Thank you for the `WindowNotFound` catch — six review rounds had not seen it, and it was a regression this PR itself introduced.

Rather than add a matching arm per message, `classify` now ends with a last-resort step before the generic fallback: if the message's leading token is a key in `SUGGESTS`, that is the code. It runs after every ordering-sensitive arm, prose cannot trigger it (prose does not begin with a bare code token), and it cannot invent a code. A sweep of `src/` found one other compact producer, `ForegroundFlashChannelNotImplemented`, which had no dictionary entry at all — it has one now. `tests/unit/oq8-failwith-suggest-routing.test.ts` derives the compact-producer list from the source and asserts each routes to its own code with non-empty advice, so this cannot silently reopen.

Worth your attention:

1. **Does the leading-token step mis-route anything?** Prefix-related families are the risk: `ForegroundFlash*`, `Terminal*` vs `Window*`, `KeyLocker*`. If a compact message exists (or could plausibly be written) whose correct code is not its leading token, that is a defect.
2. **Is the structural test's sweep complete?** It matches `new Error("<Pascal>")`. Other compact shapes — a template literal, a variable, an error class doing `super("<Code>")` — would escape it. `key-locker-manager.ts` uses `super("<code>: …")`, i.e. prose, so I believe it is covered by the prose arms, but please check for shapes I missed.
3. **`no_supported_channel` removal.** I read `reasonFromInjectCheck` plus `_check` as always carrying one of `chromium` / `uwp_sandboxed` / `class_unknown` on this path, which is why the advice no longer lists it. If it is reachable after all, the advice is now incomplete.

Also applied from your round: the `pressEnter:false` caveat for chunked `terminal(action:'send')` retries, and the rule now follows a namespace member captured into a variable (`const f = errors.failWith`).

Harusame64:
@codex review

`af38578` (plus `395140d`, which answered your `WindowNotFound` finding) is ready for another pass.

Your finding led to a wider check: rendering all 81 `SUGGESTS` keys through `classify` in both shapes a producer can emit (bare code, and `<Code>: detail`) showed exactly two that did not round-trip — `BrowserSearchTimeout` was poached by the `UiaTimeout` arm and `KeyLockerSpawnFailed` by the `spawnfailed` arm. Both are latent today because their producers go through `failCode`, and both would surface the moment a producer switched to `failWith`. They now have their own arms ahead of the generic ones, and a source-derived test asserts the whole round-trip invariant with no allow-list.

Two things worth your judgment:

1. **`ForegroundFlashChannelNotImplemented` is now documented as reserved, not emitted.** The `wm_char` branch always returns and `unsupported` returns earlier, so the reject is reachable only with `kind === "cooperative_bridge"`, which the resolver never constructs. It is out of the CHANGELOG and into the matrix §5.2 reserved table; the dictionary entry stays, per that table's pre-registration policy. If you can find a path that does reach it, that changes the classification.

2. **The producer-pin test gained a pattern, not an exemption.** `new \w+Error\("<code>"` — a typed error class taking the code as its first constructor argument, which `KeyLockerError` does. Before that, `KeyLockerSpawnFailed` was reported as a dead classify branch despite seven real producers. Verified that the change moves exactly one entry: the remaining 13 dead entries match the existing allow-lists, and no allow-list was touched. Please check the pattern cannot mask a genuinely producer-less code.

Also applied: the `no_supported_channel` removal is now consistent across `_errors.ts`, `CHANGELOG.md` and the matrix (it was a three-way desync); the compact-code CHANGELOG item is its own top-level bullet; and the src-walking guard tests carry an explicit timeout.

Harusame64:
@codex review

`786edef` answers your round-8 finding, and the fix is wider than the one arm you cited.

Measured first. `WindowNotFound: hwnd "timeout" is not a valid integer` (`src/tools/_resolve-window.ts:233`, where the hwnd string is caller-typed) classified as `UiaTimeout` and returned "the target app may be unresponsive — wait and retry" — advice that cannot fix a malformed argument. Feeding every `SUGGESTS` key through `classify` as `<Key>: <generic keyword>` showed 77 of 81 keys can lose their own prefix this way, so this is a class, not an instance — the same class the `BrowserSearchTimeout` and `KeyLockerSpawnFailed` arms each patched one at a time in earlier rounds.

Rather than add a third one-off arm, a strict declared-code arm now sits at the TOP of the cascade: PascalCase token, immediate colon, exact-case match against a registered `SUGGESTS` key. A declaration therefore resolves before any detail is scanned.

Two things worth your independent check:

1. **Blast radius of moving a match to the front.** I enumerated every `new Error("<Pascal>: …")` producer in `src/` (9 distinct prefixes) and diffed classification before and after. Only shapes with an adversarial detail change; benign details were already resolved identically by the trailing leading-code arm. `Blocked:` and `CDP:` are not `SUGGESTS` keys, so they correctly still fall through. Please check whether a `<Key>:` prefix can ever be the *wrong* answer — a message where a more specific arm below should legitimately win over the declared prefix.

2. **What the strictness leaves behind.** Non-leading shapes still reach the substring cascade: a wrapper that prepends its own prefix, and the `AutoGuardBlocked[endpoint]:` bracket variant (`mouse.ts:746`). So the ordering-sensitive arms below stay load-bearing, and I rewrote their comments — plus the round-6 rationale for keeping the bare-code arm LAST — to say which shape each one now defends. Please check those comments describe the code that is actually there.

The dictionary-wide invariant now feeds one worst-case adversarial tail per key rather than the benign tail it used before; that benign tail is why six rounds did not surface this.

Harusame64:
@codex review

New HEAD `e20fcd0` — please review this commit rather than `786edef`. An Opus pass on round 8 found a P1 that neither of us caught, and it is worth your attention because it is a failure mode the earlier rounds created.

**Round 8 silently disarmed the tests that guard it.** The declared-code arm returns before `message.toLowerCase()` is ever computed. Every existing ordering pin fed a leading `<Code>: …` string — so those pins now pass through the new arm and no longer exercise the cascade ordering they claim to protect. `tests/unit/phase7-f3-spawn-failed-typed-code.test.ts` case #6 asserted that `SpawnFailed` beats `WindowNotFound`, with a comment calling itself "the only defense layer"; after round 8 the SpawnFailed arm could be moved below WindowNotFound and it would still pass.

A sweep of every test feeding a `<Code>:` string found six more in the same state — four vacuous ordering pins (`reachable-bounds` × 2, `keyboard-input-serialization` Pin 2, `key-locker-errors-classify`) and two whose comments described a substring match that no longer happens. All are restored here by feeding a shape that actually reaches the cascade (a lowercase wrapper prefix), keeping the leading form alongside as the end-to-end check.

Three things worth your independent check:

1. **Are the restored pins really non-vacuous?** Each now depends on the wrapper-prefixed input missing the declared arm. Verified: the declared regex does not match any of them and the cascade is what returns the expected code. Please look for a pin I have restored in form but not in force.

2. **A rationale that named a shape with no producer.** Round 8's comments justified the ordering-sensitive arms as defending "wrapper prefix" messages. Grepping `src/` for that shape found only two prefix compositions — `cdp-bridge.ts:136` (`CDP:`, not a registered code, so the cascade genuinely decides) and `launch.ts::spawnDetached` (leading form, resolved by the declared arm). So no producer emits a registered code mid-message today. The comments now say that plainly and call the arms defense-in-depth rather than implying a live caller.

3. **The tail arm's colon branch was dead.** The bare-code arm at the end of `classify` matched `(?::|\s|$)`, but the declared arm at the top is a strict superset of the colon case, so that alternative was unreachable. Narrowed to `(?:\s|$)`. Checked for behavioural equivalence: the two regexes diverge only on colon forms, and in each of those the declared arm has already returned (or the token is not a registered key, giving `ToolError` either way).

Harusame64:
@codex review

New HEAD `63279c2`. Both round-9 findings are right, and rather than patch the two strings you named I have removed the reason this keeps recurring.

**You were right, measured.** Neutralising the code token in the flagged inputs shows whether any arm competes: `wrapped: KeyLockerDisabled: off via env` and the CDP-wrapped mouse messages both fall to `ToolError`, so nothing competed and relocating those arms would still have passed. Only the `focus_wait_timeout` case had a real competitor. That is the third round in a row a hand-written collision string turned out to pin nothing — the method was the problem, not the individual strings.

So the hand-written ordering pins are replaced by two invariants generated from the classify source:

1. **Literal self-routing** — every substring literal, wrapped so it reaches the cascade, must classify to its own arm. This is where the teeth are: every containment constraint (`keylockerspawnfailed` ⊃ `spawnfailed`, `browsersearchtimeout` ⊃ `timeout`, `terminal window not found` ⊃ `window not found`) is pinned for current *and future* arms with no hand-kept list.
2. **Pairwise earlier-arm-wins** — 4,245 synthesized two-token messages, proving the extracted model is the implementation, so a mis-parsed condition or missed literal surfaces as a failing pair.

Three things worth your attention:

- **Silent extraction misses are the obvious way this fails.** The arm count is reconciled against every `return { code:` in the region, the literal count against every `m.includes(`, any unmodeled predicate fails by name, and the single compound arm (`BrowserNotConnected`) is excluded by name so a second one forces a decision rather than vanishing. Please try to find a shape that slips past all four.
- **It does not claim full coverage, deliberately.** Positional intent between arms with no containment relation is not derivable, because "code token beats generic keyword" is *not* a global truth here — `TabDragBlocked`, `CrossWindowDragBlocked` and `ForegroundFlashRequiresTarget` sit AFTER the generic arms by design. Those anchors stay as hand pins that do carry a competitor. Please check that boundary is drawn where the cascade actually is.
- **Verified by mutation, not by reading.** Moving the `BrowserSearchTimeout` arm below the generic timeout arm fails three of the new assertions and names the shadowed pair. If a pin here does not bite under a reorder, it is a defect — please look for one.

Your second point is also fixed: the `mouse_drag` wording tripwire hard-copied its producer strings, so rewording `mouse.ts` would have left it pinning a string nothing emits. The messages are read out of the producer source now, and the read fails loudly if the producer stops building them in that shape.

Harusame64:
@codex review

New HEAD `c0949ac`. Your finding was correct, and I confirmed it by mutation before fixing it: adding `|| m.includes("browser offline")` to the `BrowserNotConnected` arm left all five assertions green — the return count reconciled, the literal count reconciled, the excluded-name list was unchanged, and the arm stayed invisible to both generation rules. Exactly the silent omission this file exists to prevent, in the file itself.

The blanket `continue` on any `&&` condition is gone. Conditions are now tokenized and parsed into disjunctive normal form, so an arm becomes a set of *firing token sets* (`a && (b || c)` → `{a,b}, {a,c}`). Compound arms take part in self-routing, pairwise ordering, and the well-posedness lemmas like every other arm; there is no exclusion list left.

One thing worth your judgement. Full DNF modelling alone does **not** fail the mutation — `browser offline` becomes its own firing set and self-routes correctly, so the model absorbs the change rather than flagging it. That is arguably right, but it means a compound arm could be reshaped without anyone looking. So the firing-set *shape* of compound arms is also pinned exactly, which is what fails the mutation (2 sets expected, 3 actual, with the diff printed). Please check whether that pin is drawn at the right level — too tight and it becomes churn, too loose and the blind spot returns.

Three more silent-omission holes surfaced while sweeping for others of the same shape, all closed:

1. An arm written `return { code: someConst }` escaped **both** the arm regex and the return-count reconciliation, since both required `code: "` — the counts agreed with each other while the arm was invisible. Now reconciled against every `return {` in the region.
2. An arm whose disjuncts are all wrapper-immune collapsed to zero firing sets and vanished from every rule. Now bails.
3. A wrapper-immune predicate joined by `&&` to a literal would have dropped its whole set silently. Only whole-disjunct immune predicates are accepted; anything else bails.

The tokenizer requires full consumption of the condition text, so an unknown predicate shape stops the run rather than being skipped — strictly stronger than the previous leftover check, which never looked inside `&&` conditions at all.

Detection power measured in both directions: the round-10 reorder mutation (`BrowserSearchTimeout` below the generic timeout arm) now fails 4 assertions where it failed 3, and reverting each mutation returns the file to green. `excluded` is gone with the mechanism that needed it, which also clears CodeQL #155 — no `_` prefix and no disable directive.

Unit project 296 files / 4763 passed; eslint 0 errors.

Harusame64:
@codex review

New HEAD `21bb0b4` (rounds 12 and 13). Your last finding was the sharpest of the series and I confirmed it by mutation before fixing it — it was the first one where **production behaviour actually regressed while the suite stayed green**:

```
baseline : "disabled" -> ElementDisabled       (arm at _errors.ts:809)
mutated  : "disabled" -> BrowserNotConnected   (arm at _errors.ts:806, grafted with `|| m === "disabled"`)
invariant: 6 passed
```

Four rounds in a row you found the same class — something dropped from the model without a pin — so rather than patch the fifth instance, the extraction now carries a stated rule: **every element of the condition syntax is either modelled, or dropped with an owner-and-shape pin, or bails loudly. Nothing leaves the model unpinned.** Three cases realise it: literals are modelled, compound shapes are modelled and shape-pinned, immune disjuncts are dropped with an ownership pin plus an end-to-end behaviour pin (`"disabled"` and `"guard failed for lens"` are fed bare — the wrapper cannot synthesise them — and must reach their owner arm with non-empty advice). A sweep found the immune `continue` was the last silent drop inside the region; outside the region remains a documented known limit.

Two things worth your check:

1. **Is the rule actually total?** It is only worth stating if it holds. Please look for a condition shape that is neither modelled, nor pinned, nor bails.
2. **Ownership pin tightness.** `GuardFailed` → `m.startsWith("guard failed")` and `ElementDisabled` → `m === "disabled"` are pinned exactly, so grafting either onto another arm fails. Too tight becomes churn when the classifier legitimately changes; too loose reopens this. Your read?

Regression suite, each mutation applied to `src/tools/_errors.ts`, run, then reverted (`git diff HEAD src/` empty at the end):

| mutation | result |
|---|---|
| A — `\|\| m.includes("browser offline")` on the compound arm | 1 named failure (firing-set shape pin) |
| B — `BrowserSearchTimeout` arm moved below the generic timeout arm | 4 named failures (shadow lemma, pair well-posedness, self-routing, pairwise) |
| C — tail arm returning via a helper instead of an object literal | 8 named failures, led by the completeness-count test |
| D — your finding, `\|\| m === "disabled"` grafted onto `BrowserNotConnected` | 1 named failure (immune ownership pin) |

Round 12 also closed C, which Opus found: a tail arm that returns neither an object literal nor matches `m.includes` slipped past all three completeness counts. Line-anchored `return` and `if (` counts now reconcile too, and extraction moved into a named test so a count break reports which count broke rather than surfacing as a collection error.

Unit project 296 files / 4765 passed; eslint 0 errors. `docs/operation-verification-matrix.md:291` is resynced to the current structure.

Harusame64:
@codex review

New HEAD `fa66913`. You named the fix and you were right that formatting was the wrong thing to count — I confirmed your mutation first: `} else if (m.endsWith("legacy condition")) return classify("disabled");` changed `"weird legacy condition"` from `ToolError` to `ElementDisabled` while all 8 tests passed.

Stepping back: this was the fifth escape you found in a row, and they were all the same shape — the extraction reads the source with regexes, so anything a regex cannot see stays outside the model. Round 13 stated the rule ("every element is modelled, or dropped with a pin, or bails") but a regex cannot enforce it, only approximate it over shapes it happens to match. So the extraction is now a **TypeScript AST walk** (`ts.createSourceFile`, `typescript` is already a devDependency).

The walk locates the single `classify` FunctionDeclaration and traverses every statement of its body in order. The frame is pinned positionally and structurally — `const declared` guard → `const m = message.toLowerCase()` → cascade → `const leadingCode` → guard → `return { code: "ToolError" }` → end, with a trailing statement bailing. Inside the cascade, anything that is not an `IfStatement` bails; `else if` chains are flattened (each arm body is proven to be a single return, so the chain is equivalent to sequential ifs); an arm body must be `return { code: "<string literal>" }` with plain property assignments only. Conditions are built from AST expression nodes rather than a string tokenizer, so `&&`/`||`/parens/`m.includes`/`m.startsWith`/`m === "…"` are recognised structurally and immune atoms are recorded in canonical spelling — the ownership pin is now formatting-independent. The Round 13 note about comments inside conditions is gone; comments are not AST nodes.

**The region limit is closed for `classify()` itself.** The walk covers the first statement through the last, and every statement is modelled, frame-pinned, or bails, so there is no invisible position left inside the function. What remains outside is genuinely outside — `failWith`/`errorFromMessage` routing and the contents of `SUGGESTS` — and the header says so as a relocated trust anchor rather than a silent gap.

Mutation results, each applied to `src/tools/_errors.ts`, run, then reverted (source verified byte-identical afterwards):

| mutation | result |
|---|---|
| A — `\|\| m.includes("browser offline")` on the compound arm | 1 named failure |
| B — `BrowserSearchTimeout` moved below the generic timeout arm | 4 named failures |
| C — tail arm returning via a helper | 8 named, led by `an arm must return an object literal, got CallExpression` |
| D — `\|\| m === "disabled"` grafted onto `BrowserNotConnected` | 1 named failure |
| E — your `else if` with a non-line-leading return | 8 named, same bail message |
| F — a `switch`-statement arm in the cascade (mine, to test the bail rather than trust it) | 8 named |

Worth your attention: whether a shape exists that is neither modelled, nor frame-pinned, nor bails. That is the whole claim now, and unlike the regex version it is the parser doing the enumerating rather than a pattern I wrote.

Unit project 296 files / 4765 passed; eslint 0 errors. `docs/operation-verification-matrix.md` §5.3 is resynced.

PR #552, opened 2026-07-26
docs(release): record that the release zip does not inherit `overrides`

Harusame64:
Found while confirming the postcss advisory during the 1.13.1 release. Docs only — no code changes.

## What the gap is

The "Install production dependencies" step in `.github/workflows/release.yml` builds the zip's `node_modules` from a **synthesised** `package.json` that carries only `name`, `version`, `type` and `dependencies`. The root **`overrides` block is dropped**, and the install runs with `--no-package-lock`.

So the dependency tree users download is neither the tree in our lockfile nor the tree `overrides` protects. A security fix applied through `overrides` covers local development and CI, and may or may not reach the zip depending on what the ranges resolve to on build day.

## Measured, not inferred

For `ip-address` (GHSA-v2v4-37r5-5v8g), which `overrides` pins to 10.1.1, read out of the actual v1.13.1 zip:

| | `express-rate-limit` | `ip-address` |
|---|---|---|
| lockfile (what we test) | 8.3.2 — pins `ip-address` to exactly 10.1.0, the vulnerable version | 10.1.1 via `overrides` |
| shipped zip | 8.6.0 — declares `^10.2.0` | 10.3.1 |

**No exposure today.** The shipped 10.3.1 is newer than the patched 10.1.1. But that is luck: the range moved forward past the advisory on its own, not because the override reached users. The same path would have silently shipped the vulnerable version if `express-rate-limit` had not moved.

Dev-only advisories — anything reached through `devDependencies`, such as `postcss` via `vitest` → `vite` — never enter the zip, so they are unaffected. That part of the 1.13.1 security work was correctly characterised.

## What this PR adds

- The gap, with the measurement above, under Operational Notes
- The check to run whenever a **runtime** advisory is fixed through `overrides`: read the package version back out of the built zip rather than trusting `npm ls`
- Removes a stale claim in the same section that this file is gitignored and local-only — it is tracked

## What it deliberately does not do

It does not change `release.yml`. Carrying `overrides` into the synthesised manifest, or installing from the lockfile, is a real fix but it changes what ships; that deserves its own PR with a smoke test behind it rather than riding along with a docs note.

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -651,12 +651,27 @@ is luck — the range resolved forward past the advisory on its own, not because
 the override reached the zip.
 
 **When you fix a runtime-dependency advisory with `overrides`, verify the fix in
-the built zip, not just in `npm ls`:**
+the built zip, not just in `npm ls`.** Check **every** installed copy, not the
+hoisted one: npm nests a second copy under a dependent whenever versions
+conflict, which is exactly the shape an incomplete override leaves behind. The
+v1.13.1 zip contains 114 nested `node_modules` entries (`body-parser/node_modules/content-type`,
+`execa/node_modules/semver`, …), so reading a single path can certify a release
+while a vulnerable nested copy ships beside it.
 
 ```bash
 curl -sL -o rel.zip "https://github.com/Harusame64/desktop-touch-mcp/releases/download/vX.Y.Z/desktop-touch-mcp-windows.zip"
-unzip -p rel.zip "node_modules/<pkg>/package.json" | grep '"version"'
+
+# Every copy of <pkg>, hoisted and nested. For a scoped package pass the full
+# name, e.g. @modelcontextprotocol/sdk.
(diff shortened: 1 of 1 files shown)


Harusame64:
@codex review

You were right, and this was the worst place to be wrong — the defect was in the security check the PR was adding, so it would have shipped a procedure that produces false all-clears.

`87871b2` fixes it. I confirmed the shape against the actual v1.13.1 zip rather than reasoning about it: the archive carries **114 nested `node_modules` entries**, so this is present in our own artifact, not hypothetical. `semver` demonstrates the exact miss —

```
node_modules/execa/node_modules/semver/package.json -> 5.7.2   <- invisible to the old command
node_modules/semver/package.json                    -> 7.8.5
```

The old single-path read would have reported only 7.8.5 and called it clean.

The step now enumerates every matching path and prints one line per installed copy, says to pass the full name for scoped packages, and states that empty output means the package is absent rather than that the check passed — since a silently empty result was a second way to read a pass out of nothing.

`ip-address`, the package this PR measures, has only the hoisted copy in the zip, so the measurement in the PR body still stands.

PR #557, opened 2026-07-27
ADR-031: native GDI virtual-screen capture — displayId/region works on non-primary and negative-origin monitors

Harusame64:
## Summary

`screenshot(displayId/region)` could not capture any non-primary monitor (and any negative-origin coordinate) because the absolute-coordinate pixel path delegated to libnut, whose coordinate validation is primary-monitor-only (`x >= 0` against `SM_CXSCREEN`). This PR replaces that path with an in-house Rust GDI capture that treats the virtual screen (negative origins included) as first-class, per ADR-031.

Plan: `desktop-touch-mcp-internal@895d698:docs/adr-031-plan.md` ([permalink](https://github.com/Harusame64/desktop-touch-mcp-internal/blob/895d698a44c9b32cdb7d9d08feebb36ee8531443/docs/adr-031-plan.md))

## Commits

| SHA | Scope |
|---|---|
| `259bccf` | Phase R — Rust `win32CaptureScreenRegion` (signed coords, SRCCOPY only), native 3-point set, `hasNativeCaptureRegion()` probe, cargo units |
| `c86e39a` | Phase T — `selectCaptureBackend()` / `resolveCaptureRegion()` in `reachable-bounds.ts` (static per-process selection, `DESKTOP_TOUCH_CAPTURE_BACKEND=nutjs` override, null-preserving bounds), choke-point helper in `image.ts` (validation → dispatch → whole-pipeline typed wrap → diagnostic log, `kind:"capture"` event), `RegionOutsideCapturableBounds` / `CaptureBackendFailed` + SUGGESTS, containment-conditional ui-elements clamp, 54 new unit tests + invariant test + mutation-verified guards |
| `d13adcd` | Phase W — server instructions, README (en/ja) env table, CHANGELOG |
| `eb7637e` | Test-contract update: per-window fallback assertions retargeted from the nut-js mock to the choke-point backend (same rectangles asserted; hermeticity fixes pin the mock layer against the developer's display layout) |

## Verification

- Guard checklist all green: cargo check/test (185/185), `check:native-types`, `build`, `lint`, `check:stub-catalog`, `check:failwith-fixtures`, `check:napi-safe`.
- Full suite: 3 failed / 5297 passed — the 3 are pre-existing/environmental (2× `desktop-executor-reachable-bounds` fail only while a second monitor is attached — confirmed failing on baseline with the VDD attached, green on single-monitor; 1× `browser-cdp-verification` known failure, recorded in remaining-work).
- **Multi-monitor E2E (`tests/e2e/adr-031-multimonitor-capture.test.ts`): 7/7 PASS on a live negative-origin layout** (VDD virtual display at (-1920,0) 1920x1080): displayId 0/1 exact-size, negative-origin region, seam-spanning region, dotByDot negative origin reporting, fullscreen still primary-sized with no origin line, repeat-capture dimension parity, off-desktop region → typed error through engine and tool. Skips on single-monitor CI.
- Mutation checks: removing the bounds guard fails 9 tests; removing the clamp containment condition fails 3 (including the "element must not be pulled onto the primary monitor" pin).

## Plan deviations (recorded, director-approved)

1. **The plan's "2-site allowlist" was a 3-site reality**: `scroll-capture.ts:77` held an unenumerated direct `screen.grabRegion`. It now routes through the choke-point helper (exported `@internal` — the one property changed from the plan's "non-export"), and the invariant test is stronger than specified: zero grab sites outside `image.ts`. Will be recorded in the internal plan doc in the post-merge sync.
2. **No v2-instructions line**: `desktop_act` never reaches `captureScreen` (importers verified), and naming a code it cannot produce would violate the producer-pin rule. Server instructions carry the codes.

## Dogfood checklist (does not block merge; release-gated per plan §5 R3 — environment will be the VDD virtual display, recorded as such)

- [x] ① Negative-X full capture (measured during the spike + E2E 7/7; re-confirm on the release build)
- [ ] ② Positive-X secondary (rearrange virtual monitor to the right)
- [ ] ③ Negative-Y (virtual monitor above)
- [ ] ④ Uniform 150 % DPI + same-region native/nutjs dimension comparison (nutjs via `DESKTOP_TOUCH_CAPTURE_BACKEND=nutjs` in a separate process)
- [ ] ⑤ Mixed DPI (one side 125 %)

Decision rule: native size ≠ requested w×h → release blocks until root-caused; native ≠ nutjs only → no block, native is the contract.


Harusame64:
@codex review

Harusame64:
**Review loop status — paused mid-Round 2** (local working tree holds uncommitted Round 2 fixes; do not re-review until the Round 2 commit is pushed)

Round 1 (resolved by `d04881b`): Codex P2 (gap-of-staggered-layout regions pass the bounding-rect check) + Opus P1 (containment also rejected machine-derived window rects — a maximised window's rect overshoots its monitor by the invisible resize border) → two validation modes: `contain` (caller regions, now also requires intersecting a real monitor) / `overlap` (window rects, unclamped). Plus 4 further Opus P2s and 2 P3s (see the commit message).

Round 2 findings, all confirmed against sources, fixes in progress:
- **Opus P1**: the DPI note added in Round 1 states as measured fact what has not been measured — the forked nut-js has no `getScaledRect` (grep of node_modules: zero hits) and the process is per-monitor-DPI-aware, so the "used to return 2560x1440" claim is likely false. Decision: drop the note; rewrite from dogfood measurements later, as the implementation plan already requires.
- **Opus P2**: `overlap` mode also relaxed the boundary on the nut.js backend, where libnut cannot clip and throws — swapping an accurate `RegionOutsideCapturableBounds` for a misleading `CaptureBackendFailed`. Decision: the overlap relaxation applies to the GDI backend only; the nut.js variant keeps containment.
- **Opus P3 x2**: the choke-point scan misses direct `win32CaptureScreenRegion` binding calls; a `padCaptureRegion` doc line still implies gaps are capturable.
- **Codex P2**: `GetDIBits` is called while the bitmap is still selected into the memory DC, which Win32 forbids. Confirmed, and `win32_print_window_to_buffer` in the same file has the identical pre-existing pattern — both will be fixed in the Round 2 commit.

Harusame64:
**Round 2 pushed** (657138b) — the pause noted above is lifted; re-review welcome.

All five Round 2 findings are in: GetDIBits is now called only after the SelectObject guard is dropped (both capture functions — the pre-existing `win32_print_window_to_buffer` pattern included); the `overlap` relaxation now keys on the resolution kind, so the nut.js path keeps containment in both modes and refuses with `RegionOutsideCapturableBounds` instead of surfacing libnut's throw as `CaptureBackendFailed`; the unmeasured DPI note is dropped from CHANGELOG and both READMEs; the choke-point scan also matches the raw `win32CaptureScreenRegion` export; and `padCaptureRegion`'s doc names the bounding-rect test instead of implying gaps are capturable.

Verification: cargo test green, unit suite green, and the VDD multi-monitor E2E re-run 7/7 against the rebuilt native binary (hash-verified so the green is against the Round 2 code, not a stale .node).

Harusame64:
@codex review

Harusame64:
**Round 3 pushed** (ed5bbcf) — zero P1, four P2 and four P3, all applied.

- `win32_capture_screen_region` now treats a NULL `SelectObject` as an error. Previously it carried on, so BitBlt drew into the memory DC's default 1x1 monochrome bitmap while GetDIBits read our never-drawn-into bitmap, and an all-zero buffer left the function as a *successful* black frame — this path has no blank-capture detection to catch that. `win32_print_window_to_buffer` keeps its NULL tolerance, because every caller sits inside the window-capture ladder whose lower rungs catch both the error and the all-black frame; its comment now states that contrast explicitly instead of leaving the asymmetry unexplained.
- The dimension-parity TSDoc no longer states the native backend's device-pixel return as measured fact. It now reads as the design expectation of a screen-DC BitBlt, with measurement deferred to the dogfood passes the implementation plan already schedules. The `sizeChanged` guidance to callers is unchanged. This is the same proposition dropped from CHANGELOG and both READMEs in Round 2, so no assertive form of it survives anywhere.
- `docs/system-overview.md` now attributes screen capture correctly: `nutjs.js` is listed as the primary-monitor-only fallback rather than the capture route, `image.ts` is marked as the screen-capture choke point, and Layer 0 gained a `gdi.rs` entry for the BitBlt virtual-screen capture.
- P3s: the `CaptureBackendFailed` recovery text names the diagnostic log with a Windows path (`%USERPROFILE%\.desktop-touch-mcp\logs\diagnostic.log`) instead of a tilde, matching the CHANGELOG; a step-numbering comment now matches where the `try` actually begins; and a wrong order-of-magnitude constant in two comments is corrected to `~1.8 x 10^19`.

Verification: `cargo check --workspace --locked` and `cargo test -p desktop-touch-engine --locked` green, `check:napi-safe` green, `tsc` clean.

Full-suite run under the still-enabled VDD multi-monitor environment shows two additional env-dependent failures in terminal.test.ts (conhost background input); the same suite was green on those tests before VDD was enabled, and no terminal-path code changed in this round.

Harusame64:
**Round 4 pushed** (82ee8c6) — the three Codex P2s from the last two review passes are closed, plus the final open plan item.

**Deterministic black fill.** Both capture functions now `PatBlt(..., BLACKNESS)` the bitmap before drawing into it. `CreateCompatibleBitmap` does not initialise its contents, and neither draw covers the whole destination: `BitBlt` writes only the area the source overlaps — an `"overlap"` capture of a window overhanging its monitor leaves a strip untouched — and `PrintWindow` is explicitly allowed to fail or render partially. That uncovered area was undefined memory rather than the black margin every caller assumes, and on the window ladder the noise could also slip past the all-black + zero-variance blank check, which exists to catch precisely that frame. Two captures of an unchanged rectangle are now byte-identical. The `reachable-bounds` comment that asserted BitBlt returns the off-screen part black now explains why that holds.

**Overhang-aware refusal.** `RegionOutsideCapturableBounds` no longer blames stale coordinates for a region that overlaps a real monitor but extends past the capturable area. Re-discovering hands back the same rectangle, so the old advice sent the caller round a loop that could not terminate. The refusal now names the monitors the region does overlap and advises shrinking it or capturing the window directly. Both wordings are pinned by unit tests, arranged so that deleting the branch fails one and making it unconditional fails the other.

**Layout-aware seam E2E.** The seam test derived its boundary from "the left monitor's right edge", which on a vertically stacked or gapped layout points at no monitor — the region then falls outside the union and is correctly refused, failing the test for a layout the implementation actually supports. It now searches every monitor pair for a genuinely shared edge whose overlap is long enough to hold the region, centres the region on that overlap, and skips with the current layout printed when no such edge exists. On the test layout here it executes rather than skipping (confirmed in the JSON report, not just the pass count).

**Plan Phase V.** The `(full screen, no origin offset)` disclosure is now pinned by a unit test that invokes the real screenshot handler, closing the last item that was tracked as unevidenced.

Verification: `cargo check --workspace --locked` and `cargo test -p desktop-touch-engine --locked` green, `check:napi-safe` green, `tsc` clean, the native module rebuilt and SHA-256-verified against the freshly built artifact so the E2E green is against Round 4 code, the unit project green, and the ADR-031 multi-monitor E2E 7/7.

Two deliberate exclusions from this run. The unit project still shows the two `desktop-executor-reachable-bounds` failures that appear only while the virtual second monitor is attached, and the remaining e2e suites were not run at all: their conhost and blank-window helpers surface windows on the primary display, and that test-infrastructure issue is being fixed separately. Neither set touches the capture path changed here.

Harusame64:
**Round 5 pushed** (60ce667) — one P1, two P2 and three P3, all applied.

**P1 — the recovery steps had not caught up with the message.** Round 4 split the refusal into three cases, but `RegionOutsideCapturableBounds`'s recovery steps still described two. An overhanging region — one that overlaps a monitor but runs past the capturable area — therefore fell through to the "off every monitor" advice and was told to re-run `desktop_discover`, which hands back the very same rectangle. That is a loop with no exit, and it is exactly the failure Round 4's message change was meant to end; the suggests were quietly undoing it. The third case is now stated explicitly and says the coordinates are current and the region is what has to change.

**P2 — four other surfaces still said "two cases".** The server instructions, both READMEs and the typed-error TSDoc all described the old split. Each now reads as three: on no monitor (stale coordinates), overlapping but overhanging (coordinates current — shrink the region or capture the window), or a server limited to the primary monitor.

**P2 — two comments still credited BitBlt with the black margin.** BitBlt writes nothing outside the source it copies; Round 4's explicit clear is what makes that area black. Both comments now say so. The `overlap` relaxation note is the load-bearing one, so it now states plainly that its conclusion is unchanged — what earns the relaxation is that BitBlt *clips* instead of failing, which was true before and after the clear.

**P3s.** The choke-point fail-open comment now records that black is guaranteed on the native backend only, since the nut.js backend throws rather than returning a black frame (assertions untouched). CHANGELOG gains the deterministic-black behaviour described in user terms, and its account of the refusal is corrected to three cases.

Verification: `tsc` clean; `check:stub-catalog` exit 0 — worth noting the suggests text does not reach the generated catalogue, so the usual catalog-drift regeneration was not needed here; `check:failwith-fixtures` exit 0; unit project `2 failed | 4945 passed | 6 skipped`, the two failures being the known `desktop-executor-reachable-bounds` pair that fails only while the virtual second monitor is attached. No existing test that pins this error's wording broke.

This round is strings, comments and docs only — no Rust and no behaviour change — so the native binary and the multi-monitor E2E result from Round 4 still stand. The remaining e2e suites were again left unrun while their window-spawning helpers are being fixed separately.

Harusame64:
**Round 6 pushed** (688f43c) — Round 5 came back with zero P1 and zero P2; this round clears the remaining three P3s.

**The nut.js refusal no longer tells a window to move where it already is.** `GetWindowRect` reports a maximised window a few pixels outside its own monitor, so its rect overlaps the primary monitor while overhanging it. On the nut.js path that combination was refused with "move the target window onto the primary monitor" — an instruction to do what has already been done. The message now splits the same way the GDI branch has since Round 4: a region that overlaps the primary is told the window is already there and to shrink the region or capture the window itself, and only a region that misses the primary entirely keeps the move-the-window advice. The new test asserts both directions in a single case, so deleting the branch fails one half and making it unconditional fails the other.

**The deterministic-black CHANGELOG note is now scoped.** It said uncovered areas come back black "every time"; it now records that this describes the normal capture path and does not apply under `DESKTOP_TOUCH_CAPTURE_BACKEND=nutjs` — the older path, which the very same entry introduces a few lines down.

**A comment left ragged by the Round 5 edit is reflowed.**

Verification: `tsc` clean, and the two ADR-031 unit files 71/71 including the new case.

On the unit suite: besides the two `desktop-executor-reachable-bounds` failures that occur only while the virtual second monitor is attached, the full unit run was additionally impacted by a cursor hold from the user's own game-input layer (`cursor_placement_blocked`), diagnosed and excluded as a code cause.

Harusame64:
**Round 7 pushed** (d6a3ccf) — both Codex P2s from the Round 5 pass are closed. One of them turned out to be considerably larger than reported.

**The missing loader export was one of eleven.** Codex flagged that `win32CaptureScreenRegion` is declared in `index.d.ts` but never re-exported from the hand-maintained `index.js` wrapper, so a consumer writing `import { win32CaptureScreenRegion } from "./index.js"` type-checks and then fails at module link time. Rather than fix the one name, we swept the two files against each other — and found the identical gap in ten more bindings:

```
win32MoveCursorAbsolute        win32GetImeOpenStatus
win32MoveCursorPath            win32SetImeOpenStatus
win32GetCursorPos              win32GetProcessSessionId
win32ForegroundFlashInject     win32GetActiveConsoleSessionId
win32ConsolePasteNoFocus       wtsEnumerateSessions
```

Nothing inside `src/` had noticed any of them, because `native-engine.ts` consumes the whole binding object rather than named imports — the breakage was reserved for anyone trusting the published `.d.ts`. All eleven are now exported.

**That class of drift is now a build failure.** `scripts/check-native-types.mjs` gained a parity check between `index.d.ts` declarations and `index.js` exports, in **both** directions — an undeclared export leaves TS consumers untyped, so it fails too. Two notes on how it was built: it was written *before* the fix and confirmed to fail on all eleven names, so it is known to have detection power rather than merely being green; and it matches declared classes as well as functions, otherwise `DirtyRectSubscription` reports as a phantom `index.js`-only export.

**The addon-less build could not enforce its own limitation.** When no native addon is present at all, the process selects nut.js — but every monitor lookup (`enumMonitors`, `getVirtualScreen`, `getPrimaryMonitorBounds`) routes through the addon and throws, so the bounds resolver saw "unknown layout" and failed open. That happened on precisely the build whose capture *is* limited to the primary monitor: an off-primary region sailed through the check, reached libnut, and came back as `CaptureBackendFailed` — advice about locked screens and UAC prompts for a process whose real problem is that it has no multi-monitor capture.

The fix deliberately leaves `resolveCaptureRegion()` synchronous and cache-free, exactly as specified, along with all of its existing callers and tests. A new async wrapper sits on top: it calls the sync core first, and only when that returns null *and* the backend is nut.js does it ask the nut.js backend for its own primary screen size and synthesise the primary-rect bounds from it. If nut.js cannot report a size either, nothing is known and the long-standing fail-open stance is unchanged. The GDI path is not touched. `nutjs.js` is imported dynamically so the sync path never drags nut.js's native backend into modules that only need the resolver.

Verification: `tsc` clean, `check-native-types` green with the new parity check, `check:napi-safe` green, and the two ADR-031 unit files 73/73 — including a test that the refusal happens *before* libnut is called (the whole point of the fix) and its contrast that an unreadable screen size still fails open. The capture choke-point scan is unaffected: reading a screen *size* is not a grab, and the new wrapper never names the capture binding.

Harusame64:
**Round 8 pushed** (f048d6d) — one P2 and three P3 from the Round 6+7 pass; every earlier finding stays closed.

**`bounds_unknown` was being recorded before failing open was actually decided.** The synchronous resolver wrote it the moment it returned null. On the nut.js path that is premature: the async wrapper added in Round 7 can still recover bounds from the nut.js backend and go on to *refuse* the region — so the diagnostic log asserted that a capture had passed through unchecked when in fact it had been checked and rejected. The second-order effect was worse than the wrong line itself. The record is warn-once, so that false entry burnt the latch, and the first genuine unknown — the case the record exists for — would never be written down at all.

The record and its latch now live in the wrapper, at the point where failing open becomes final: immediately for the GDI backend, which has no fallback to wait for, and only after the nut.js fallback has been attempted and failed otherwise. A successful fallback logs `bounds_from_nutjs` alone and leaves the latch unburnt. Nothing else the resolver records was moved — refusal logging stays exactly where it was.

Both Round 7 tests gained the corresponding pins: fallback succeeds ⇒ `bounds_from_nutjs` 1 / `bounds_unknown` 0; fallback fails ⇒ `bounds_unknown` 1 / `bounds_from_nutjs` 0. The pre-existing "reports unknown, once" test keeps its `resolveCaptureRegion()` null assertions against the sync core and now additionally pins that the core stays **silent**, with the logging assertion moved onto the wrapper. The `diagnostic-log.ts` definition comment was rewritten to match the new placement rather than left describing the old one.

**Two guard fixes.** The parity check no longer hard-fails on a name the Rust drift check only soft-warns about — `stale` declarations may be feature-gated, so demanding an `index.js` re-export for them put the two rules in direct contradiction. They are excluded in that direction, still checked in the other, and the relationship is now written down where both rules are visible.

The guard also fails loudly on an `export` line whose shape it cannot parse, rather than skipping it. This one matters more than it looks: an unparsed export drops out of *both* directions of the comparison, so it reads as parity while proving nothing — the same vacuous-guard failure mode the choke-point scan was hardened against earlier in this PR. Confirmed by feeding it an `export { x as y };` form, which fails with "unrecognized export form — extend the parser", and the tree is green again once removed.

**One preventive test mock.** `adr-031-ui-elements-scope-clamp` now stubs `nutjs.js`. That path is dormant today because the win32 mock still returns bounds, but if it ever stops doing so the real nut.js native backend would load inside a unit test.

Verification: `tsc` clean, `check-native-types` green with both guard changes, and 193 unit tests green across the three ADR-031 files plus the `diagnostic-log`, `reachable-bounds`, `cursor`, `classify-cascade-order-invariant`, `screenshot-printwindow-default` and `scroll-capture-ref` suites.

Harusame64:
**Round 9 pushed** (bfb0e30) — this closes the Codex P2 raised against the Round 6 commit. The other Codex P2, raised against Round 7, was the same defect Round 8 had already fixed, so no separate change was needed for it: Round 8 moved the `bounds_unknown` record into the async wrapper and emits it only after the nut.js fallback has been tried and failed, which is exactly the second remedy that review suggested.

**The capture buffer is now allocated fallibly.** Both capture functions used `vec![0u8; n]`, which has no failure path — on OOM Rust aborts the process instead of returning the `CaptureBackendFailed` the caller is promised, so a recoverable capture failure took the whole server down with it.

The size ceiling added earlier does not prevent this, and it is worth being precise about why. It bounds how large a request may be, but a bound is not a guarantee the allocation succeeds, and the gap is reachable rather than theoretical: when monitor enumeration fails, the bounds check deliberately fails open, so a region just under the ceiling can reach the allocator on a machine that cannot satisfy it. By that point the GDI bitmap of the same dimensions is already resident, so the real peak requirement is roughly double the size that was checked.

Both sites now reserve with `try_reserve_exact` and return an error on failure. The `resize` that follows cannot reallocate against the already-reserved capacity, so it cannot abort either.

**Allocation audit of the file**, since a fix on two sites is only as good as the claim that there are two: these are the only heap allocations that scale with the requested rectangle. Everything else is fixed-size and small — `BITMAPINFO` / `BITMAPINFOHEADER` are stack values — and `Buffer::from` hands the already-allocated vector to napi rather than allocating a second time.

**Two comments were repaired rather than left behind.** Both justified the size ceiling with "`vec!` aborts", which is no longer why it exists. They now state what it still does: refuse a nonsensical rectangle *by name*, before a bitmap of those dimensions is ever asked of GDI, instead of letting the attempt surface later as an opaque out-of-memory.

Verification: `cargo check --workspace --locked` and `cargo test -p desktop-touch-engine --locked` green on the first run (187 passed, 0 failed), the native module rebuilt and SHA-256-verified against the freshly built artifact so the binary matches this source, and the two ADR-031 unit files 73/73 as a smoke check. No TypeScript changed in this round.

Harusame64:
**Round 10 pushed** (6c70c60) — this closes the Codex P2 raised against Round 8.

**The recovery advice was correct for one reason a process can be primary-limited, and a dead end for the other.** `RegionOutsideCapturableBounds` recommended `screenshot(windowTitle=…)` unconditionally. Under the `DESKTOP_TOUCH_CAPTURE_BACKEND` override that is sound: the native module is installed and only the region path is pinned to nut.js, so per-window capture still reads through PrintWindow on any monitor. Without the module it is not, and the chain is worth spelling out because every link is in this PR's own code:

1. `printWindowToBuffer` calls `requireNativeWin32()`, which throws when the addon is absent.
2. `captureWindowRawPrintWindow` catches that and returns `null`.
3. The ladder falls through to a BitBlt of the window rect, routed through the capture choke point.
4. That lands on this same primary-monitor guard, which refuses an off-primary window rect.

So the one case that most needed a way out was being sent into a second guaranteed failure. It now gets the two routes that actually work — move the window onto the primary monitor, or reinstall to restore multi-monitor capture — and the message says the per-window route needs the same missing module rather than leaving the caller to discover that by trying it.

**Synchronised across every surface carrying the claim**, so the fix does not leave behind the drift it exists to remove. A grep turned up five: `describeUncapturable` (both nut.js variants), the `SUGGESTS` list, the server instructions, both READMEs, the CHANGELOG entry, and the typed-error TSDoc. The `SUGGESTS` list is a static array and cannot branch on the determinant, so its single line became two, following the "read the message first" idiom already used there — which now also states that the message names *why*, since that is what selects the recovery.

Two details worth flagging. The overhang variant keeps "shrink the region" in **both** determinants, because that advice is correct regardless, and it keeps saying the window is already on the primary monitor so moving it will not help — writing the module-less advice naively would have produced a message that says moving will not help and then advises moving.

And `CaptureBackendFailed`'s suggests are deliberately untouched. Reaching *that* error on a module-less build means the read was on the primary monitor, where per-window capture still succeeds through the nut.js fallback, so its window-capture advice remains true.

Verification: `tsc` clean; `check:failwith-fixtures` and `check:stub-catalog` both exit 0 with no regeneration needed despite the suggests list changing length; and the two ADR-031 unit files 75/75. The two new assertions cover both determinants and are arranged so that deleting the branch fails one and making it unconditional fails the other — and the Round 6 assertions on this same message all had to keep passing, which constrained the wording rather than the other way round.

Harusame64:
**Round 11 pushed** (c654b33) — two small closures on the Round 10 change.

**The overhang branch was unpinned.** It carries its own copy of the determinant split added in Round 10, and none of the existing assertions could reach it: they all judge a region that misses the primary monitor entirely, which returns from the other branch. So the ternary could have been deleted in either direction without a test noticing. There is now a matching pair — an overhang under the env override must still offer per-window capture, and the same overhang on a module-less build must not — and each direction of the mutation fails exactly one of them.

**The overhang recovery step no longer names `screenshot(windowTitle=…)`.** An overhang can occur on either backend, and whether that route works depends on the determinant, which the two lines below it already own; naming it unconditionally there reintroduced the same wrong advice Round 10 removed. Shrinking the region is the one answer that holds in every case, so it now stands alone.

Verification: `tsc` clean, `check:failwith-fixtures` exit 0 with no regeneration despite the suggests text changing, and the two ADR-031 unit files 77/77.

Harusame64:
**Round 12 pushed** (d520288) — the CI build failure and the Codex P2 on Round 11.

**CI failure: an eslint error, not a type error.** `no-useless-assignment` on the async bounds wrapper added in Round 7 — `size` was initialised to `null` and then overwritten on both paths through the `try`/`catch`, so the initialiser was dead. It is gone, relying on definite-assignment analysis, which is the idiom this repo already documents for `fallbackReason` in `engine/image.ts`.

The more useful outcome is the process fix: every round so far ran `tsc` but never `npm run lint`, which is exactly why this reached CI instead of being caught locally. **`npm run lint` is part of the standing guard list from this round onward.**

**Codex P2: the advice was inferred from configuration when it is a question about capability.** Whether to recommend `screenshot(windowTitle=…)` was decided from the determinant, but per-window capture and region capture are separate bindings, and the two come apart in *both* directions:

- an addon predating ADR-031 exports `win32PrintWindowToBuffer` but not `win32CaptureScreenRegion`. The probe only tests the latter, so the determinant reads `no-native-module` — while per-window capture in fact works on every monitor, and we were withholding it.
- an addon missing entirely while `DESKTOP_TOUCH_CAPTURE_BACKEND` happens to be set reports `env-override`. PrintWindow is gone, per-window capture fails too — and we were recommending it.

Both refusal variants now ask a new `hasNativePerWindowCapture()` probe. It keys on PrintWindow alone, deliberately: WGC is a rescue rung layered above it rather than an independent route, and the ladder's last rung — BitBlt of the window rect — goes through the capture choke point and is refused for an off-primary window on a build without region capture. So PrintWindow is the one binding that makes per-window capture work *everywhere*, which is the claim the advice makes.

`why` still explains what made the process primary-limited, because that genuinely is a determinant question. Only the recommendation moved to capability.

**The recovery steps are relabelled to match.** They previously said "BY THE ENV OVERRIDE" / "BY A MISSING MODULE" — a distinction the static list cannot actually evaluate, and which we now know is the wrong axis anyway. They key on whether the message *offers* per-window capture, and the lead-in says so.

**Tests take the two probes independently.** Both configurations Codex named are pinned, and the pre-existing pairs are renamed to capability terms and made capability-consistent — a module-less build now mocks both bindings absent rather than only the region one, which is what a real module-less build looks like.

Verification: lint clean, `tsc` clean, `check:failwith-fixtures` / `check:stub-catalog` / `check:native-types` all exit 0, and 199 unit tests green across the three ADR-031 files plus the six suites that touch these modules.

Harusame64:
**Round 13 pushed** (8cddbf1) — a wording follow-up to Round 12, three lines.

Round 12 moved the code's advice from the determinant to a direct capability probe, because per-window capture and region capture are separate bindings and a build can have one without the other. Three surfaces still explained the split in determinant terms: the server instructions and both READMEs. That framing is true of the builds people actually have, and misleading on a partial one.

Each keeps its existing explanation as the typical case and gains one clause — the two capture abilities are separate parts of the module, so read what the message says rather than inferring it from the cause. The user-facing wording stays in plain terms ("whole-screen capture and single-window capture are separate parts of that module"), with no internal vocabulary.

Verification: lint clean, `tsc` clean.

PR #558, opened 2026-07-28
test(e2e): place spawned test windows on a non-primary monitor when one exists

Harusame64:
## Why

The E2E suite drives real windows on a real desktop. On a multi-monitor machine every window it spawns — the blank click target and the PowerShell consoles — landed on the **primary** monitor, i.e. right on top of the screen the developer is working on. Windows pop over the editor, the foreground focus is yanked away mid-test, and the user's own clicks and keystrokes race with the suite's synthetic input (which also makes focus-sensitive tests flaky).

This PR moves those windows onto a non-primary monitor when the machine has one. Test infrastructure only — no assertions and no product code were touched.

## What changed

New helper `tests/e2e/helpers/e2e-screen.ts`:

- `pickE2eScreen(size?)` — returns the first non-primary monitor's placement origin (its work area, offset by a 120px margin and clamped so a window of `size` still fits) plus that monitor's full bounds. Returns `null` when there is only one monitor, when monitor enumeration is unavailable (non-Windows / headless), or when no monitor reports itself primary.
- `moveWindowToE2eScreen(hwnd, size)` — best-effort relocation of an already-spawned window, preserving its size.

Call sites:

- `blank-window.ts` — the WinForms form is created at the picked origin instead of the hardcoded `120,120`.
- `powershell-launcher.ts` — Windows offers no launch-time placement flag for `conhost` / `wt`, so the tagged window is moved immediately after `findByTag`. Size is read from the enumerated window region and preserved; the move goes through `setWindowBounds` (`SWP_NOZORDER`), so Z-order, TopMost state and foreground focus are unchanged.

Coordinates stay in virtual-screen space and are deliberately **not** clamped to non-negative values — a monitor to the left of the primary has a negative origin, and that is exactly the layout worth exercising.

**Single-monitor machines and CI behave exactly as before**: `pickE2eScreen()` returns `null` and each call site keeps its previous placement.

## Verification

Run on a two-monitor layout (primary `0,0 1920x1080`; secondary `-1920,0 1920x1080`):

- `tests/e2e/mouse-verify-delivery.test.ts` — 6/6 passed.
- `tests/e2e/terminal.test.ts` — 31 passed, 1 skipped. Also re-run with the helper changes reverted: identical result, so this PR neither fixes nor regresses that file.
- Placement evidence from a throwaway probe: `pickE2eScreen({width:480,height:360})` → `origin {x:-1800,y:120}`; blank window rect `{-1800,120,480,360}`; conhost window rect `{-1800,120,993,399}` (size preserved). Both windows landed on the secondary monitor, off the primary screen.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -382,7 +382,8 @@ export declare function win32ShowWindow(hwnd: bigint, nCmdShow: number): boolean
 export declare function win32SetForegroundWindow(hwnd: bigint): boolean
 export declare function win32SetWindowTopmost(hwnd: bigint): boolean
 export declare function win32ClearWindowTopmost(hwnd: bigint): boolean
-export declare function win32SetWindowBounds(hwnd: bigint, x: number, y: number, cx: number, cy: number): boolean
+/** `noActivate` (optional): adds SWP_NOACTIVATE so the move never brings the window forward. Omitted/false = SWP_NOZORDER only (historical behaviour). */
+export declare function win32SetWindowBounds(hwnd: bigint, x: number, y: number, cx: number, cy: number, noActivate?: boolean): boolean
 export declare function win32ForceSetForegroundWindow(hwnd: bigint): NativeForceFocusResult
 export declare function win32ForegroundFlashInject(targetHwnd: bigint, targetPid: number, text: string, options: NativeForegroundFlashOptions): NativeForegroundFlashResult
 export declare function win32GetFocusedChildHwnd(targetHwnd: bigint): bigint | null
--- a/src/engine/native-engine.ts
+++ b/src/engine/native-engine.ts
@@ -166,7 +166,8 @@ export interface NativeWin32 {
   win32SetForegroundWindow?(hwnd: bigint): boolean;
   win32SetWindowTopmost?(hwnd: bigint): boolean;
   win32ClearWindowTopmost?(hwnd: bigint): boolean;
-  win32SetWindowBounds?(hwnd: bigint, x: number, y: number, cx: number, cy: number): boolean;
+  /** `noActivate` is optional; omitting it keeps the pre-existing flag set (SWP_NOZORDER only). */
+  win32SetWindowBounds?(hwnd: bigint, x: number, y: number, cx: number, cy: number, noActivate?: boolean): boolean;
   win32ForceSetForegroundWindow?(hwnd: bigint): NativeForceFocusResult;
   win32GetFocusedChildHwnd?(targetHwnd: bigint): bigint | null;
   win32BuildProcessParentMap?(): NativeProcessParentEntry[];
--- a/src/engine/win32.ts
+++ b/src/engine/win32.ts
@@ -534,16 +534,22 @@ export function findAncestorWindow(startPid: number): {
  * Move and resize a window in a single SetWindowPos call, without changing Z-order.
  * x/y/width/height are in virtual screen coordinates (Per-Monitor DPI aware).
  * Returns true on success.
+ *
+ * `noActivate` is optional and defaults to the historical behaviour (flags =
+ * SWP_NOZORDER only), so existing callers are unaffected. Pass `true` to add
+ * SWP_NOACTIVATE when the window must be repositioned WITHOUT being brought to
+ * the foreground — SetWindowPos may otherwise activate the window it moves.
  */
 export function setWindowBounds(
   hwnd: unknown,
   x: number,
   y: number,
   width: number,
-  height: number
+  height: number,
+  noActivate?: boolean
 ): boolean {
   if (typeof hwnd !== "bigint") return false;
-  return requireNativeWin32().win32SetWindowBounds!(hwnd, x, y, width, height);
+  return requireNativeWin32().win32SetWindowBounds!(hwnd, x, y, width, height, noActivate);
 }
 
 /**
--- a/src/win32/window_op.rs
+++ b/src/win32/window_op.rs
@@ -95,6 +95,13 @@ pub fn win32_clear_window_topmost(hwnd: BigInt) -> napi::Result<bool> {
 /// the caller says so — which is wrong for any caller that repositions a window
 /// it does not want to bring forward (e.g. relocating a test window off the
 /// user's screen must not steal the focus that user is typing into).
+///
+/// Measured on Windows 11 (PR #558): moving a background window with
+/// `SWP_NOZORDER` alone did NOT change the foreground, with or without this
+/// flag — activation appears tied to the Z-order move that `SWP_NOZORDER`
+/// suppresses. So this is a contract guarantee, not a fix for an observed
+/// focus steal; do not weaken it on the strength of that measurement, which
+/// covers one OS build.
 #[napi]
 pub fn win32_set_window_bounds(
     hwnd: BigInt,
--- a/src/win32/window_op.rs
+++ b/src/win32/window_op.rs
@@ -13,7 +13,7 @@ use napi_derive::napi;
 use windows::Win32::Foundation::HWND;
 use windows::Win32::UI::WindowsAndMessaging::{
     SetForegroundWindow, SetWindowPos, ShowWindow, SET_WINDOW_POS_FLAGS, SHOW_WINDOW_CMD,
-    SWP_NOMOVE, SWP_NOSIZE, SWP_NOZORDER,
+    SWP_NOACTIVATE, SWP_NOMOVE, SWP_NOSIZE, SWP_NOZORDER,
 };
 
 use super::safety::napi_safe_call;
@@ -85,6 +85,16 @@ pub fn win32_clear_window_topmost(hwnd: BigInt) -> napi::Result<bool> {
 }
 
 /// Move and resize a window without changing Z-order (`SWP_NOZORDER`).
+///
+/// `no_activate` is an OPTIONAL trailing argument and defaults to the previous
+/// behaviour when omitted (`None`) or false: flags stay exactly `SWP_NOZORDER`,
+/// so every pre-existing caller is bit-for-bit unchanged.
+///
+/// Pass `true` to add `SWP_NOACTIVATE`. Without that flag SetWindowPos may
+/// activate the window it moves — Windows only leaves the foreground alone when
+/// the caller says so — which is wrong for any caller that repositions a window
+/// it does not want to bring forward (e.g. relocating a test window off the
+/// user's screen must not steal the focus that user is typing into).
 #[napi]
 pub fn win32_set_window_bounds(
     hwnd: BigInt,
@@ -92,14 +102,19 @@ pub fn win32_set_window_bounds(
     y: i32,
     cx: i32,
     cy: i32,
+    no_activate: Option<bool>,
 ) -> napi::Result<bool> {
     napi_safe_call("win32_set_window_bounds", || {
+        let mut flags = SWP_NOZORDER.0;
+        if no_activate == Some(true) {
+            flags |= SWP_NOACTIVATE.0;
+        }
         let result = unsafe {
             SetWindowPos(
                 hwnd_from_bigint(hwnd),
                 None, // SWP_NOZORDER => hwndInsertAfter ignored
                 x, y, cx, cy,
-                SWP_NOZORDER,
+                SET_WINDOW_POS_FLAGS(flags),
             )
         };
         Ok(result.is_ok())
--- a/tests/e2e/adr-029-viewport-gate.test.ts
+++ b/tests/e2e/adr-029-viewport-gate.test.ts
@@ -80,10 +80,14 @@ describe.skipIf(canvas === null || blank === null)("ADR-029 AC1 — viewport gat
     restoreAndFocusWindow(foreground!.hwnd);
     await new Promise((r) => setTimeout(r, 300));
 
-    // The canvas fixture opens centred and the blank window sits at the top-left,
-    // so a point inside the canvas and outside the foreground window always
-    // exists. Asserted rather than skipped: a layout change that makes the
-    // scenario un-stageable must fail loudly, not pass vacuously.
+    // Geometry premise: both fixtures land on the SAME screen — the non-primary
+    // monitor when the machine has one, the primary otherwise (see
+    // helpers/e2e-screen.ts) — with the canvas centred on it and the blank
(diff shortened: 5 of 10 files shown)


Harusame64:
@codex review

Harusame64:
## Round 1 findings applied — `5878526`

### F1 (P2) — cursor-move gate

`pickE2eScreen()` now requires `hasNativeCursorMove()` on top of the multi-monitor check. On a partial addon build (monitor enumeration present, cursor-move bindings missing) the reachable-bounds guard falls back to its primary-monitor variant, so every real click aimed at a non-primary window would be refused with `coordinate_outside_reachable_bounds`. A test window may only leave the primary monitor when the input side can follow it there; otherwise the fixtures stay exactly where they used to be.

### F2 (P2) — "focus is untouched" was wrong

`win32_set_window_bounds` passes `SWP_NOZORDER` but **not** `SWP_NOACTIVATE`, so `SetWindowPos` may activate the window it moves. Production (`src/win32/window_op.rs`) is deliberately untouched in this PR. The helper comments in `e2e-screen.ts` and `powershell-launcher.ts` now state the real behaviour: the move happens at spawn time, when the window is already the foreground one, so nothing observable changes — and the helper must not be reused to reposition a background window mid-test.

### F4 (P3) — the retreat is now complete

`untitled-window.ts` and `visual-only-canvas.ts` use the same placement, so every window-spawning fixture the suite owns leaves the user's screen.

- `pickE2eScreen` gained an optional `offset` so a fixture's historical primary-monitor coordinates are *translated* onto the e2e screen rather than every fixture collapsing onto one corner: the untitled window keeps its `700,420` offset (deliberately non-overlapping with the blank window's `120,120`) and the canvas stays centred on its monitor, which is the property the fixture's own comment relies on.
- The canvas is moved *after* spawn instead of inside `benches/fixtures/visual-only-canvas.ps1`, because that fixture is shared byte-for-byte with the ADR-024 round-trip bench — the e2e placement policy must not leak into the bench's canvas definition.
- **Nothing was held back.** The consumers were read first and the risky one was measured rather than assumed: the two HEADED roiCapture e2e files (`desktop-act-roi-capture`, `desktop-act-roi-carry-forward`) do a real OS click plus PrintWindow capture, OCR and ROI SSIM on the canvas, and both pass with the canvas at negative X. The viewport gate itself compares an entity against its origin window's rect with no monitor-level logic, so `untitled-window` was safe by construction.
- Out of scope, and documented as such in the helper: the notepad / chrome / ssh-wsl launchers — they either go through the PowerShell launcher already or drive a real application whose own placement is part of what the test observes.

### F3 (P2) — viewport-gate geometry comment

Rewritten to the real premise: both fixtures land on the same screen (non-primary when one exists, primary otherwise), canvas centred on it, blank window near its top-left, coordinates possibly negative. Assertions unchanged.

### F5 (P3) — DPI note

`$f.Location` is applied by a DPI-unaware-by-default PowerShell host, so a secondary monitor at a scale other than 100% can shift the landing point. Cosmetic only — every consumer reads the window's real rect back — and the note records that the measurement was taken on a 100%-scale virtual display only.

### Verification

Primary `0,0 1920x1080` + secondary `-1920,0 1920x1080`:

| Run | Result |
| --- | --- |
| `adr-029-viewport-gate.test.ts` | 5/5 passed |
| `mouse-verify-delivery.test.ts` | 6/6 passed |
| `desktop-act-roi-capture.test.ts` (HEADED) | 1/1 passed |
| `desktop-act-roi-carry-forward.test.ts` (HEADED) | 1/1 passed |
| `terminal.test.ts` | not re-run — launcher behaviour unchanged since the green run reported above |

Fixture rects observed on the secondary monitor: blank `{-1800,120,480x360}`, untitled `{-1220,420,320x240}`, canvas `{-1368,221,816x639}`, conhost `{-1800,120,993x399}`. `hasNativeCursorMove()` = `true` on this build.

Harusame64:
## Round 2 findings applied — `f09212f`

### P3-1 — the centring reference and the fit-clamp now read the same rect

The canvas centre offset was computed from the monitor's full `bounds` while the clamp inside `pickE2eScreen` worked on its `workArea`, so the two disagreed by half the taskbar height. WinForms' `StartPosition='CenterScreen'` — what `visual-only-canvas.ps1` does on a single-monitor machine — centres in the **work area**, so `pickE2eScreen` now accepts `"centre"` as its placement and derives the offset from the same work-area rect it clamps against. The relocated position is the historical one, translated.

### P3-3 — one enumeration instead of two

`visual-only-canvas.ts` called `pickE2eScreen` twice: once to read the bounds for the centring maths, once inside `moveWindowToE2eScreen`. It now makes a single call (`moveWindowToE2eScreen(hwnd, size, "centre")`), so a layout change between the two calls can no longer yield a centring offset computed against a different monitor than the clamp used.

### P3-2 — resolution assumption written down

The behaviour contract in the helper header now states that the chosen monitor is expected to be at least as large as the primary, since that is where the fixtures' historical coordinates come from. On a smaller secondary the fit-clamp pulls windows back toward the work-area origin and fixture rects that never overlapped on the primary can start to intersect — noted explicitly as *not* silent, because the ADR-029 viewport-gate premise ("a point inside the canvas and outside the blank window exists") is asserted and fails loudly.

### Verification

Primary `0,0 1920x1080` + secondary `-1920,0 1920x1080`:

| Run | Result |
| --- | --- |
| `adr-029-viewport-gate.test.ts` | 5/5 passed |
| `desktop-act-roi-capture.test.ts` (HEADED) | 1/1 passed |
| `eslint tests/e2e/helpers/ tests/e2e/adr-029-viewport-gate.test.ts` | clean |

Only the canvas placement maths changed in this round, so the blank-window / launcher / untitled-window runs from the previous round still stand.

Harusame64:
## Round 3 — Codex findings applied (`22e714e`)

This round takes the PR beyond test infrastructure: it adds one **optional** parameter to a production native binding.

### ① Codex P1 — "Move windows without activating them"

`win32_set_window_bounds` gains an optional trailing `no_activate: Option<bool>`. With `Some(true)` the SetWindowPos flags become `SWP_NOZORDER | SWP_NOACTIVATE`; omitted or `false` keeps exactly `SWP_NOZORDER`, the pre-existing flag set. The TS wrapper `setWindowBounds` passes it through and the hand-maintained `index.d.ts` declares it. `moveWindowToE2eScreen` now passes `true`.

Existing callers are untouched — both still call with five arguments, so their flag set is bit-for-bit unchanged:

| Caller | Args | Effect |
| --- | --- | --- |
| `src/tools/dock.ts:163` (via the wrapper) | 5 | `noActivate` = `undefined` → Rust `None` → `SWP_NOZORDER` only |
| `tests/unit/native-win32-panic-fuzz.test.ts:182` (raw binding) | 5 | same; napi optional arg, so no arity error |
| `tests/e2e/helpers/e2e-screen.ts` | 6 (`true`) | the only caller on the new path |

**Honest note on the severity.** The flag is documented-correct hardening, but the reported focus-steal did not reproduce on this machine, and the change should be read as defensive rather than as a measured bug fix. Two throwaway controls moved a **background** window and compared `GetForegroundWindow` before/after, with and without the flag:

- a TopMost form owned by another process: foreground unchanged in both cases;
- a non-topmost conhost console — the exact shape in the finding — while a different window held the foreground: foreground unchanged in both cases (`stolen = false` for the no-flag move as well).

That is consistent with `SWP_NOZORDER` suppressing the activation path (activation is tied to the Z-order change that flag removes). Passing `SWP_NOACTIVATE` explicitly is still the right contract for a relocation that must never come forward, and it costs nothing.

### ② Codex P2 — "Keep click targets on a reachable monitor"

Already closed in `5878526`: `pickE2eScreen()` requires `hasNativeCursorMove()`, so a partial addon build keeps every fixture on the primary monitor.

### ③ Codex P2 — "Skip relocation when the target cannot contain the window"

`pickE2eScreen` now refuses to relocate a window that does not fit the target work area and returns `null`, i.e. the window stays exactly where it is. Previously the origin was clamped to the work-area corner while the original size was still applied, so an over-large window (a maximised console from a 4K primary sent to a 1080p secondary) hung off the edge back onto the primary and covered **both** screens. With the fit gate in front, the clamp only ever pulls an over-large *offset* back, never an over-large window. The behaviour contract in the helper header was updated to match.

### ④ Codex P2 — "Make untitled-window lookup tolerate DPI scaling"

`untitled-window.ts` identifies its form by HWND instead of by an exact `320x240` rect match. The form writes `$f.Handle` to a temp file from its `Shown` handler (the launcher's PID-file idiom) and the lookup reads it back, then uses the rect exactly as Win32 reports it. On a scaled monitor the DPI-unaware WinForms host reports a scaled physical rect, so the old size comparison could never match: the fixture waited 10s, returned `null`, and the ADR-029 probe-path coverage was skipped silently. The temp file is removed in `close()`.

One deliberate deviation: the rect is read with `win32GetWindowRect`, not `enumWindowsInZOrder`. The enumeration drops untitled windows by design — that is precisely this fixture's premise and the reason the probe path exists — so an enumeration-based lookup could never find it.

### Guards

`build:rs` OK · `cargo check --workspace --locked` OK · `cargo test -p desktop-touch-engine` 180 passed, 1 environment-flaky (`duplication::thread::tests::dxgi_emit_fork_enable_disable_roundtrip`, a DXGI dirty-rect test unrelated to `window_op.rs`; passes on a targeted re-run) · `check:napi-safe` OK · `check:native-types` OK (73 exports) · `check:no-koffi` OK · `tsc` OK · eslint clean.

### Live verification

Primary `0,0 1920x1080` + secondary `-1920,0 1920x1080`:

| Run | Result |
| --- | --- |
| `adr-029-viewport-gate.test.ts` | 5/5 passed |
| `mouse-verify-delivery.test.ts` | 6/6 passed |
| `desktop-act-roi-capture.test.ts` (HEADED) | 1/1 passed |

Fixture rects, all on the secondary monitor: blank `{-1800,120,480x360}`, untitled `{-1220,420,320x240}`, canvas `{-1368,197,816x639}`, conhost `{-1800,120,993x399}`. The canvas `y` moved from `221` to `197` versus the previous round, which is the work-area centring from Round 2 landing: `(1032 - 639) / 2 ≈ 197`.

No test assertions were changed in this round.

Harusame64:
## Round 3 review closed — `c8c3302`

No P1 or P2 findings. Two P3s, one applied and one accepted as-is.

**P3-1 (applied).** The `SWP_NOACTIVATE` contract was documented as if it fixed an observed focus steal. It did not, and the code now says so at the three places a future reader could be misled — the Rust doc comment on `win32_set_window_bounds`, `moveWindowToE2eScreen`, and the console relocation in the PowerShell launcher. The measurement: moving a background window with `SWP_NOZORDER` alone left the foreground untouched **with and without** the flag, across two controls (a TopMost form owned by another process, and a non-topmost conhost console while a different window held the foreground). Activation appears tied to the Z-order move that `SWP_NOZORDER` suppresses. The flag stays — it is the right contract for a relocation that must never come forward, and one OS build is not licence to drop it.

**P3-2 (accepted, not changed).** The suggestion was to restructure the untitled fixture's HWND temp file. It is deliberately the same idiom as the PowerShell launcher's PID file (write from the child, read back, unlink on close), and the reviewer agreed the change was not required. Splitting one fixture off from that shared pattern would cost more in consistency than it buys.

Docs only in this commit; `tsc` and eslint clean. The live verification from the previous comment still stands — nothing executable changed.

Harusame64:
## Round 4 — Codex mixed-DPI P2 addressed (`2366dce`)

**The finding.** On a secondary monitor scaled above 100%, a DPI-unaware PowerShell host treats a fixture's `W`/`H` as *logical* units while `pickE2eScreen` compares them against the native *physical* work area. The form can then be physically larger than the fit check allowed, hang off the far edge and spill back onto the primary — covering the screen this placement exists to protect.

**The fix, scoped to where the mechanism actually applies.** Of the three fixtures, only two position themselves at spawn time from logical constants: `blank-window.ts` and `untitled-window.ts`. Both now call `SetProcessDpiAwarenessContext(PER_MONITOR_AWARE_V2)` before constructing their Form, so `$f.Size` / `$f.Location` are physical pixels — the same units the placement helper reads. Probed on Windows 11 before implementing: `powershell.exe` starts DPI-unaware (awareness `0`), the call returns `true`, and the process ends at awareness `2`. A `false` return (a host that declared awareness in its manifest) is ignored, which is exactly the previous behaviour. Note this removes an asymmetry rather than introducing awareness anywhere new: our own process has been per-monitor aware since module init (`win32.ts` → `SetProcessDpiAwareness(2)`), which is why its rects are physical in the first place.

**`visual-only-canvas` is deliberately left DPI-unaware.** It is placed by reading its real physical rect back after spawn and relocating it with that physical size, so the logical-vs-physical mismatch cannot occur there — PMv2 buys nothing for the fit check. It was nonetheless attempted in this round (via a `-Command` PMv2 prelude, chosen so the bench-shared `benches/fixtures/visual-only-canvas.ps1` stayed byte-identical) and the HEADED `desktop-act-roi-capture` e2e **failed twice in a row** — `roiCapture` undefined, i.e. the click stopped producing a detectable localized repaint — on a path that has now passed four times unchanged. Reverted to the `-File` spawn. The helper comment records why the fixture does not need PMv2, and what the residual mixed-DPI concern actually is: a DPI-unaware window that crosses a scale boundary is stretched by the OS, so its physical size changes after the move.

**What is NOT verified.** Both displays on this machine are at 100% scale, where PMv2 is a no-op — so this change is reasoned and regression-tested, never exercised on a scaled monitor. Mixed-DPI measurement is deferred to the multi-DPI dogfood and recorded in the internal remaining-work notes (two rows: the PMv2 fix awaiting a scaled-monitor run, and the canvas OS-stretch residue) so it cannot be lost.

### Verification (no-regression)

| Run | Result |
| --- | --- |
| `adr-029-viewport-gate.test.ts` | 5/5 passed |
| `mouse-verify-delivery.test.ts` | 6/6 passed |
| `desktop-act-roi-capture.test.ts` (HEADED) | 1/1 passed |
| `tsc` / eslint | clean |

Fixture rects are byte-identical to the previous round, as expected at 100% scale: blank `{-1800,120,480x360}`, untitled `{-1220,420,320x240}`, canvas `{-1368,197,816x639}`, conhost `{-1800,120,993x399}`.

No test assertions were changed.

PR #559, opened 2026-07-29
feat(clipboard): native Win32 clipboard read/write with verified delivery (ADR-033 PR-1)

Harusame64:
## What

Native Win32 clipboard read/write for the `clipboard` tool (ADR-033 PR-1). The tool now uses an in-process `CF_UNICODETEXT` transaction when the compiled addon is present, and keeps the PowerShell path as a fallback for addon-less builds.

Plan: `desktop-touch-mcp-internal@e65be6a:docs/adr-033-plan.md` (Approved; this PR is Phase 0 + Phase 1 / PR-1 of that plan)

## Why

- **Defender false positive**: the PowerShell path spawns `powershell.exe` with an inline base64-decoding command line — the shape of a well-known malware TTP. Microsoft Defender scored the server as `Trojan:Win32/Commando.A!ml` and killed it mid-session (2026-07-29). The native path removes the spawn entirely; there is no command line left for a heuristic to score.
- **Performance**: measured on this branch after the async conversion — write+verify p50 **4.2 ms native vs 372 ms PowerShell (~89x)**, read p50 **3.4 ms vs 318 ms (~92x)**. (Both native calls now run on the libuv threadpool with a watchdog timeout — see the review thread — which costs read about 1 ms of dispatch overhead versus the synchronous binding it replaced.)
- **Capacity**: the fallback embeds base64 in a command line and physically caps out near ~12k characters (raw `ENAMETOOLONG` today, an opaque error against a schema that advertises 100k). The native path delivers the full documented 100,000 characters (measured: 4.3 ms p50 at 100k).

### Rust: `src/win32/clipboard_text.rs` (new)

- `win32_clipboard_read_text()` / `win32_clipboard_write_text_verified(Buffer)` — async napi calls (`AsyncTask` on the libuv threadpool, with an optional `AbortSignal` that cancels still-queued work), never throwing on Win32 failure (`ok:false` + `reason`, the `ConsolePasteResult` idiom). They started out synchronous; review round 5 established that a hung delayed-rendering clipboard owner can block `GetClipboardData` indefinitely, which on a synchronous binding froze the whole server — the TS layer now enforces the same 4 s read / 5 s write timeouts the PowerShell path always had, shared as constants across both backends, and aborts a timed-out task so a not-yet-started write can never run late and overwrite the clipboard after its caller was told it failed.
- Write verifies **twice** (plan D-2): an in-session read-back while the clipboard is still open (proves the OS stored the bytes), and a post-close re-read (the only leg that can observe a clipboard manager / DLP agent / RDP redirector replacing the payload — they receive `WM_CLIPBOARDUPDATE` only after the lock is released). Failing to re-open for the second leg is **not** a write failure: contention on a verification read is not evidence of loss (`postCloseChecked:false`, `ok:true`).
- **Verdict = byte comparison; sequence number is diagnostic only** (plan D-5). Measured (P0-2): one `EmptyClipboard`+`SetClipboardData` transaction bumps the sequence number 5 times, 3 of them inside `CloseClipboard` when the OS synthesises `CF_LOCALE`/`CF_TEXT`/`CF_OEMTEXT` — a sequence-based verdict would need to know that magic number. All synthesis completes inside `CloseClipboard`, and reading synthesised formats bumps nothing, so the existing 3-point race anchors in `clipboard_snapshot.rs` (taken post-close) are unaffected.
- Payload crosses the napi boundary as **UTF-16LE bytes, not a JS string**: napi's String bridge transcodes through UTF-8 and would replace an unpaired surrogate with U+FFFD *before* the byte comparison ran — the verification would pass on mutated text.
- Reuses `clipboard_snapshot.rs`'s hidden-owner window (I-11) and `open_clipboard_with_retry` (I-12). Production changes to `clipboard_snapshot.rs` are deliberately minimal: `pub(crate)` on `open_clipboard_with_retry` / `global_free` (visibility only), plus one conscious exception to the paths-D/E freeze added in review round 6 — the hidden-owner class registration guard moved from an `AtomicBool` swap (which published "registered" *before* `RegisterClassExW` ran, a latent first-use race that the async conversion made reachable from concurrent worker threads) to `std::sync::Once`, which publishes completion only after registration finishes. No observable behaviour change for single-threaded callers; a multi-thread stress test pins the fix (it catches the old implementation).
- HGLOBAL ownership rules (I-9) spelled out in the module doc; both failure branches free explicitly.

### TS: `src/tools/clipboard.ts`

- `selectBackend()` is evaluated once at the top of each handler, so even the catch path reports which backend served the call.
- **`backend: "native" | "powershell"` is now part of the public envelope** (success payload and failure context — plan D-7), so a Defender-related regression is diagnosable from the response.
- Fallback write gets an upfront `FALLBACK_MAX_CHARS = 12_000` check → new compact code **`ClipboardWriteTooLargeForFallback`** with SUGGESTS entries, instead of a raw `ENAMETOOLONG` from `spawn`. The schema's `max(100_000)` stays as the native path's contract.
- Native failure messages are lowercase on purpose (generic classification; a PascalCase single-token message is this repo's compact-code producer shape and would demand its own SUGGESTS entry — `oq8` sweep).

## Verification

| check | result |
|---|---|
| `check:stub-catalog` / `check:failwith-fixtures` (both regenerated, diffs reviewed) | PASS |
| `check:napi-safe` / `check:native-types` / `check:no-koffi` | PASS |
| `npm run build` (tsc) / `npm run lint` | PASS |
| `cargo check --workspace --locked` / `cargo test -p desktop-touch-engine` (17 unit tests in the new module) | PASS |
| Baseline before the change (P0-3, on `main`): e2e `clipboard-readback` 3/3 + Rust `#[ignore]` clipboard tests 3/3 | PASS |
| e2e on this branch: `clipboard-native-backend` + `clipboard-readback` — 14/14 (backend:'native', 100k round-trip, CJK / ZWJ emoji / astral / lone surrogate / combining / CRLF byte-exact, embedded NUL rejected) | PASS |
| Unit suite with the addon present | PASS (see thread) |

**Mutation checks on the no-spawn guard** (the guard is not vacuous):
- M1: force `selectBackend()` to `"powershell"` → 20 tests fail (only fallback-path cases survive).
- M2: delete only the native dispatch line while keeping the label `"native"` → 4 guard tests fail.
Both mutations reverted; 29/29 green after restore.

**Bench** (this branch at the final, async HEAD; first four rows from `npm run bench:clipboard-latency -- latency 20`, size rows from `npm run bench:clipboard-latency -- sizes 10`):

| path | n | min ms | p50 ms | p95 ms | max ms |
|---|---|---|---|---|---|
| native write+verify | 20 | 3.26 | 4.18 | 4.93 | 5.19 |
| powershell write+verify | 20 | 322.5 | 371.5 | 1643.2 | 1734.1 |
| native read | 20 | 2.90 | 3.44 | 3.74 | 3.75 |
| powershell read | 20 | 299.6 | 317.9 | 326.5 | 328.9 |
| native write+verify — 11 chars | 10 | 4.47 | 5.73 | 6.92 | 6.92 |
| native write+verify — 12,000 chars | 10 | 5.24 | 6.26 | 12.69 | 12.69 |
| native write+verify — 100,000 chars | 10 | 5.24 | 5.65 | 6.94 | 6.94 |

## Notes for reviewers

- **`tests/unit/clipboard-envelope-secrecy.test.ts` is new in this PR and was touched twice**: the first commit adds it as the I-2 pin against the unchanged PowerShell path (plan P0-1 — green *before* the migration, so the native path inherits a pin that already passed rather than one written to fit it). The test commit then extends it: a native-engine partial mock pins the PowerShell backend for the original describes (without it, any machine with a built `.node` would silently switch backends and reach the real clipboard from a unit test), plus two native-backend describes for the same I-2 invariant. The original assertions are unchanged between the two commits.
- The spike branch (`spike/native-clipboard`) was used as a reference implementation and rewritten clean, per plan; intentional divergences are listed in the plan thread.
- Not in scope (plan §2): `typeViaClipboard` (PR-2 / F-1), key-locker `-EncodedCommand` (F-2), any behaviour change to `foreground_flash` / console-paste (I-27). Merging this PR does **not** close the Defender exposure — path C (`typeViaClipboard`) is the hotter path and lands in PR-2.
- Review-loop by-catch, recorded as follow-up risk R-g in the internal plan (`desktop-touch-mcp-internal@27388a4:docs/adr-033-plan.md`): the pre-existing `clipboard_snapshot.rs` set/restore paths share the EmptyClipboard-before-alloc destructive window fixed here for `clipboard_text.rs`; deliberately untouched (paths D/E are frozen for this ADR). **Merge ordering**: internal PR #43 (which carries R-g) should merge before or together with this PR so the follow-up record lands on internal `main`.


Harusame64:
@codex review

Harusame64:
## Review loop — Opus Round 1 (record)

Findings, all confirmed against the code by the director; fixes land as one "Round 1 findings apply (Opus)" commit.

**P1-1** — plan D-5 says the sequence number stays in the envelope as a diagnostic hint; the implementation exposes `sequenceAfterWrite` at the napi layer but drops it (and `postCloseChecked`) from the tool envelope, so a caller cannot distinguish a both-legs-verified write from one whose post-close verification leg was skipped. Fix: write success payload gains `sequenceAfterWrite` + `postCloseChecked` (+ `postCloseSkipReason` when the leg was skipped); catalog/fixtures regenerated.

**P2-1** — `map_write_outcome` fails the write with `clipboard_get_data_failed` when the in-session read-back could not run, even if the post-close re-read matched byte-for-byte — but a post-close match is delivery proven with exactly the old PowerShell path's verification semantics, and a failed *verification read* is not evidence of a failed *write* (the module doc already applies this logic to the post-close leg). Fix: post-close match overrides an unreadable in-session leg (`ok:true`, disclosed via `inSessionReadable:false`); a *readable but mismatching* in-session leg still fails as `readback_mismatch`.

**P2-2** — failure `actualBytes` falls back to `inSessionBytes`, which is 0 when the write never ran (open failure), producing a "0 vs N → cleared" misread of an untouched clipboard. Fix: `actualBytes` is only reported from a leg that actually read something (post-close first, then in-session); omitted on write-failure paths.

**P2-3** — the "blocking ≤100 ms" doc claim is wrong for write: two `open_clipboard_with_retry` transactions make the worst case ~200 ms of synchronous V8-thread block under contention (the replaced async PowerShell path never blocked the loop). Fix: doc corrected; risk persisted in the internal plan.

**P3-1** — module doc cites the spike's ~60x/~100x while the PR re-measured 84x/119x on this branch. Fix: doc updated to the clean-branch numbers.

**P3-2 / P3-3** — PR description inaccuracies (the envelope-secrecy test file is new in this PR, not pre-existing; bench table rows come from two different bench modes). Fixed directly in the description above.

Confirmed clean by the reviewer: `clipboard_snapshot.rs` visibility-only diff (paths D/E untouched), no public-API breakage, I-9 ownership on all four branches, lone-surrogate byte path lossless both directions, and the 12k/100k fact sweep consistent everywhere except P3-1.

Harusame64:
## Review loop — Codex Round 1 (record) + Round 1 fixes pushed

Codex Round 1 (reviewed `8f9381f6`) raised two findings; both confirmed by the director and fixed in `38c21f19`:

**P2 (accepted)** — `read_unicode_text_locked` collapsed "format absent" and "retrieval failure" into one `None`: a `GetClipboardData` failure after `IsClipboardFormatAvailable` advertised the format surfaced as `ok:true, text:""` on read, and as a spurious `clipboard_replaced_after_write` on the write's post-close leg. Fix: the primitive now returns three values (`Ok(None)` = format absent → contractual empty read; `Ok(Some)` = read; `Err` = retrieval failure), with per-call-site pure mapping helpers pinned by unit tests — read reports `clipboard_get_data_failed`, the in-session leg treats a missing format as "readable and empty" (a store that did not take effect must fail as `readback_mismatch`), and a post-close retrieval failure becomes `postCloseChecked:false` + skip reason instead of a false "replaced" verdict (same logic as Opus R1 P2-1: a failed verification read is not evidence of a failed write). A counter-test pins that a genuinely vanished payload still fails.

**P3 (accepted)** — the benchmark restored the user's clipboard only on the happy path. Fix: measurement wrapped in `try`, restore (and partial-results table) moved to `finally` in both modes.

Round 1 is now fully applied: `5e2e0298` (Opus findings) + `38c21f19` (Codex findings).

Re-verification on `38c21f19` (real clipboard): e2e `clipboard-native-backend` + `clipboard-readback` **14/14**, Rust `#[ignore]` clipboard tests **3/3**, `cargo test -p desktop-touch-engine` **213/213**, all check scripts green. One observation for the record: running the `#[ignore]` clipboard tests *immediately after* the e2e suite in the same shell produced a one-off failure (2/3) that did not reproduce in isolation — consistent with the real-clipboard contention flake risk that keeps these tests out of CI (plan P0-4), not with a code defect.

@codex review

Harusame64:
## Review loop — Round 2 (record) + fixes pushed

Opus Round 2: **P1 zero**, all Round 1 items verified closed, both director rulings (PowerShell fallback reporting `postCloseChecked:true` while omitting `sequenceAfterWrite`; in-session `Ok(None)` compared as "readable and empty") judged sound. New findings P2×2 / P3×1. Codex Round 2 (on `5e2e0298` / `38c21f19`): P2×3. All five confirmed by the director and fixed:

`7185daae` — **Round 2 findings apply (Opus)**
- P2-1: the fallback write success shape is now pinned (`postCloseChecked:true` asserted; `sequenceAfterWrite` asserted absent — the fallback never observes the sequence number, so reporting one would be a lying diagnostic).
- P2-2: verdict priority reordered — a post-close leg that *observed* a mismatch now reports `clipboard_replaced_after_write` (with its recovery hint) even when the in-session leg was unreadable, instead of the less useful `clipboard_get_data_failed`. Same principle as R1 P2-1: a leg that observed content beats a leg that failed to observe. `ok` is unchanged in all eight input combinations; only reason/hint quality improves.
- P3-1: the tool-description diagnostics sentence moved behind Caveats/Recovery under a `Diagnostics:` label, native-only scope moved to the front of the sentence, and the backend difference in `postCloseChecked` spelled out.

`e3cb1711` — **Round 2 findings apply (Codex)**
- P2-a: the chunking suggestion no longer implies back-to-back writes accumulate — it now says to paste each chunk into the target application before writing the next, since every write replaces the clipboard.
- P2-b: `set_unicode_text_locked` is now prepare-first: `GlobalAlloc` → lock → copy → unlock → `EmptyClipboard` → `SetClipboardData`. An allocation failure can no longer leave the user's clipboard emptied with nothing written; on `EmptyClipboard` failure the prepared handle is freed (I-9 held on every branch). The same pre-existing ordering in `clipboard_snapshot.rs` (paths D/E) is deliberately untouched — recorded as follow-up risk R-g in the internal plan.
- P2-c: bench and e2e cleanup now restore based on whether the snapshot *succeeded*, not on text truthiness, so an initially-empty clipboard is restored to empty instead of being left with the last payload.

Re-verification on `e3cb1711` (real clipboard): e2e `clipboard-native-backend` + `clipboard-readback` **14/14** — this also live-exercises the reordered set path, which the `#[ignore]` snapshot round-trip does not reach. `cargo test -p desktop-touch-engine` **213/213** (+ round-trip `--ignored` 1/1), all check scripts green, catalog/fixture regeneration diff-clean.

@codex review

Harusame64:
## Review loop — Round 3 (record) + fixes pushed

Opus Round 3: **P1 zero**; all Round 2 fixes verified closed by independent re-derivation (full truth-table expansion of the verdict priority — nine reachable states, `ok` invariant in all of them; I-9 ownership on all five branches of the reordered set path; no interference with inventory I-8 since `clipboard_snapshot.rs` is untouched). New: P2×1, P3×1. Codex Round 3 (on `e3cb1711`): P2×1. All three confirmed and fixed:

`98336064` — **Round 3 findings apply (Opus)**
- P2-1: `npm run test:native-clipboard` exercised only the pre-existing `clipboard_snapshot.rs` tests — the gate's name promised coverage of the new module it never ran. Three `#[ignore]` real-clipboard tests now live in `clipboard_text.rs` (CJK/emoji/CRLF round-trip; empty payload present-but-empty; unpaired surrogate preserved end-to-end), picked up automatically by the script's `clipboard` filter. Each snapshots and restores the user clipboard, with the R2 P2-c distinction (failed snapshot → skip restore; genuinely empty → restore empty). The script also gains `--test-threads=1`: the clipboard is one global resource and the pre-existing race test plays a deliberate competing writer, so parallel runs fail for non-defect reasons — a gate that flakes is not a gate. Result: **6/6 green**.
- P3-1: follow-up risk R-g lives on the internal progress branch; the PR description now references `desktop-touch-mcp-internal@27388a4` and notes the merge ordering (internal PR #43 first or together).

`33064d63` — **Round 3 findings apply (Codex)**
- P2: the native read trusted `GlobalSize` unboundedly — a foreign multi-gigabyte `CF_UNICODETEXT` allocation would be scanned and copied on the V8 thread, and an infallible `to_vec` allocation failure would abort the server process (the replaced PowerShell path paid this in a child process). Fix: `MAX_READ_BYTES = 8 MiB` checked O(1) before any slice is built (`clipboard_text_too_large`), and the copy is now fallible (`try_reserve_exact`, per the `gdi.rs` precedent) returning `clipboard_alloc_failed` instead of aborting. The limit clears the 100k-character write contract 40× over; constant + boundary pinned by unit test, and an oversized read is pinned to flow through the existing three-value leg mapping with no new plumbing. Documented consequence: a leg that hits the limit reports "could not read", not "replaced" — an accepted false-negative of the same class as the same-text rewrite already documented under D-5.

Verification on `33064d63`: `test:native-clipboard` **6/6**, `cargo test -p desktop-touch-engine` **215/215** (+7 ignored), clipboard unit files **82/82**, all check scripts green, catalog regenerated (8 MB read note added to the tool description).

@codex review

Harusame64:
## Review loop — Round 4 (record) + fixes pushed

Opus Round 4: **P1 zero / P2 zero**, all Round 3 fixes verified closed (8 MiB rejection confirmed O(1) before any pointer deref, lock accounting balanced on all three early returns, oversized reads pinned through the existing leg mapping; the merge gate of plan P1-7 judged satisfied). New: P3×2. Codex Round 4 (on `33064d63`): P2×1. All three confirmed and fixed:

`6c134bf3` — **Round 4 findings apply (Opus)**
- P3-1: the `reason` enumerations in the Rust result doc and `native-types.ts` now list all four values (`ClipboardError::as_reason()`, `clipboard_get_data_failed`, `clipboard_text_too_large`, `clipboard_alloc_failed`) — the TS side had drifted since the three-value read landed. Also spells out that `hasText:false` with `ok:true` is absence, not failure.
- P3-2: the three real-clipboard tests now restore via a `Drop` guard (`ClipboardGuard`), so an assertion failure — the moment a developer most wants their clipboard back — restores it too. Effectiveness proven by mutation: with a deliberate failing assertion injected, the test failed AND a pre-set sentinel survived; mutation reverted, 6/6 green.

`cec27de7` — **Round 4 findings apply (Codex)**
- P2: a green `npm run test:native-clipboard` still left the user's clipboard holding the race test's intruder payload, because that pre-existing test *asserts* that restoration is skipped. Fix: `ClipboardGuard` promoted to a shared `#[cfg(test)] test_support` module and applied to the three pre-existing `#[ignore]` tests — each test's claim is untouched (the race test still asserts the skip), while the guard puts the machine back as the suite found it, panic path included. Sentinel-verified: clipboard set before the gate, 6/6 green, sentinel intact after. `clipboard_snapshot.rs` remains byte-identical outside `#[cfg(test)]` (all hunks sit below the `mod tests` line; the branch-total production diff of that file is still the two visibility keywords). The second half of the finding (per-test restore skipped on panic) had already been fixed in `6c134bf3`, which this review predates.

Verification on `cec27de7`: `test:native-clipboard` **6/6 with sentinel surviving**, `cargo test -p desktop-touch-engine` **215/215** (+7 ignored), clipboard unit files **82/82**, all check scripts green. Production code is unchanged by `cec27de7` outside `#[cfg(test)]`, so the e2e 14/14 run on the prior HEAD remains valid for the production paths.

@codex review

Harusame64:
## Review loop — Round 5 (record) + fixes pushed

Opus Round 5: **Approved, P1/P2/P3 zero** — reason enumerations verified exhaustive against the actual return sites, the `Drop` guard shown structurally panic-safe (it calls through the `catch_unwind`-wrapped entry point), the `test_support` promotion confirmed absent from production builds, and the full branch re-sweep clean.

Codex Round 5 (on `cec27de7`): **P1×1** — confirmed by the director and fixed:

`901c97ea` — **Round 5 findings apply (Codex)**
- P1: a delayed-rendered `CF_UNICODETEXT` whose owning application is hung makes `GetClipboardData` block indefinitely (synchronous `WM_RENDERFORMAT` round-trip), and on a synchronous napi binding that froze the entire server — an availability regression against the replaced PowerShell path, which ran out of process under a timeout. Fix: both native entry points are now `AsyncTask`s on the libuv threadpool (the Win32 logic is untouched, split into non-napi blocking cores that the `#[ignore]` tests and `ClipboardGuard` keep using directly), and the TS layer enforces the same timeouts the PowerShell path always had — measured from the code as read 4000 ms / write 5000 ms, now shared constants across both backends. On timeout the caller gets a named, lowercase failure with a hint explaining that the clipboard owner is hung and the clipboard is unusable system-wide until it recovers (retrying will not help). The write path is converted symmetrically because its post-close verification leg can also read foreign, delayed-rendered content.
- Disclosed residual hazard (module doc): a timed-out worker cannot be killed — it stays blocked until the owner recovers, and repeated reads against a persistently hung owner could exhaust the libuv pool. That window is exactly the window in which the clipboard is already unusable system-wide for every application. The timeout wording deliberately avoids the phrase "timed out", which this repo's error classifier would route to a retry-suggesting code — the opposite of the correct advice here (verified by measurement).
- Bonus: the R1-documented ~200 ms worst-case synchronous block (risk R-f) is gone with the thread hop; the internal plan records R-f as resolved.

`5330e8a6` — doc re-measure: the module doc's latency figures were from the synchronous build; re-measured at the async HEAD (write p50 4.18 ms vs 371.5 ms ≈ 89x; read p50 3.44 ms vs 317.9 ms ≈ 92x — the thread hop costs read ~1 ms, kept in the doc as context against ever reverting to sync). PR description updated to match.

Verification on the async HEAD: e2e `clipboard-native-backend` + `clipboard-readback` **14/14** (first live run through the Promise-returning bindings), `test:native-clipboard` **6/6 with sentinel surviving**, clipboard unit files **84/84** (two new fake-timer tests pin the give-up path on both read and write, plus a slow-but-answering counter-case), `cargo test -p desktop-touch-engine` clean except the known pre-existing DXGI flake (passes in isolation, `src/duplication/` untouched by this branch), all check scripts green.

@codex review

Harusame64:
## Review loop — Round 6 (record) + fixes pushed

Opus Round 6: **P1 zero** — the async conversion verified sound on all four requested axes (Send boundaries and the read-only intermediate `Task::Output`; `Promise.race` leak/unhandled-rejection claims re-derived; the "gave up …" wording checked against every classifier arm including the class-name path; blocking-core split leaves the `#[ignore]` semantics intact). New: P2×1, P3×2. Codex Round 6 (on `5330e8a6`): **P1×1, P2×1** — two concurrency findings the async conversion made reachable. All five confirmed and fixed:

`e3c5fc9a` — **Round 6 findings apply (Opus)**
- P2-1: a timed-out **write** now gets its own hint: the abandoned write may still complete once the hung owner recovers (it can be past `EmptyClipboard`, or past delivery entirely and hung only in the post-close verification), so the clipboard contents are **indeterminate** — re-read before relying on them. This mirrors, in the opposite direction, the `ClipboardWriteNotDelivered` guidance to treat the clipboard as un-written; the tool description carries the same contrast.
- P3-1: the fallback now shares the *disclosure*, not just the timeout values — an `execFile` timeout kill is detected by its measured shape (`killed:true` + a signal; the other three failure shapes measured to confirm they don't match) and gets the same hung-owner hint. Pinned in both directions.
- P3-2: the `#[ignore]` docstrings no longer overstate coverage (the core-calling tests skip the napi `Buffer` boundary; that half belongs to the e2e suite).

`3788ccd7` — **Round 6 findings apply (Codex)**
- P1: a timed-out write whose `AsyncTask` was still **queued** behind a saturated pool would run later and overwrite the clipboard long after its caller was told it failed. Fix: both entry points take an optional `AbortSignal` (`AsyncTask::with_optional_signal`, napi 2.16.17); `withTimeout` aborts on timeout, and napi's cancel path (verified in its source) cancels queued work while letting in-flight work finish — exactly matching the "may still complete" disclosure above, which now describes the *only* remaining late-completion window. Reject-then-abort ordering keeps the envelope from being won by the hintless `AbortError`.
- P2: `ensure_class_registered` published "registered" **before** `RegisterClassExW` ran (`AtomicBool::swap`), so the first pair of concurrent native calls could race the second thread into `CreateWindowExW` against an unregistered class. Fixed with `std::sync::Once` — completion is published only after registration finishes. This is a deliberate, recorded exception to the paths-D/E freeze (the async conversion made the latent race reachable; no observable change for single-threaded callers), pinned by an 8-thread barrier stress test whose teeth were measured by mutation: reverted to the old swap it fails 1 in 5 runs, with `Once` it's 8-for-8 green. Recorded as risk R-h in the internal plan (`desktop-touch-mcp-internal@bd5f7c1`), and the PR description's earlier "visibility-only" claim about `clipboard_snapshot.rs` is corrected accordingly.

Verification on `3788ccd7`: e2e **14/14** (live through the `AbortSignal`-bearing bindings), `test:native-clipboard` **6/6** with sentinel surviving, clipboard unit files **88/88**, all check scripts green (the only `cargo test -p` failure is the known pre-existing DXGI flake, untouched by this branch and passing in isolation).

@codex review

Harusame64:
## Review loop — Round 7 (record) + fix pushed

Opus Round 7: **P1 zero / P3 zero**, all Round 6 fixes verified closed — the execFile timeout-kill discriminator checked against Node's own `exithandler` implementation, the reject-before-abort ordering derived structurally (a synchronously-settled race cannot be won by the later `AbortError`), the `Once` closure shown poison-free, and the stress test's honesty about its own detection power confirmed (its non-`#[ignore]` status is CI-safe because CI compiles but does not run cargo tests). One new finding, P2×1, fixed:

`92899564` — **Round 7 findings apply (Opus)** (test-only, +41)
- P2-1: the reject-before-abort ordering in `withTimeout` is load-bearing — swapped, a hung-clipboard timeout envelope degrades into the addon's hintless `AbortError` — yet every existing timeout test mocked the native call as a promise that never settles, so the swap mutation passed the whole suite. A new test mocks the addon faithfully (rejects with `AbortError` when the signal fires) and asserts the envelope still carries the hung-owner hint. Mutation-verified: with the two lines swapped, exactly this test fails (`expected '' to contain 'not responding'`) while the three pre-existing timeout tests stay green — demonstrating both the new pin's teeth and the prior blind spot; restored, 43/43 green and the production file is byte-identical.

@codex review

Harusame64:
## Post-merge dogfood (plan P1-8) — decisive gate PASS

Ran against the merged `main` (`ea10b16d`, freshly built addon), driving the server over stdio JSON-RPC while an external watcher polled (~40 ms) for every `powershell.exe` process created on the machine, recording PID + parent + command line.

**Clipboard exercise** (all through the real `clipboard` tool):
- 15 rounds of write+read with a ~52.5k-character CJK/emoji payload: every round `ok:true`, `backend:"native"`, `postCloseChecked:true`, byte-exact read-back.
- 100,000-character write+read: `ok:true`, exact read-back — the old PowerShell path capped out near 12k with a raw `ENAMETOOLONG`.
- The user clipboard was saved via the tool at the start and restored at the end (`RESTORED ok=true`).

**Process observation (the decisive gate)**:
- Watcher sensitivity proven by a positive control: a deliberately spawned `powershell.exe` from the driving shell was caught.
- **Zero `powershell.exe` spawns from the clipboard path.** The single server-child spawn observed carries the `Add-Type -AssemblyName System.Windows.Forms … NotifyIcon` command line of `src/utils/tray.ts` — path H (tray), which the ADR-033 inventory documents as out of scope for this ADR (no base64 payload, user-action origin, low TTP resemblance) and which is already queued for follow-up recording in Phase 3.

Defender event log (1116/1117) remains a secondary signal over the next week per plan P3-1; per plan D-1, this merge does **not** close the Defender exposure — path C (`typeViaClipboard`) is the hotter path and lands in PR-2.

